Issue #20, opened 2022-11-04
Syntax Error in CmsElementText.vue


benjamin-ott:
The CmsElementText.vue component produces the following error:

Uncaught (in promise) SyntaxError: The requested module '/_nuxt/node_modules/he/he.js?v=536aebf1' does not provide an export named 'default' (at CmsElementText.vue:8:1)

![Bildschirmfoto 2022-11-04 um 09 37 52](https://user-images.githubusercontent.com/8365682/199929218-cf97e57b-1215-4bab-9399-2baa9190901c.png)

benjamin-ott:
Adding this to the tsconfig.json file seems to solve the problem:
`"compilerOptions": {
        "esModuleInterop": true
    }`

elkmod:
Hi @benjamin-ott, we're looking at this issue right now. he.js is a dependency which is required to parse the string value of a text element into a DOM-tree, so we can replace all `<a>` tags with the native `<nuxt-link>` link elements. We are looking into cheaper replacements at the moment.

patzick:
thanks @benjamin-ott , that should be now fixed with `0.2.0`
https://github.com/shopware/frontends/blob/main/packages/cms-base/CHANGELOG.md#020

please let us know if there is still a problem, then we'll reopen the issue 🖖

PR #25, opened 2022-11-16
replace he with entities to decode html

BrocksiNet:
- he.js package size is to heavy
- only the CmsElementText.vue was more then 90kb size
- we will use [entities](https://github.com/fb55/entities#readme) now, should also be faster


BrocksiNet:
This change will also fix this issue https://github.com/shopware/frontends/issues/20

patzick:
copied in: https://github.com/shopware/frontends/commit/d7da2c11b6cbef23a83d0ffdb95d7e382795b7f3
sorry for not merging directly, w're not yet having two-way repo sync between internal Gitlab. You'll be mention in changelog though, thank you for your input! :)

PR #29, opened 2022-12-21
fix: added optional chaining in LayoutTopNavigation.vue (NTR)

niklaswolf:


patzick:
Thank you @niklaswolf!
Merged with https://github.com/shopware/frontends/commit/ee601feb0813b8b5a117cf9eaf0fad1b4d142418

sorry, we have an issue with gitlab mirroring so it's not merged via GH, we'll fix that soon but you'll be mentioned in the version changelog!

Issue #31, opened 2022-12-21
[Bug] Uncaught (in promise) SyntaxError: ambiguous indirect export: default


niklaswolf:
There seems to be an issue with the component `CmsElementText.vue`.
I get the error `Uncaught (in promise) SyntaxError: ambiguous indirect export: default` in the browser console when setting up a new demo-store template and connecting it to my shopware-instance.

<details>
<summary>Complete log is here</summary>

```
[Vue warn]: Unhandled error during execution of setup function 
  at <AsyncComponentWrapper content= 
Object { type: "text", slot: "content", block: null, blockId: "352f6d6d28b84c929c3ee8449b750fb1", config: {…}, fieldConfig: (2) […], translations: null, data: {…}, locked: false, cmsBlockVersionId: "0fa91ce3e96a4bc2be4bd9ce752c3425", … }
 style= 
Object { backgroundColor: null, backgroundImage: null, backgroundSize: null, sizingMode: null, marginBottom: null, marginLeft: null, marginRight: null, marginTop: null }
 class="" > 
  at <DynamicRender style= 
Object { backgroundColor: null, backgroundImage: null, backgroundSize: null, sizingMode: null, marginBottom: null, marginLeft: null, marginRight: null, marginTop: null }
 class="" > 
  at <CmsGenericElement content= 
Object { type: "text", slot: "content", block: null, blockId: "352f6d6d28b84c929c3ee8449b750fb1", config: {…}, fieldConfig: (2) […], translations: null, data: {…}, locked: false, cmsBlockVersionId: "0fa91ce3e96a4bc2be4bd9ce752c3425", … }
 style= 
Object { backgroundColor: null, backgroundImage: null, backgroundSize: null, sizingMode: null, marginBottom: null, marginLeft: null, marginRight: null, marginTop: null }
 class="" > 
  at <CmsBlockText content= 
Object { type: "text", slots: (1) […], sectionId: "f3c4165109fb453b83448f6c55f297d3", position: 2, name: "category-description", sectionPosition: "main", marginTop: null, marginBottom: null, marginLeft: null, marginRight: null, … }
 style= 
Object { backgroundColor: null, backgroundImage: null, backgroundSize: null, sizingMode: null, marginBottom: null, marginLeft: null, marginRight: null, marginTop: null }
 class="" > 
  at <AsyncComponentWrapper content= 
[...]

PR #40, opened 2023-01-03
[#31]: Add brackets to import, to prevent "ambiguous indirect export:…

schipperjeroen:
Solves issue #31


patzick:
hey @schipperjeroen 
thank you for your PR, unfortunately this solution causes another error, we already tried this some time ago. `html-to-vue` is a cjs export and that's why even `@ts-ignore` directive is there.
I'll take a closer look this week and will get back to you!

schipperjeroen:
Hi @patzick ,

Updated the import with how the readme of `html-to-vue` does is. You're correct about the `@ts-ignore`. Added that also. What is the other error? Couldn't see any errors in console or the watcher?

patzick:
hey @schipperjeroen!
Thank you very much for contributing! @mdanilowicz moved the code temporarily to our repo to avoid this error, but we'll be looking for a better solution in future!
Closing this one, but we're grateful for your input! 🙌

PR #53, opened 2023-02-02
Implement optional browser endpoint for the api client

GM-Alex:
Axios supports a browser base url which should be also configurable via the plugin settings.


patzick:
Closing this for now, it's available in community module for middleware proxy: https://frontends.shopware.com/resources/community-modules/

PR #61, opened 2023-03-10
fix(types): Type fixes

niklaswolf:
- adjusted customFields type in every entity that used it
- added customFields to ListingFilter-type
- adjusted OrderLineItem type


patzick:
thanks @niklaswolf !
Merged with: https://github.com/shopware/frontends/commit/3a90d299279b451e391a946dafecc857fe1f67fc

PR #71, opened 2023-03-27
feat: sitemap

patzick:


BrocksiNet:
Tested and works nice :+1: 

There is only one thing that looks wrong (copied from sitemap.xml.gz):
```
<url><loc>https://demo-frontends.shopware.store//Summer-Trends/</loc><lastmod>2022-10-18</lastmod><changefreq>daily</changefreq><priority>0.5</priority></url>
```

We have two slashes `//` inside the <loc> but this could also be related to the admin domain config. Maybe verify this.

mdanilowicz:
> Tested and works nice 👍
> 
> There is only one thing that looks wrong (copied from sitemap.xml.gz):
> 
> ```
> <url><loc>https://demo-frontends.shopware.store//Summer-Trends/</loc><lastmod>2022-10-18</lastmod><changefreq>daily</changefreq><priority>0.5</priority></url>
> ```
> 
> We have two slashes `//` inside the `<loc>` but this could also be related to the admin domain config. Maybe verify this.

Looks like a backend problem. I will create a new backend task for that

patzick:
@mdanilowicz @BrocksiNet, so let's have a backend issue for that, as for this scope we'll merge it. Good work Maciek!

PR #73, opened 2023-03-27
feat: new dynamic api client

patzick:
Provides a base for the new API client approach.


BrocksiNet:
We have some lint warnings and errors we should fix.
You can see them here: https://github.com/shopware/frontends/actions/runs/4879719159/jobs/8706585185

PR #82, opened 2023-03-27
feat: double opt-in registration (SWF-215)

mdanilowicz:


packages/composables/src/useUser.ts
```diff
@@ -170,7 +170,7 @@ export function useUser(): UseUserReturn {
       apiInstance
     );
     _user.value = customer;
-    await refreshSessionContext();
+    if (_user.value?.active) await refreshSessionContext();
```
patzick, line 173: we need a changeset entry for composables logic change

patzick (changes requested):
All good, changelog entry needed

patzick (approved):

PR #86, opened 2023-03-28
feat(helpers): extract navigation URL

mkucmus:


packages/helpers/src/category/getCategoryUrl.ts
```diff
@@ -1,26 +1,44 @@
...
  * Get URL for category.
  * Some link {@link isLinkCategory}
  *
  * @param {Category} category category entity
  *
- * @beta
+ * @public
  */
-export const getCategoryUrl = (category: Partial<Category>): string => {
+export function getCategoryUrl(category: LinkedCategory): string {
```
patzick, line 30: we need unit tests for that, especially that's becoming a `public` method

patzick (changes requested):
Can you create `getCategoryUrl.test.ts` file to test this helper method?

patzick (approved):
👌

Issue #77, opened 2023-03-27
vue-demo-store locally in production mode


patzick:
Check whether is possible to run the vue-demo-store template in production mode for node-ssr nitro preset.

Acceptance criteria:
check preset for default nitro server (node) for building and running
add description how to run it in prod mode under “production” chapter of templates/vue-demo-store/README.md

PR #90, opened 2023-03-28
feat(templates): Production build example for vue-demo-store

mkucmus:
closes: #77


mkucmus:
Hey @patzick, I'm also thinking of changing the default production preset from `vercel` to `node-server` to be more local friendly: 
https://github.com/shopware/frontends/pull/90/files#diff-037c0c63ce96c99947956155933fa492ca0022f8f2f42aeb5245de8528a41053R2

in this case we could simplify the way how the template is built (we say: run build and start and that's it) without changing a preset. 
For our purposes we could change the .env.production or env itself in the build time to be a `vercel`.

patzick (approved):
Love this!
Let's create another PR on top to change the default`NITRO_PRESET`, then we'll set `vercel` on environment variables only for preview building and CI will ensure us whole thing is working properly :)

PR #95, opened 2023-03-30
fix(demo): update useAuthGuard, redirects

BrocksiNet:
Changes

- If you logged in and try to reach the login page you are redirected to account
- If you logged in and try to reach the register page you are redirected to account
- After log in from login page you are redirected to account instead of homepage
- Params can be used for account page in combination with useAuthGuardRedirection
- useAuthGuardRedirection will only execute if the routing is on client side (on server side the value of isLoggedIn is not correct)

### DIFF
--- a/templates/vue-demo-store/components/account/AccountMenu.vue
+++ b/templates/vue-demo-store/components/account/AccountMenu.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import { SharedModal } from "~~/components/shared/SharedModal.vue";
-const { isLoggedIn, logout, user, refreshUser } = useUser();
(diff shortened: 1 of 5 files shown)


patzick (approved):

.changeset/clean-glasses-walk.md
```diff
@@ -0,0 +1,6 @@
+---
+"vue-demo-store": minor
+"@shopware-pwa/cms-base": minor
```
patzick, line 3:
```suggestion
```
this is not the change in cms package, we wouldn't want that kind of logic there :)

patzick (approved):

Issue #94, opened 2023-03-29
Missing regulationPrice


mattbra1155:
In the docs there is a mention of 

![image](https://user-images.githubusercontent.com/50120402/228529296-c5c558b5-7caf-45be-a221-2303281536e4.png)

Regulation price is received from useProductPrice -> price call and product.calculatedPrice

![image](https://user-images.githubusercontent.com/50120402/228528438-0538e758-9217-4f69-b844-2c4a49085baa.png)

but doesn't appear in CalculatedPrice type

![image](https://user-images.githubusercontent.com/50120402/228528754-0266d046-38e6-43b0-98b8-22a892620955.png)

therefore we can't use it

PR #101, opened 2023-04-03
feat(cms-base): regulation price displaying

mkucmus:
closes: #94 

- [x] types enhancement
- [x] cms-base package adjusted
- [x] vue-demo-store adjusted
- [x] tier prices condition to avoid wrong price displaying
- [x] added missing tier prices on pdp built in cms-page 
- [x] check for other occurrences when regulation price can be displayed 
- [x] report found issues in the core

### DIFF
--- a/packages/cms-base/components/SwListingProductPrice.vue
+++ b/packages/cms-base/components/SwListingProductPrice.vue
@@ -9,7 +9,7 @@ const { product } = toRefs(props);
 
 const { price, unitPrice, displayFromVariants, displayFrom, isListPrice } =
   useProductPrice(product);
-const regulationPrice = computed(() => price?.value?.regulationPrice?.price)
+const regulationPrice = computed(() => price.value?.regulationPrice?.price)
 
 </script>
 
--- a/packages/cms-base/components/SwListingProductPrice.vue
+++ b/packages/cms-base/components/SwListingProductPrice.vue
@@ -9,6 +9,8 @@ const { product } = toRefs(props);
 
 const { price, unitPrice, displayFromVariants, displayFrom, isListPrice } =
   useProductPrice(product);
+const regulationPrice = computed(() => price?.value?.regulationPrice?.price)
+
 </script>
 
 <template>
@@ -38,5 +40,8 @@ const { price, unitPrice, displayFromVariants, displayFrom, isListPrice } =
         ><span v-if="displayFrom || displayFromVariants">to</span></template
       >
     </SharedPrice>
+    <div class="text-xs flex text-gray-500" v-if="regulationPrice">
+      Previously <SharedPrice class="ml-1" :value="regulationPrice"/>
+    </div>
   </div>
 </template>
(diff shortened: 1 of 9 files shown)


packages/cms-base/components/SwListingProductPrice.vue
```diff
@@ -9,6 +9,8 @@ const { product } = toRefs(props);
 
 const { price, unitPrice, displayFromVariants, displayFrom, isListPrice } =
   useProductPrice(product);
+const regulationPrice = computed(() => price?.value?.regulationPrice?.price)
```
patzick, line 12:
```suggestion
const regulationPrice = computed(() => price.value?.regulationPrice?.price)
```
one too many :)

packages/cms-base/components/public/cms/element/CmsElementBuyBox.vue
```diff
@@ -22,19 +18,10 @@ const { product, changeVariant } = useProduct(
...
-    );
-  } else {
-    return null;
-  }
-});
-const listPrice = computed(() =>
-  getProductCalculatedListingPrice(product.value)
-);
+const { unitPrice, price, tierPrices, isListPrice } = useProductPrice(product);
+const regulationPrice = computed(() => price?.value?.regulationPrice?.price)
```
patzick, line 22:
```suggestion
const regulationPrice = computed(() => price.value?.regulationPrice?.price)
```

packages/types/shopware-6-client/models/checkout/cart/price/CalculatedPrice.d.ts
```diff
@@ -19,4 +19,8 @@ export type CalculatedPrice = {
     percentage: number;
     apiAlias: string;
   };
+  regulationPrice: {
```
patzick, line 22: Does it mean it's always visible if calculated prices are visible?

packages/types/shopware-6-client/models/framework/pricing/Price.d.ts
```diff
@@ -1,6 +1,10 @@
 export type Price = {
-  currencyId: number;
   net: number;
   gross: number;
   linked: boolean;
+  listPrice: null;
+  currencyId: number;
+  regulationPrice: Partial<Price>;
```
patzick, line 7: not the same as in `CalculatedPrice`?

patzick (changes requested):
Looks good, I have some questions for typing :)

if the regular price is always if calculatedPrice exist we could improve:

```ts
const regulationPrice = computed(() => price.value?.regulationPrice.price)
```

or even we can use it without computed directly in the template as it's not something which is reactive
```html
<div class="text-xs flex text-gray-500" v-if="price?.regulationPrice">
    Previously <SharedPrice class="ml-1" :value="price.regulationPrice.price"/>
</div>
```

packages/types/shopware-6-client/models/framework/pricing/Price.d.ts
mkucmus: It's a different place and format :)

patzick (approved):

PR #108, opened 2023-04-05
fix(demo-template): open graph site name default value

patzick:
fixes default value for `og:site_name`


mkucmus (approved):
🥇 for making it dynamic!

PR #135, opened 2023-04-13
chore: add stackblitz preview links to PR

mkucmus:
adds stackblitz links to the PR's comment for templates preview easy access:

image


patzick:
Hmm, this is quite cool but might be misleading. As in PR you probably change composable or something in another package, then Stackblitz example won't work even if PR is correct. 
We would need to either:
- have a whole repo in PR template so it would use workspace packages and then display specific template
- publish every PR as a package and then change it in template dependency in runtime to use it

do you see any other solutions?

PR #137, opened 2023-04-13
fix(composables): safe check for locale on SSR

mkucmus:


packages/composables/src/usePrice.ts
```diff
@@ -34,7 +34,11 @@ export function usePrice(): UsePriceReturn {
     currencyCode: string;
   }): void {
     _setCurrencyCode(params.currencyCode);
-    _setLocaleCode(params.localeCode || navigator?.language);
+    _setLocaleCode(
+      params.localeCode ||
+        (typeof navigator !== "undefined" && navigator?.language) ||
+        "en-US",
```
patzick, line 40:
```suggestion
      params.localeCode || navigator?.language || "en-US",
```
Why not this?

patzick (changes requested):

packages/composables/src/usePrice.ts
mkucmus: accessing `navigator` will not work in SSR as the global variable is not available in node, but only in the browser, the good example is enforced SSR environment in [astro](https://stackblitz.com/github/shopware/frontends/tree/main/templates/astro) for example.

```
ReferenceError: navigator is not defined
```

patzick (approved):
Great stuff!

Issue #131, opened 2023-04-12
(Docu) ​Example: Footer


BrocksiNet:
We are missing an Example for **Footer** here:
https://shopware-frontends-docs.vercel.app/resources/examples/

- [ ] Add example image to overview page
- [ ] Create and test the footer example
- [ ] Add working example to sup-page

PR #153, opened 2023-04-19
docs: footer navigation example

mkucmus:
### Description

closes: #131 

### Type of change

image


pnpm-lock.yaml
```diff
@@ -904,13 +904,6 @@ packages:
       '@algolia/requester-common': 4.17.0
     dev: false
 
-  /@ampproject/remapping@2.2.0:
```
patzick, line 907: Why has the package lock changed in this PR? No dependency changes. Let's revert this file

patzick (changes requested):
Looks good. Let's just revert package lock changes

pnpm-lock.yaml
mkucmus: I rerun install on original lock and the change is gone now

patzick (approved):

Issue #157, opened 2023-04-20
[BUG] useCartItem composable returns invalid values


ignasradzius:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Hello, currently itemRegularPrice and itemSpecialPrice in useCartItem composable returns the same single unitPrice value.

<img width="937" alt="Screenshot 2023-04-20 at 14 57 36" src="https://user-images.githubusercontent.com/38155746/233359108-e3fc5977-4add-488c-a1cf-4f848c492d09.png">

### Expected Behavior

Since cartItem can have different itemQuantities:
1. I expect that itemRegularPrice will return the price of unit * itemQuantity. 
2. I expect that itemSpecialPrice will return discounted price of unit * itemQuantity.

mkucmus:
hey @ignasradzius, thanks for reporting the issue 💙 

I've added new property named `itemTotalPrice` within #168 which is responsible now for keeping the total price for quantities of a product added to the cart.

I left the regular and special (discounted) price as it was because we don't wont to do any calculations on browser side due to known issues with float and rounding numbers, especially for the prices in ecommerce. So I could only expose what comes from API, as it's our single source of truth. The precision and other currency settings can be also set in admin panel, so it would be hard to mainain the client and backend side - it's easy to forget 🤔 

for special price case you can always do the calculation on your as you have an access to `cartItem` object as well - using the library you can trust in that matter. 

Please leave your feedback if it's that OK 👍🏻 for you or not 👎🏻

PR #168, opened 2023-04-24
feat(composables): new total price property for useCartItem composable

mkucmus:
### Description

closes: #157 

```js
const { itemTotalPrice } = useCartItem(cartItem);
```

### Type of change

image

### DIFF
--- a/.changeset/sour-sheep-smile.md
+++ b/.changeset/sour-sheep-smile.md
@@ -2,4 +2,4 @@
 "@shopware-pwa/composables-next": patch
 ---
 
-getProductItemSeoUrlData method of useCart marked as deprecated
+`getProductItemSeoUrlData` method of `useCart` marked as deprecated


packages/composables/src/useCartItem.ts
```diff
@@ -76,19 +115,20 @@ export function useCartItem(cartItem: Ref<LineItem>): UseCartItemReturn {
   const isPromotion = computed(() => cartItem.value.type === "promotion");
 
   async function removeItem() {
-    const result = await removeCartItem(cartItem.value.id, apiInstance);
-    // broadcastUpcomingErrors(result);
-    await refreshCart();
+    await removeCartItem(cartItem.value.id, apiInstance);
+    refreshCart();
```
patzick, line 119: cart operation is returning cart, what would you say to do something like this?
```suggestion
    const newCart = await removeCartItem(cartItem.value.id, apiInstance);
    await refreshCart(newCart);
```

and then in useCart
```ts
async function refreshCart(newCart?: Cart): Promise<Cart> {
    if (newCart) {
      _storeCart.value = newCart;
      return newCart;
    }
    const result = await getCart(apiInstance);
    _storeCart.value = result;
    return result;
  }
```
so we can skip additional server request?

packages/composables/src/useCartItem.ts
```diff
@@ -76,19 +115,20 @@ export function useCartItem(cartItem: Ref<LineItem>): UseCartItemReturn {
...
   }
 
   async function changeItemQuantity(quantity: number): Promise<void> {
-    await changeProductQuantity({
+    changeProductQuantity({
```
patzick, line 123:
```suggestion
    await changeProductQuantity({
```
why are we removing that? Then method is complete before the whole action is

packages/composables/src/useCartItem.ts
```diff
@@ -76,19 +115,20 @@ export function useCartItem(cartItem: Ref<LineItem>): UseCartItemReturn {
...
       id: cartItem.value.id,
       quantity: quantity,
     });
-    // broadcastUpcomingErrors(result);
   }
 
+  /**
+   * @deprecated Method is not used anymore and the case should be solved on project level instead due to performance reasons.
```
patzick, line 130: let's add patch changeset entry for that

templates/vue-demo-store/components/checkout/CheckoutCartItem.vue
```diff
@@ -1,51 +1,51 @@
 <script setup lang="ts">
-import { getSmallestThumbnailUrl } from "@shopware-pwa/helpers-next";
-import { LineItem } from "@shopware-pwa/types";
+  import { getSmallestThumbnailUrl } from "@shopware-pwa/helpers-next";
+  import { LineItem } from "@shopware-pwa/types";
```
patzick, line 3: script setup components have formatting from the first line without tabs, can you make sure you formatted this file? :)

patzick (changes requested):
Looks really good, some suggestions left for improvements

templates/vue-demo-store/components/checkout/CheckoutCartItem.vue
mkucmus: idk why but I had a prettier as default formatter. Switched back to volar and works good as expected :)

packages/composables/src/useCartItem.ts
mkucmus: good improvement. I will do 🚀

.changeset/sour-sheep-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-pwa/composables-next": patch
+---
+
+getProductItemSeoUrlData method of useCart marked as deprecated
```
patzick, line 5:
```suggestion
`getProductItemSeoUrlData` method of `useCart` marked as deprecated
```

patzick (approved):

PR #165, opened 2023-04-24
feat: poc modal teleport

patzick:
### Description

another view of #148 

- removed ClientOnly issue
- no need for provide/inject
- ability to create shared modals
- ability to create multiple modals

Preview link:
https://stackblitz.com/github/shopware/frontends/tree/feat/poc-modal-teleport/examples/modal-teleport?file=pages%2Findex.vue

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

PR #148, opened 2023-04-17
feat: example modal and teleport

BrocksiNet:
### Description

closes: #64

### Type of change

- [X] Example Implementation

### Additional context

Look at the Readme in this PR.

PR #169, opened 2023-04-24
feat(demo): new modals strategy with useModal and Teleport

### LINKED ISSUE #64: Most composables can not be used in modal or other injected overlays (vue-demo-store)
Hi,
I'll try to explain my problem with a specific use case, but I'm sure there are many others.

I moved the listing filters in an overlay, but to my surprise the filters were not there anymore.
The reason for this should be the new useListing composable call. I think the inject in the composable is not working, because the overlay itself is higher up in the component tree (in the layout).

I also noticed the useAddress composable is called in the app.vue. This way the address modals in the account are working correctly because the highest provide is happening in the app.vue. If this global call for useAdress is removed it is not working anymore.

And the global call is not working for the useListing composable and anyways i think it's not a good idea to call every composable in the app which has an overlay somewhere.

Do you have any tipps for me how this problem could be solved? I think for most of the use cases an overlay is used to move functionality inside the overlay and so it is important that the composables can be used there.

patzick:
### Description

closes #64

- improved Modals with Teleport, related to #148 and #165
- added transitions for modals and sidebars

### Type of change

### DIFF
--- a/packages/composables/src/useOrderDetails.ts
+++ b/packages/composables/src/useOrderDetails.ts
@@ -7,7 +7,7 @@ import {
   PaymentMethod,
   ShopwareSearchParams,
   OrderDocument,
-  ShopwareAssociation
+  ShopwareAssociation,
 } from "@shopware-pwa/types";
 import {
   cancelOrder,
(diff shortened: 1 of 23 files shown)


mkucmus (approved):
nice combination with teleport 👍🏻

packages/composables/src/useOrderDetails.ts
```diff
@@ -192,7 +195,7 @@ export function useOrderDetails(orderId: string, associations?: ShopwareAssociat
   async function loadOrderDetails() {
     const orderDetailsResponse = await getOrderDetails(
       orderId,
-      deepMerge(orderAssociations, associations? associations: {}),
+      deepMerge(orderAssociations, associations ? associations : {}),
```
mdanilowicz, line 198: `deepMerge(orderAssociations, associations),`

DeepMerge checks values

mdanilowicz (approved):

packages/composables/src/useOrderDetails.ts
patzick: This is only a formatter change, no logic is touched here

PR #175, opened 2023-04-26
feat: add husky and lint-staged pre-commit hook

### LINKED ISSUE #93: Fix lint-staged workflow
Related to #85 

We use lint-staged, but checking out the repository is not create a git hook which runs this command.

### Acceptance criteria
1. after checkout and init repository, git gook for lint-staged is set

### Additional note (nice to have)
1. Lint command is also running prettier over linted files, so there is no local difference to see

BrocksiNet:
### Description

This PR will add husky as dev package and make sure that after `pnpm install` husky is initialized and the pre-commit hook is installed.

closes #93 

### Type of change

- [X] New feature (non-breaking change which adds functionality)

### Additional context

If you have problems during commit or an error is thrown in vscode that `npx command is not found` you maybe need to do [this](https://typicode.github.io/husky/#/?id=command-not-found).


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npx lint-staged
```
patzick, line 4:
```suggestion
pnpm run lint-staged
```
wouldn't this be better? We have locally installed `lint-staged`

package.json
```diff
@@ -21,7 +21,8 @@
     "version": "changeset version",
     "postversion": "pnpm run versionReadme",
     "versionReadme": "pnpx ts-node scripts/addChangelogBodyToReadmeFile.ts",
-    "taze": "taze major -wr"
+    "taze": "taze major -wr",
+    "prepare": "husky install"
```
patzick, line 25:
```suggestion
    "postinstall": "husky install"
```
this would do this automatically without need to remember about running `pnpm run prepare`, the question is how much time does it take to run install even if it's already installed?

patzick (changes requested):

.husky/pre-commit
BrocksiNet: It is not a script. At least for me with run it was not working. And lint-staged is not below scipts in packages.

package.json
BrocksiNet: husky documentation suggest prepare, also prepare is executed during install, no need to execute pnpm run prepare manual

patzick: nice!

patzick (approved):

PR #194, opened 2023-05-05
feat: use history state in links to resolve seoUrls

patzick:
### Description

based on the excellent presentation from @niklaswolf regarding the SEO URL fetching problem while navigating on the client side.
The solution reduces internal navigation by almost half.

Changes:
- added new URL helpers. Instead of the `string` we can leverage object and [History State API](https://developer.mozilla.org/en-US/docs/Web/API/History/state)
- while resolving URL in the generic `[...all].vue` component, we're checking the state, if there is a value, we already know what resource we need to fetch - no `seo-url` invocation [code](https://github.com/shopware/frontends/pull/194/files#diff-bfbe7337daadfff45e9a82a5f96065d567fc92e77619c3996981c391cef3d292R25)

Things to complete are added as comments in the code to be resolved.
Early feedback is welcomed.

### Type of change

#### Move from homepage to Summer BBQ
Before: 
image

After:
image


packages/helpers/src/category/getCategoryUrl.ts
```diff
@@ -45,3 +45,27 @@ export function getCategoryUrl(category: LinkedCategory): string {
         : `/${getEntityPrefix(category)}/${category.id}`;
   }
 }
+
+// TODO: move to separate file and add tests
```
patzick, line 49: As in comment
TODO: move to separate file and add tests

packages/helpers/src/category/getCategoryUrl.ts
```diff
@@ -45,3 +45,27 @@ export function getCategoryUrl(category: LinkedCategory): string {
...
+  return {
+    path: getCategoryUrl(category),
+    state: {
+      routeName: "frontend.navigation.page",
+      foreignKey: category?.id,
+    },
+  };
+}
+
+// TODO: move to separate file and add tests
```
patzick, line 60: As in comment
TODO: move to separate file and add tests

packages/helpers/src/category/getCategoryUrl.ts
```diff
@@ -45,3 +45,27 @@ export function getCategoryUrl(category: LinkedCategory): string {
...
+export function getNavigationRoute(navigationElement: LinkedCategory) {
+  if (navigationElement.linkType == "product") {
```
patzick, line 62: This is a specific navigation element. You can see that in menu position CMS -> Product with tier prices.
image

it links directly to the product page, not to the category page like any other menu link
@niklaswolf did you encounter such case?

packages/helpers/src/product/getProductUrl.ts
```diff
@@ -9,8 +9,26 @@ import { Product } from "@shopware-pwa/types";
...
+    id: string;
+    seoUrls: SeoUrl[] | null;
+  }
+>(product?: T): string {
   if (!product) return "/";
   const seoUrl = product.seoUrls?.[0]?.seoPathInfo;
   return seoUrl ? `/${seoUrl}` : `/detail/${product.id}`;
 }
+
+// TODO: move to separate file and add tests
```
patzick, line 23: As in comment
TODO: move to separate file and add tests

packages/helpers/src/category/getCategoryUrl.ts
niklaswolf: No :D didn't even know that this is possible.
I assume there is no reference to the `productId` in the `navigationElement`? If there were, we of course could easily provide the information to the state object.

niklaswolf:
@patzick woah, I like this approach a lot, even more than my initial approach, because it is framework agnostic and uses browser features. That's why you should have conversations with other people, I did not think the idea to the end in my euphoria that it worked so well :D 

I have still one question (because I haven't tried it): do these state-objects in the links also hydrate to the client side when in a SSR context?

packages/helpers/src/category/getCategoryUrl.ts
patzick: There is a `type` of navigation element. It's now as it should :)

packages/helpers/src/category/getCategoryUrl.ts
patzick: done

packages/helpers/src/category/getCategoryUrl.ts
patzick: done

packages/helpers/src/product/getProductUrl.ts
patzick: done

patzick:
@niklaswolf I very much appreciate that collaboration, great stuff that we can push each other to create better solutions!

I completed all TODOs, added tests and additional conditions for different possible link types in main navigation routing, should be merged soon after review :)

As for your question - it renders the same way from SSR and then on client side additional data are passed during hydration. The link will work either way, as there is always a fallback for checking the `seo-url` link. Testing it by browsing through demo website the `seo-url` endpoint is not invoked anywhere

packages/cms-base/components/SwVariantConfigurator.vue
```diff
@@ -50,10 +51,12 @@ const onHandleChange = async () => {
...
     unref(selectedOptions)
   );
-  const selectedOptionsVariantPath = variantFound?.seoUrls?.[0]?.seoPathInfo;
+  // const selectedOptionsVariantPath = variantFound?.seoUrls?.[0]?.seoPathInfo;
+  const selectedOptionsVariantPath = getProductRoute(variantFound);
   if (props.allowRedirect && selectedOptionsVariantPath) {
     try {
-      router.push("/" + selectedOptionsVariantPath);
+      console.error("pushing route", selectedOptionsVariantPath);
+      router.push(selectedOptionsVariantPath);
```
mkucmus, line 59: little cleanup and will be cool 🧹

.changeset/itchy-hats-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-pwa/helpers-next": minor
+---
+
+Added `getCategoryRoute` and `getProductRoute` helpers to create router lins with history API metadata
```
mkucmus, line 5: How does it work with router hash mode if someone would need to change it (donno why)? does it have anything to do with the implementation?

mkucmus (commented):
Nice implementation! 🥇 

I think it would be nice to have it explained also in the 📔 -> https://frontends.shopware.com/getting-started/routing.html

as an alternative or one of possible ways how to deal with routes in vue app.

.changeset/itchy-hats-ring.md
patzick: Haven't tested it but it shouldn't affect it at all, but then someone probably will build links in another way. The mechanism should work though

packages/helpers/src/category/getCategoryRoute.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Returns category or navigation URL and route informations for resolving SEO url.
+ * Use it with combination of <RouterLink> or <NuxtLink> in Vue.js or Nuxt.js projects.
+ *
+ * Example:
+ * ```html
+ * <RouterLink :to="getCategoryRoute(navigationElement)">
+ * ```
+ */
```
mkucmus, line 16: maybe we can categorize the method to some existing groups: https://frontends.shopware.com/packages/helpers.html (see the side navi)

.changeset/itchy-hats-ring.md
mkucmus: ok it's fine, it's just a template with proposed implementation. We will see if someone faces a problem with that or not 👍🏻

mdanilowicz (approved):

templates/vue-demo-store/pages/[...all].vue
```diff
@@ -21,6 +21,15 @@ const route = useRoute();
 const { data: seoResult } = await useAsyncData(
   "cmsResponse" + route.path,
   async () => {
+    // For client links if the history state contains seo url information we can omit the api call
+    if (process.client) {
```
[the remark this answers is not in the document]
Comment (patzick) [reply templates/vue-demo-store/pages/[...all].vue:25]: needed to be in nested if, as `history` is not existing on server side and there were problems with that

Issue #147, opened 2023-04-17
Reduce unneccessary API call on client side for useListing


niklaswolf:
I'm wondering how to use `useListing` in a SSR-Environment (in this case Nuxt3).
Because the internal state of the composable regarding the `initialListing` ist stored in `_storeInitialListing`, which **can** be injected. But because this state doesn't get hydrated on the client side, the API-call gets made another time on the client side, which is not necessary if the state could be hydrated.
https://github.com/shopware/frontends/blob/main/packages/composables/src/useListing.ts#L277 

So I'm wondering if there is a clever way to make this work nicely with SSR :)

BrocksiNet:
Hi @niklaswolf 👋 

I created a [branch](https://github.com/shopware/frontends/tree/feat/listing-with-astro) with this [commit](https://github.com/shopware/frontends/commit/c03669f92c9cde7ed038adda9fadea1487185ac8) and tested `useListing` inside the Astro template.

You can try it with:
`git checkout feat/listing-with-astro`
`pnpm run dev --filter=shopware-astro `

So I guess your question is not about SSR in general ... more like how we could reduce this one not needed call that is done on client side again?

Did you already tried with Nuxt `useHydration` [(see Docu)](https://nuxt.com/docs/api/composables/use-hydration)?

BrocksiNet:
Hi @niklaswolf, I was researching today again about this topic. 

Some statements to have us in the same context:
- We need the SSR call for SEO reasons (google bot etc.)
- We need the client-side calls because users want to filter and elements change during client-side usage (navigation)
- So we could save one call that takes between 200ms and 400ms on the first "hard" page load **AND** only when a filter is already set (!) [I measured the setInitialListing function]

The main problem is this part here (_packages/composables/src/useListing.ts_):
```
const allFiltersResult = await searchMethod({
  query: initialListing.currentFilters.search || undefined,
});
initialListing = Object.assign({}, initialListing, {
  aggregations: allFiltersResult?.aggregations,
});
```

Maybe I do not see an easy solution at the moment and will discuss this again with the team.

PS: I was able to manage the data-passing from server to client but then I lost the reactivity. And it is not easy to do this in package cms because some NUXT stuff is missing and the package should work also without NUXT.

BrocksiNet:
Thx to the great @mkucmus he found out that this is totally not need anymore. 🎉 
See his PR ➡ https://github.com/shopware/frontends/pull/255

niklaswolf:
[...]

PR #255, opened 2023-06-14
fix(composables): remove redundant api call within useListing search

mkucmus:
### Description

it turned out that the second call for all aggregation is not needed anymore. therefore it can be removed.

closes: #147 

### Type of change


packages/composables/src/useListing.ts
```diff
@@ -267,20 +267,7 @@ export function createListingComposable<ELEMENTS_TYPE>({
...
-    ) {
-      loading.value = true;
-      const allFiltersResult = await searchMethod({
-        query: initialListing.currentFilters.search || undefined,
-      });
-      initialListing = Object.assign({}, initialListing, {
-        aggregations: allFiltersResult?.aggregations,
-      });
-    }
+    loading.value = true;
```
mdanilowicz, line 270: Do we have to change a loader status if we only assign value in this function? This will be done in the same tick

patzick, line 270:
```suggestion
```

packages/composables/src/useListing.ts
```diff
@@ -267,20 +267,7 @@ export function createListingComposable<ELEMENTS_TYPE>({
...
     _storeInitialListing.value = initialListing;
     _storeAppliedListing.value = null;
     loading.value = false;
```
patzick, line 273:
```suggestion
```

packages/composables/src/useListing.ts
patzick: +1

patzick (approved):
Looks great, just one suggestion from @mdanilowicz to apply and we can merge 👌

Issue #262, opened 2023-06-15
[BUG] Clear wishlist for logged in users


kstala:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

As a logged in user, the clear wishlist option doesn't work

https://github.com/shopware/frontends/assets/30267600/83ca77a9-17ab-4c45-8db7-fef6c9d46356

`{"errors":[{"code":"0","status":"405","title":"Method Not Allowed","detail":"No route found for \u0022GET https:\/\/demo-frontends.shopware.store\/store-api\/customer\/wishlist\/delete\/2623b5f974f3438a850254c86b758478\u0022: Method Not Allowed (Allow: DELETE)"}]}`

### Environment

```markdown
- OS:
- Node:
- pnpm:
```

BrocksiNet:
There is much more broken. I can not reproduce this exact error, but I can produce other errors that are mostly only there when you are logged in.

patzick:
Reopening as PR was reverted

PR #267, opened 2023-06-16
fix(demo): wishlist, move funciton calls

BrocksiNet:
### Description

closes: #262

### Type of change

### Additional context

The main problem was that after removing a product from the wishlist the items were not updated. All other errors were consequential problems because of that also the wrong requests to the API when products were already deleted.


patzick (approved):
🎯 💯

PR #270, opened 2023-06-16
feat(vue-demo):  create state and country component

### LINKED ISSUE #245: [FEATURE] States in Checkout
### Description

Currently, we do not support the option that States are required in our demo-store checkout.

### Use Case

The checkout form should be so flexible that it displays a select box when states are required depending on country.
So if you change the country you also need to check if states are required or not and display the new select.

### Proposed Solution

For me, this is like a default function that should be there and at the moment it looks a bit like we "fake" the checkout with some config (all states not required) we set via the admin instead of reacting on the config via frontends/templates.

### Alternatives Considered

We can also decide to not implement this in the demo store. 
But then we maybe should explain how to do this because it is a common use-case. 

### Additional Context

This should work for guests and logged-in users.
Decide if we always display the state select and just make it required or not or if we hide it if it is optional.

mdanilowicz:
### Description
 closes #245 

### Type of change


templates/vue-demo-store/components/shared/SharedCountryStateInput.vue
```diff
@@ -0,0 +1,112 @@
...
+  localState.value = "";
+  emit("update:stateId", "");
+  emit("update:countryId", target.value);
+};
+
+const handleStateUpdate = (e: Event) => {
+  const target = e.target as HTMLInputElement;
+  localState.value = target.value;
+  emit("update:stateId", target.value);
+};
```
patzick, line 43: using useVModels would make it more clean without keeping additional local state https://vueuse.org/core/useVModels/#usevmodels
```suggestion
const { countryId, stateId } = useVModels(props, emit)

const { getCountries, getStatesForCountry } = useCountries();

const states = computed(() => {
  return getStatesForCountry(countryId.value);
});

watch(countryId, () => {
  stateId.value = '';
}) // maybe with 'immediate'
```

templates/vue-demo-store/components/shared/SharedCountryStateInput.vue
```diff
@@ -0,0 +1,112 @@
...
+      }}</label>
+      <select
+        id="country"
+        v-model="localCountry"
+        required
+        name="country"
+        autocomplete="country-name"
+        class="mt-1 block w-full p-2.5 border border-gray-300 text-gray-900 text-sm rounded-md shadow-sm focus:ring-brand-light focus:border-brand-light"
+        data-testid="country-select"
+        @change="handleCountryUpdate"
```
patzick, line 60: then input handling is also easier
```suggestion
        v-model="countryId"
        required
        name="country"
        autocomplete="country-name"
        class="mt-1 block w-full p-2.5 border border-gray-300 text-gray-900 text-sm rounded-md shadow-sm focus:ring-brand-light focus:border-brand-light"
        data-testid="country-select"
```

patzick (changes requested):
Nice work. Overall I'd go for extracting whole `SharedAddressForm`/`SharedAddressFields`, as the API of this component would be much simpler and validation rules would be handled in a single place as well instead of being passed as props. Additionally I left comment for simplification of data passing. Let me know what do you think about moving all address fields to single component

patzick:
- component watchers are standard Vue3 solution - simple and reactive way to do something in response for a change in other reactive value(s)
- code is much cleaner, VueUse is now practically official set of composables for Vue3 + a lot of stuff from it is adapted into the Vue3 itfelf. We should use it as much as we can, so let's do it
- one thing before "magic" - storing additional local state is almost always ending with problems, when the data is changed in the parent (for example, some late loading/refresh/change address to edit) then suddenly your data is wrong as you have two sources of truth
- the "magic" is standard Vue way (same as first point) and the code is (again) much cleaner with proper responsibility
  - `v-model` is changing field value in parent
  - `watcher` for country value is resetting state value
  - a single place where field values are saved (in parent, where the data belongs)

So I'm still standing with previous review and suggestions, I hope that now you're more convinced though :)

patzick (changes requested):
I see a pending review as the previous was dismissed, so setting it up for last comment

templates/vue-demo-store/components/shared/SharedCountryStateInput.vue
patzick: Updated suggestion after the discussion:
```suggestion
const { countryId, stateId } = useVModels(props, emit)
const { getCountries, getStatesForCountry } = useCountries();
const states = computed(() => {
  return getStatesForCountry(countryId.value);
});

function onCountrySelectChanged() {
  stateId.value = '';
}
```

templates/vue-demo-store/components/shared/SharedCountryStateInput.vue
patzick: updated suggestion
```suggestion
        v-model="localCountry"
        required
        name="country"
        autocomplete="country-name"
        class="mt-1 block w-full p-2.5 border border-gray-300 text-gray-900 text-sm rounded-md shadow-sm focus:ring-brand-light focus:border-brand-light"
        data-testid="country-select"
        @change="onCountrySelectChanged"
```

kstala (approved):

patzick (approved):
great work!

Issue #265, opened 2023-06-15
[BUG] useAddress composable is missing ShopwareSearchParams


ignasradzius:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently only `loadCustomerAddresses` function accepts `ShopwareSearchParams`. After creating, updating or deleting customer address `loadCustomerAddresses` function without parameters is called.

<img width="655" alt="Screenshot 2023-06-15 at 17 50 58" src="https://github.com/shopware/frontends/assets/38155746/4a5bef67-7bd0-494a-820f-4e6a75287d9f">

### Expected Behavior

I would like to get address list sorted or filtered after creating, updating or deleting an address. That would require for `useAddress` composable to accept `ShopwareSearchParams` and them passed to `loadCustomerAddresses`.

### Environment

```markdown
- OS:
- Node:
- pnpm:
```

mkucmus:
Hey, thanks for bringing it up 👍 

what do you think about solving this by applying a single responsibility approach here, to make the method do what they should do by its name:

`updateCustomerAddress` -> only update an address, other stuff like invoking `loadCustomerAddresses()` (refreshing) should be done manually/manifestly in the component (I know, it's very convenient).

because adding some extra params (to constructor of composable or..) for the second method in the not-directly related function might be confusing. what if we had more invocations inside, that would a really big mess If we would like to have a full control of all of them 🤔

ignasradzius:
Hey @mkucmus 👋 Yea, that sounds good 💯

BrocksiNet:
@mkucmus there is also another issue #277 related to address.
Can this be solved together? At least it sounds to me like this.

PR #297, opened 2023-07-03
fix: refresh addresses explicitly in a component

mkucmus:
### Description
closes: #265 

### Type of change


templates/vue-demo-store/components/account/AccountAddressCard.vue
```diff
@@ -55,6 +56,7 @@ const setDefaultBillingAddress = async () => {
 const removeAddress = async (addressId: string) => {
   try {
     await deleteCustomerAddress(addressId);
+    loadCustomerAddresses();
```
mdanilowicz, line 59: I think that this should be added after `try and catch` because we are catching an error related to deleting the address not loading, same with the success

mkucmus: hmmm, good point @mdanilowicz, but I also pretty sure it won't be caught because of now `await` there and it's just like "let it happen" afterwards and it was intended.  So what would you prefer better: another try catch for loadCustomAddresses in that case?

kstala (approved):

templates/vue-demo-store/components/account/AccountAddressCard.vue
mkucmus: alright, I'm putting it in finally according our discussion in slack

Issue #276, opened 2023-06-20
Remove default credentials for api-client


mkucmus:
acceptance criteria:
- https://github.com/shopware/frontends/blob/main/packages/api-client/src/settings.ts#L32-L33 has no default values
- there is a validation for client settings while creating an instance, to avoid having an empty credentials in runtime and further requests

PR #299, opened 2023-07-03
feat(api-client): remove default credentials and provide early validation

mkucmus:
### Description

closes: #276 

TODO:
- [ ] update issue about the topic
- [ ] adjust unit tests

### Type of change


packages/api-client/src/apiService.ts
```diff
@@ -33,6 +33,25 @@ export interface ShopwareApiInstance {
...
+/**
+ * Check if api-client has minimum settings set
+ *
+ * @throws {Error}
+ *
+ * @param initialConfig {ClientSettings}
+ */
+function checkMinimumConfig(initialConfig: ClientSettings): void {
+  const getMessage = (missing: string) =>
+    `${missing} was not provided. Check your api-client settings. Read more at https://github.com/shopware/frontends/issues/276`;
```
patzick, line 45: It would be better to link to docs with proper config setup information instead of the issue, which only tells to remove default config

patzick (commented):
I like this idea, one suggestion

packages/api-client/src/apiService.ts
mkucmus: currently it tells not much indeed, but I will elaborate about it more there in the issue for some reasons:
- there is no dedicated page for api-client in our docs (except reference) as it's becoming replaced by api-client-next
- not much in readme there as well

or I can put some info in the home page of a package reference in docs 🤔

BrocksiNet:
I see lot of this errors in CI / Test:
```
@shopware-pwa/composables-next:test: FAIL src/cms/useCmsElementImage.test.ts [ src/cms/useCmsElementImage.test.ts ]
@shopware-pwa/composables-next:test: Error: API Endpoint was not provided. Check your api-client settings. Read more at https://github.com/shopware/frontends/issues/276
```

Is that because of your change and do we need credentials for these tests?

PR #300, opened 2023-07-04
feat(api-client): exposed ApiRequestParams to use in predefined typings

patzick:
### Description

- exported `RequestReturnType ` and `RequestParameters` types to properly type project and allow `Predefining methods` (explained in docs)
- added documentation for api-client readme
- generated new types for 6.5.3.0 - using as defaults
- fixed new-api-client example on stackblitz - using canary version


packages/api-client-next/README.md
```diff
@@ -4,7 +4,105 @@ Welcome to `@shopware/api-client` package.
 
 For getting started documentation visit [https://frontends.shopware.com/](https://frontends.shopware.com/)
 
-Documentation specific for this package: [api-client](https://frontends.shopware.com/packages/api-client.html)
+## Installation
```
mkucmus, line 7: Please use https://github.com/shopware/frontends/blob/dbd34120a025efa20813ad631ea0b5b3f5f0fb2e/packages/composables/README.md as a template to make api-client-next's README similar to other packages :)

patzick: thanks, good suggestion, applied

mkucmus (approved):
👌🏻 🥇

PR #303, opened 2023-07-06
feat: typecheck on linting

patzick:
### Description

 Our linting process was not doing typechecking, which allowed some typing errors to come into application.

- every package now runs typecheck after lint process (or with build), so GH action will be more restrictive
- eslint rules are upgraded, so after `lint` 3 filed from cms-base are no longer showing up
- build processes of `vue3-plugin` and `types` are switched to `unbuild` for consistency


packages/api-gen/src/patches.ts
```diff
@@ -7,7 +7,12 @@ export const patches = {
      * Fixes the problem  with schema >=5.0 where ContextTokenResponse is not defined
      */
     name: "ContextTokenResponse",
-    patch: (schema: Object) => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    patch: (schema: any) => {
```
patzick, line 11: couldn't find proper type for schema, will get back to this line in another PR, but didn't want to hole whole typecheck because of that

patzick: fixed

packages/typer/package.json
```diff
@@ -15,37 +15,41 @@
...
+    "lint": "eslint src/**/*.ts* --fix --max-warnings=0 && pnpm run typecheck",
+    "typecheck": "tsc --noEmit"
+  },
+  "dependencies": {
+    "find-in-files": "^0.5.0",
+    "fs-extra": "^11.1.0",
+    "ts-dox": "^0.1.0",
+    "typedoc": "^0.24.8",
+    "vinyl": "^3.0.0",
+    "vite": "^4.3.9"
```
mkucmus, line 44: is vite still needed here?

packages/typer/build.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineBuildConfig } from "unbuild";
+
+export default defineBuildConfig({
+  entries: ["src/index"],
+  declaration: true,
+  rollup: {
+    emitCJS: true,
+    cjsBridge: true,
+  },
+  externals: ["typedoc", "fs-extra", "find-in-files", "ts-dox", "vite"],
```
mkucmus, line 10: do we need vite in typer package?

mkucmus (changes requested):
Looks huge, but still works great, or better ;) one question regarding `vite` in typer package, because you've changed build tool to unbuild

packages/typer/build.config.ts
mkucmus: nvm, it's for `Plugin` type, which typer actually produces.

mkucmus (approved):

packages/typer/package.json
patzick: for types, yes :)

PR #307, opened 2023-07-10
fix(vue-demo-store): extract lang code from accept-language header

mkucmus:
### Description

Browser can have a setting with multilanguages allowed as an accept-language header `"de,en-US;q=0.7,en;q=0.3";`
the order sticks to the priority. 
thanks to this we can find the first matching language code from the string, and then, the language code can be used by 
```
new Intl.NumberFormat('de', {
      style: "currency",
      currency: 'EUR',
    }).format(123)
'123,00 €'
```

### Type of change


BrocksiNet (approved):
Tested :+1: 

At the moment also when you link from Twitter to the demo page the page is broken. With that fix, it was working as expected. We should merge this fast.

patzick (approved):

PR #309, opened 2023-07-10
feat(vue-demo): leverage edge cache

patzick:
### Type of change
edge cache leverage:
- [x] add cache to public websites, edge cache should be hit
- [x] Ensure session context is not cached
- [x] CSR client-sensitive pages like checkout, account page

Additionaly compare lighthouse:

image

### DIFF
--- a/.changeset/dull-steaks-beam.md
+++ b/.changeset/dull-steaks-beam.md
@@ -1,0 +1,5 @@
+---
+"@shopware-pwa/nuxt3-module": minor
+---
+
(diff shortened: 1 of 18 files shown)


mkucmus (approved):
I didn't expect to have it working so fast 🚀 
+ `loading=lazy` didn't broke the LCP in overall score 🥇 

one note: please add a corresponding section in [README.md](https://github.com/shopware/frontends/blob/b95dd7bb3f72dddbe598d7d43cee91c4fd80fa0f/templates/vue-demo-store/README.md) in this or another PR 🙏🏻

PR #325, opened 2023-07-17
feat(vue-demo): add private shopwareEndpoint

### LINKED ISSUE #323: Nuxt 3: Different private & public shopwareEndpoint runtime configs
### Description

I'm currently trying to set up Shopware frontends with Docker compose. While doing so I have noticed a problem with Nuxt 3: It doesn't work within Docker, because we need the Shopware endpoint runtime config to be more dynamic.

At the moment, there are basically two context in Nuxt 3: The server & client context. The problem is that the Shopware endpoint is different based on the context. In my case it's `http://localhost:8080` on the client and `http://shopware` on the server side.

This leads to the problem that the server side requests are failing. Here is my config:

docker-compose.yml
```
version: "3"

services:
  frontend:
    build:
      dockerfile: Dockerfile
      context: ./apps/frontend
      args:
        NODE_VERSION: 18
    ports:
      - "80:3000"
      - "24678:24678"
    networks:
      - web
    environment:
      - SHOPWARE_ENDPOINT_PUBLIC=http://localhost:8080
      - SHOPWARE_ENDPOINT_PRIVATE=http://shopware
      - SHOPWARE_ACCESS_TOKEN=123456789
    volumes:
      - "./apps/frontend:/src"

  shopware:
    build:
      dockerfile: Dockerfile
      context: ./apps/shopware
    ports:
      - "8080:80"
      - "22:22"
      - "9999:9999"
      - "9998:9998"
    networks:
      - web
    environment:
      - APP_URL=http://localhost
      - XDEBUG_ENABLED=1
      - PHP_VERSION=8.1
      - NODE_VERSION=18
    volumes:
      - "./apps/shopware:/var/www/html"
```

nuxt.config.ts
```
  runtimeConfig: {
    shopware: {
      shopwareEndpoint: process.env.SHOPWARE_ENDPOINT_PRIVATE,
      shopwareAccessToken: process.env.SHOPWARE_ACCESS_TOKEN,
    },
    public: {
      shopware: {
        shopwareEndpoint: process.env.SHOPWARE_ENDPOINT_PUBLIC,
        shopwareAccessToken: process.env.SHOPWARE_ACCESS_TOKEN,
      },
    },
  },
})
```

The problem lies within `@shopware-pwa/nuxt3-module` in the `plugin.ts` file. As you can see, there is no distinction between the client & server context:
```
const instance = createInstance({
      endpoint: runtimeConfig.public.shopware.shopwareEndpoint,
      accessToken: runtimeConfig.public.shopware.shopwareAccessToken,
      timeout: runtimeConfig.public.shopware.shopwareAccessToken || 10000,
      auth: {
        username:
          "<%= options.shopwareApiClient.auth ? options.shopwareApiClient.auth.username : undefined %>",
        password:
          "<%=  options.shopwareApiClient.auth ? options.shopwareApiClient.auth.password : undefined %>",
      },
      contextToken: contextToken.value || cookieContextToken,
      languageId: languageId.value || cookieLanguageId,
    });
```

### Use Case

Docker setups are working again with Nuxt 3.

At the moment, only public Shopware shops (and thus public endpoints) are working.

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

mdanilowicz:
### Description

closes #323 

### Type of change


.changeset/gorgeous-keys-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-pwa/nuxt3-module": minor
+---
+
+Add private `shopwareEndpoint` config for SSR instance creation
```
patzick, line 5: we need some docs how to use it and where it is necessary

patzick (changes requested):
Looks good, we need some documentation for this config - how and when to use it 📖

.changeset/gorgeous-keys-crash.md
BrocksiNet: And also explain when to use what ... because we also have this one:
https://frontends.shopware.com/packages/api-client/docs/storefront-url.html#preview-instance

templates/vue-demo-store/nuxt.config.ts
```diff
@@ -3,6 +3,9 @@ import { VueDisableInputsBeforeMount } from "vite-vue-plugin-disable-inputs";
 // https://v3.nuxtjs.org/docs/directory-structure/nuxt.config
 export default defineNuxtConfig({
   runtimeConfig: {
+    shopware: {
+      shopwareEndpoint: "",
+    },
```
patzick, line 8: let's not add this to template as it's only an edge case when someone will use it, instead this "example config" can be moved to documentation as well
```suggestion
```

mdanilowicz: Without this, env will not work and may be confusing. It is easier to just work on env without changing the config file.

patzick: okay I understand, so at least comment with a link to docs section explaining why there are two shopwareEndpoint settings in the config; this would be confusing for me if I'd fetch this template for the first time :)

patzick (approved):
Great job!

PR #332, opened 2023-07-20
feat(e2e): test with different workflows

BrocksiNet:
### Description

Use one workflow as a trigger and only run e2e when the other workflow is completed. Pass target_url via cache.

### Type of change

Improvement for e2e execution only on demo-store.

### Additional context

grafik


patzick (changes requested):
I can see `trigger` is fired, but `e2e-tests` is not in checks

BrocksiNet:
This always shows the last run of a workflow:
image

So if the workflow is executed twice the easiest solution is to set a wait timer in environments (image is from another example repo):
image

If we set the wait timer for the demo store deployment and the demo store is then always the last workflow that runs ... the display should be fine.

Issue #337, opened 2023-07-24
[BUG] Search reactivity broken


BrocksiNet:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Search looks like working when you use the search from a CMS Page or Category Page.

Searching for "base" (looks correct):
![grafik](https://github.com/shopware/frontends/assets/3763023/07837a10-d54c-4ce9-911b-0eefb843d46d)

Searching for "chili" (looks correct):
![grafik](https://github.com/shopware/frontends/assets/3763023/d4884e98-4b26-42d3-8c74-7d37f12bc8ad)

**But** when you do **search again** after you have already searched something (**stay on the search page**) the search is not working anymore.

I searched for "base" after searching for "chili" ... the product grid is not updated:
![grafik](https://github.com/shopware/frontends/assets/3763023/5aafc195-5e1e-4226-b5d2-2d29a9348b24)

Doing a hard reload on the search page also does not work:
![grafik](https://github.com/shopware/frontends/assets/3763023/2153ea16-eac7-46f2-adf8-31c2dc07c894)

### Expected Behavior

Search works in these cases:
- After hard reload with query
- After I search for something the products get updated
- Additional searches working like the first search
- Searching on search page or other pages does not make any difference

PR #341, opened 2023-07-25
fix(demo): search and filters

BrocksiNet:
### Description

fixes:  #337

### Type of change

### Additional context

- The main problem was that the watch was not fired correctly
- There was no cache key set for the productSearch
- Fixed also a problem with the "resetButton" on the search page
- Filters are now updated when you search for different things (setInitialListing)
- Search page is now ignored in routeRules for ssr (fixes hard reload)


kstala (approved):

templates/vue-demo-store/components/listing-filters/ListingFilters.vue
```diff
@@ -133,6 +133,19 @@ const selectedOptionIds = computed(() => [
   ...searchSelectedFilters.properties,
   ...searchSelectedFilters.manufacturer,
 ]);
+const showResetFiltersButton = computed<boolean>(() => {
+  if (
+    selectedOptionIds.value.length !== 0 ||
+    searchSelectedFilters["max-price"] ||
+    searchSelectedFilters["min-price"] ||
+    searchSelectedFilters["rating"] ||
+    searchSelectedFilters["shipping-free"]
```
mdanilowicz, line 142: We should create more flexible functions, there can be custom filters.
Something like this: 
```suggestion
    searchSelectedFilters.some(element => {
     return element !== '' && !!element
    })
```
What do you think?

patzick: `searchSelectedFilters` has `reactive` type, it's not iterable. But for future would be worth refactoring that 👌

Issue #340, opened 2023-07-25
[BUG] Addresses saving problem 


kstala:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When we add a new address (in my account or on checkout) it is not immediately visible on the list, it appears only after refreshing the page.

https://github.com/shopware/frontends/assets/30267600/99b700f7-3e8a-4c93-9cc9-f16db1d35bd3

### Expected Behavior

address is visible immediately after saving the form

### Environment

```markdown
- OS:
- Node:
- pnpm:
```

PR #344, opened 2023-07-28
fix(demo): address refresh after save

BrocksiNet:
### Description

closes: https://github.com/shopware/frontends/issues/340

### Type of change

### Additional context

Passing the success event to the next component to trigger a reload after the address is saved.


kstala:
@BrocksiNet  I checked it in PR preview and the problem still exist

kstala (changes requested):
I checked it in PR preview and the problem still exist

BrocksiNet:
Thx for testing @kstala. And yes I missed one thing.
Pushed the fix for the "New Address" Button.
I was only testing with the "Edit Address" Button.

:+1:

templates/vue-demo-store/pages/account/address.vue
```diff
@@ -72,7 +79,12 @@ onBeforeMount(async () => {
       {{ $t("account.addressAddNew") }}
     </button>
     <SharedModal :controller="addAddressModalController">
-      <SharedAccountAddressForm @success="addAddressModalController.close" />
+      <SharedAccountAddressForm
+        @success="
+          async () =>
```
mdanilowicz, line 84: Maybe adding it as a function in the script would be better? Only for a better readable. What do you think?

templates/vue-demo-store/pages/checkout/index.vue
```diff
@@ -224,6 +224,18 @@ onMounted(async () => {
   });
 });
 
+const refreshAddresses = async (addressId?: string) => {
```
mdanilowicz, line 227: There can be a situation where more than one address will be refreshed, for example, when you add a new address in a separate tab. I think it is better to add a general page loader, and remove `addressId` param from this function

mdanilowicz (changes requested):

templates/vue-demo-store/pages/account/address.vue
BrocksiNet: I had it like that before but then I was thinking why a new function that uses some already existing functions and then I find it better to put it on the spot where the function is executed.

templates/vue-demo-store/pages/checkout/index.vue
BrocksiNet: Improved this

mdanilowicz (approved):

templates/vue-demo-store/pages/checkout/index.vue
```diff
@@ -218,12 +217,17 @@ onMounted(async () => {
...
     getPaymentMethods(),
   ]).finally(() => {
-    isLoading["shippingAddress"] = false;
     isLoading["shippingMethods"] = false;
     isLoading["paymentMethods"] = false;
   });
 });
 
+const refreshAddresses = async () => {
+  isLoading["shipping-address"], (isLoading["billing-address"] = true);
```
patzick, line 226: looks like there is something wrong with the loading here

packages/composables/src/cms/useCmsElementImage.ts
```diff
@@ -18,7 +18,7 @@ export type UseCmsElementImage = {
   containerStyle: ComputedRef<CSSProperties>;
   anchorAttrs: ComputedRef<AnchorHTMLAttributes>;
   imageAttrs: ComputedRef<ImgHTMLAttributes>;
-  imageContainerAttrs: ComputedRef<CSSProperties | ImgHTMLAttributes>;
+  imageContainerAttrs: ComputedRef<{ [k: string]: string }>;
```
patzick, line 21: I do not like this change - it's more implicit, and we should aim for explicitly - there should not be just "string: string" - we should know what we can expect from this field

patzick (changes requested):
two things from my side

packages/composables/src/cms/useCmsElementImage.ts
BrocksiNet: @mdanilowicz I can agree with that :top:

patzick (approved):

PR #345, opened 2023-07-28
feat(demo): reduce bg image size

### LINKED ISSUE #310: [FEATURE] reduce background-size images
### Description

Reduce background images displayed on CMS pages.

### Use Case

Go to: https://frontends-demo.vercel.app/Summer-BBQ/
 you can see that background image has 2MB of size and it's not cropped like other images
image

### Proposed Solution

1. At least we should attach image helper to reduce image size, for example `?width=800&fit=crop,smart` - dynamically like in other images
2. The best would be also to have a solution like `load="lazy"`. It does not exist on CSS, so maybe moving that to separate `img` tag or similar magic would do the trick

### Alternatives Considered

Reducing these images size and loading them on demand will improve performance of the CMS pages

### Additional Context

_No response_

BrocksiNet:
### Description

closes #310

### Type of change

Image sizes before changes (prod):
image

Image sizes after changes (dev):
image

### Additional context
Added a new helper to change the background image URL.


kstala (approved):

packages/helpers/src/cms/getBackgroundImageUrl.ts
```diff
@@ -0,0 +1,31 @@
...
+  return num ? Math.ceil(num / 100) * 100 : DEFAULT_BG_IMAGE_SIZE;
+}
+
+const getUrlFromBackgroundImage = (url: string) => {
+  const regex = /(?:\(['"]?)(.*?)(?:['"]?\))/;
+  const match = url.match(regex);
+  if (!match) {
+    return url;
+  }
+  return match[1];
```
mdanilowicz, line 14:
```suggestion
return !match ? url : match[1]
```

packages/helpers/src/cms/getBackgroundImageUrl.ts
```diff
@@ -0,0 +1,31 @@
+import { CmsBlock, CmsSection } from "@shopware-pwa/types";
+
+const DEFAULT_BG_IMAGE_SIZE = 800;
+function roundUp(num: number) {
+  return num ? Math.ceil(num / 100) * 100 : DEFAULT_BG_IMAGE_SIZE;
+}
+
+const getUrlFromBackgroundImage = (url: string) => {
+  const regex = /(?:\(['"]?)(.*?)(?:['"]?\))/;
```
mdanilowicz, line 9: We should change the variable name or add a comment to describe this regex

mdanilowicz (approved):

packages/helpers/src/cms/getBackgroundImageUrl.ts
BrocksiNet: What would be a better variable name?

BrocksiNet: I added tests for the new helper so I guess it is clear what this regex is doing now.

PR #346, opened 2023-07-28
fix: regex for disable inputs plugin

### LINKED ISSUE #329: [BUG] vite-vue-plugin-disable-inputs regex bug

### Current Behavior

In new vite-vue-plugin-disable-inputs package, VueDisableInputsBeforeMount function have regex to find select elements (packages/vite-vue-plugin-disable-inputs/src/index.ts line: 46-47):
``` ts
const foundSelectElements =
          component.match(/<(select|input|button)(.*\s?)([\s\S]*?)?\>/gm) || [];
```
Its meant to select specific elements (select|input|button) opening tag:
Screenshot from 2023-07-18 14-23-33

But in our case, we have v-if condition in button with > symbol (v-if="current > 0") and regex mistakes that > with tag close symbol:
Screenshot from 2023-07-18 14-27-23

Because of that, :disabled attribute is left outside of foundSelectElements and is not found with hasDisabled (line 50 in same index.ts) function. Therefore second disabled attribute is added, resulting duplicate attribute error in nuxt build.

For now, we fixed this issue in our side, but this bug can cause unexpected errors in app.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
- Node:
- pnpm:
```

### Anything else?

_No response_

BrocksiNet:
### Description

closes #329

### Type of change

### Additional context

For review maybe @kstala and @mkucmus are needed.

I just tested the regex.

image

### DIFF
--- a/packages/vite-vue-plugin-disable-inputs/src/index.ts
+++ b/packages/vite-vue-plugin-disable-inputs/src/index.ts
@@ -44,7 +44,9 @@ function VueDisableInputsBeforeMount(): PluginOption {
         );
 
         const foundSelectElements =
-          component.match(/<(select|input|button)(.*\s?)([\s\S]*?)?\>/gm) || [];
+          component.match(
+            /<(select|input|button)[^>]*>(.*?)<\/(select|input|button)>/gm,
+          ) || [];
 
         for (const selectBlock of foundSelectElements) {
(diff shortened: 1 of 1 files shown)


mkucmus:
I'm closing it due removed package

PR #355, opened 2023-08-06
feat(e2e): add tests for check category filters

kstala:
### Type of change

### DIFF
--- a/apps/e2e-tests/page-objects/CategoryPage.ts
+++ b/apps/e2e-tests/page-objects/CategoryPage.ts
@@ -1,0 +1,27 @@
+import { expect, Locator, Page } from "@playwright/test";
+
+export class CategoryPage {
+  readonly page: Page;
+  readonly selectedFiltersCategory: Locator;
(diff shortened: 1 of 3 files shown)


BrocksiNet:
I see some errors in e2e but the test is not failing.

Look at this screen:
image

BrocksiNet (changes requested):
We talked about this e2e test in some daily. Currently, we depend on the Color with Option "Clear White" and the test will fail when our demo data is changing. So, my wish would be that this test works more dynamically. Like choosing the first filter we can find and then checking if the filter was applied by checking the property in the URL. Also, you could remember the product count that should also be changed when filtering (in most cases).

patzick:
any further work here applying Bjorn's suggestions?

BrocksiNet:
I added a new Test where we select some random filter option from the manufacturer and later check if the filter is present in the URL. See this [commit](https://github.com/shopware/frontends/pull/355/commits/584dcd9a88a7bb26410f9e8a72cd605c756b7734).

What do you think @kstala?

PR #373, opened 2023-08-23
feat(api-client): access admin client session data and change in in the runtime

patzick:
### Description

Idea from @jleifeld

Creates the ability to access and change admin client session data for testing purposes. This comes with fixture example of creating a new API instance per test


mkucmus (approved):
test-tube-national-geographic

PR #380, opened 2023-09-02
fix(composables): adds parameter to loadMore to accept an optional query

### LINKED ISSUE #379: [BUG] loadmore does not respect current route query

### Current Behavior

Currently you can't use the `loadMore` function in the `useListing` composable, because it does not respect the router query. You can't define new search criterias (which is annoying) and you can't update the defaultSearchCriteria once the function is initialized (as far as I can see).

This is breaking the functionality.

### Expected Behavior

Ideally we could just also define new search criteria when using the function, similar to the `changeCurrentPage` function, where we can define `ShopwareSearchParams`.

Here's a suggestion that should work:

```ts
const loadMore = async (query?: Partial<ShopwareSearchParams>): Promise<void> => {
  loadingMore.value = true;
  try {
    const query = query ? query : {
      ...router.currentRoute.query,
      p: getCurrentPage.value + 1,
    };

    const searchCriteria = merge({}, searchDefaults, query);
    const result = await searchMethod(searchCriteria);
    _storeAppliedListing.value = {
      ...getCurrentListing.value,
      page: result.page,
      elements: [
        ...(getCurrentListing.value?.elements || []),
        ...result.elements,
      ],
    };
  } catch (e) {
    throw e;
  } finally {
    loadingMore.value = false;
  }
};
```

### Steps To Reproduce

_No response_

### Environment

_No response_

### Anything else?

_No response_

SuddenDev:
### Description

I'm marking this as a bugfix, as I assume it was not intended for the function to not have a query parameter. Should close #379.

The current problem is that filters, if set, will not be applied to the query on loadMore. This means, that when using a non-pagination style of content fetching, the functionality breaks there. 

### Type of change

### Additional context

I was trying to follow the contributing guidelines as well as I could, but please check if I missed something when making this edit (like documentation or something like that).

### DIFF
--- a/.changeset/chilly-cameras-speak.md
+++ b/.changeset/chilly-cameras-speak.md
@@ -1,0 +1,5 @@
+---
+"@shopware-pwa/composables-next": patch
+---
+
+Adds the ability to specify a query for the loadMore function in the useListing composable.
--- a/packages/composables/src/useListing.ts
+++ b/packages/composables/src/useListing.ts
@@ -81,7 +81,7 @@ export type UseListingReturn<ELEMENTS_TYPE> = {
   /**
    * Loads more (next page) elements to the listing
    */
-  loadMore(): Promise<void>;
+  loadMore(query?: Partial<ShopwareSearchParams>): Promise<void>;
   /**
    * Listing that is currently set
    */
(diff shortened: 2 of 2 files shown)


patzick (approved):
Looks good! Thanks :)

PR #404, opened 2023-09-25
feat: improve setup support

BrocksiNet:
### Description

- Extends the error.vue template in demo store when in development mode
- Returns an error message in the terminal when a connection to API is not possible
- Add a new useConnectionTest composable
- Sorting from A to Z for composables in nuxt module
- Improved code a bit for better readability
- Adding two new entries for troubleshooting page

### Type of change

image
image


BrocksiNet:
Hmmm maybe also a bit connected with https://github.com/shopware/frontends/issues/276 :thinking:

packages/nuxt3-module/src/index.ts
```diff
@@ -11,25 +11,27 @@ export default defineNuxtModule<ShopwareNuxtOptions>({
...
+        shopwareAccessToken: accessToken,
         shopwareApiClient: {
           timeout: moduleConfig.apiClientConfig?.timeout ?? "10000",
         },
       },
     });
 
     // TODO: remove it once nitro server build contains all external packages of nuxt3-module (composables-next)
-    nuxt.options.build.transpile.push("@shopware-pwa/composables-next");
+    //nuxt.options.build.transpile.push("@shopware-pwa/composables-next");
```
mkucmus, line 34: shall we remove it completely?

templates/vue-demo-store/layouts/default.vue
```diff
@@ -4,6 +4,11 @@ export default {
 };
 </script>
 <script setup lang="ts">
+// in development mode we test if your API connection is working
+if (process.env.NODE_ENV == "development") {
```
mkucmus, line 8: this won't work in SSG (with settings in nuxt.config.ts -> `ssr: false`). 

we can move https://github.com/shopware/frontends/pull/404/files#diff-c0db108b331ef293b83598ea004b65e640ea63cbe550676d7c18248c2b60f8d2R20-R32 into `error.vue` file in order to add some `console.error` instead of invoking a getter for session context in a composable. Then the `useConnectionTest` can be removed. 

Pros: 
- will work on ssg as well as ssr
- one file less 👍🏼 

Cons:
- it does not follow an "interceptor" approach (but it's ok as we plan to migrate to new api client)

mkucmus (changes requested):
I like the functional programming approach in a plugin 👍🏼 

See the comments, especially the `ssg` related.

packages/nuxt3-module/src/index.ts
BrocksiNet: I just commented out to see if some test or something is failing in the preview.
But currently, I do not see any problem with removing it completely.

So remove it?

mkucmus: I'm not sure what the purpose was, but maybe to tell nuxt that `composables-next` package must not be imported by nuxt end project (because it's a part of nuxt3-module bundle?). @patzick please help here 😄

mkucmus (approved):
👌🏼

packages/nuxt3-module/src/index.ts
patzick: yeah it looks like was needed on early stages, but if that's working now we can remove it or clean afterwards after we release it

patzick (approved):

Issue #406, opened 2023-09-27
[FEATURE] adyen payments implementation


mkucmus:
### Description

Implementation of how Adyen payments can be used within Composable Frontends. Including corresponding page in docs.

acceptance criteria:
1. payment intent is being created via gateway (cloud app)
2. frontend components are displayed in the checkout step (credit card, and maybe some others available in not only async flow)
3. docs are extended

### Use Case

It's a showcase how Adyen payments can be integrated in a presentation layer like vue-demo-store. Moreover, it's a good benchmark of existing app (not a plugin) in terms of headless approach.

BrocksiNet:
https://github.com/shopware/frontends/pull/407

PR #407, opened 2023-09-29
feat(examples): adyen payments integration

mkucmus:
### Description

the PR shows how to integrate Adyen drop-in component working with Shopware 6 using Composable Frontends.

closes: #406 

- adds a nuxt plugin that exposes the adyenCheckout instance factory
- adds a component to display a drop-in by Adyen
- adds an example logic to communicate between frontend and Shopware 6 instance

### TODO
- [x] add interaction with sw6 via store-api
- [x] consider more dynamic configuration of adyen checkout -> move from module options to composable function
- [x] add README.md
- [x] add styling

### Type of change

image


patzick (commented):
Good work here, let's just change tailwind to UnoCSS for consistency :)

mkucmus:
> Good work here, let's just change tailwind to UnoCSS for consistency :)

sure, I made it consistent now 🙌🏼

examples/adyen-dropin-component/app.vue
```diff
@@ -0,0 +1,98 @@
...
+// composable used to add to cart a product
+const { addProduct } = useCart();
+// configured apiClient instance in order to make a custom API call
+const { apiInstance } = useShopwareContext();
+// current step for progress bar displaying <ProgressBar /> in a template
+const activeStep = ref(0);
+// container for keeping and displaying an incoming payment state from adyen
+const paymentState = ref();
+
+try {
```
mdanilowicz, line 24: It is not good practice to have so many actions in single try/catch block, also console error can be mislead if errors comes `addProduct`.

examples/adyen-dropin-component/app.vue
```diff
@@ -0,0 +1,98 @@
...
+    orderId: order.id,
+    finishUrl: "http://localhost:3000/success",
+    errorUrl: "http://localhost:3000/failure?payment-failed",
+    // adyen specific data
+    stateData: JSON.stringify(state.data),
+  });
+  activeStep.value = 4;
+
+  // check for current payment status after payment authorization attempt
+  const paymentStateResponse = await apiInstance.invoke.post(
```
mdanilowicz, line 65: Maybe we can add this as a composable and use it in this example? It will be more reusable.

mdanilowicz (changes requested):

examples/adyen-dropin-component/app.vue
mkucmus: good point, would be enough to just pass error object to console.error? the example should be as short as possible 👀

examples/adyen-dropin-component/app.vue
mkucmus: good point, we already have such method in useOrder composable, but the example is for extraordinary situation when some extension like "adyen" adds some handling not in shopware's manner. so in this case, we could use our composable if only the parameters passed would be in a different key. 

the value of the example is how to deal with unexpected situations like this one.

mdanilowicz (approved):

PR #414, opened 2023-10-09
fix(docs): adding security headers

BrocksiNet:
### Description

Closes a security issue that was reported (x-frame-options).

### Type of change

- [X] Security issue, missing headers for docs

### Additional context
I used the same config as the developer portal ([see](https://github.com/shopware/developer-portal/blob/main/vercel.json)).
Docu for vercel.json is [here](https://vercel.com/docs/projects/project-configuration#).


BrocksiNet:
@patzick I am not sure if the `vercel.json` is used in preview? 
Currently I do not see the new headers in the preview.

patzick:
should be - but it's not picked up by vercel, maybe because of root project is main repo, but we shouldn't put the file there. What about setting up these headers directly in the vitepress?

BrocksiNet:
I tried two different (config) positions for Vite ... server headers where not added ... at least I do not see them in network tab ... also I am wondering why we have two Vite (config) positions (anyway) in docs ... :confused:

BrocksiNet:
**Conclusion:**
- Only vercel.json at the root level helps to correctly set the needed security headers
- Plugins with middleware working in dev mode but not with serve
- Moved SearchPlugin() to the other plugins and removed not needed duplicated config
- x-robots-tag no-index is only set on preview instances, that's fine

@patzick So now tell me why we should not have vercel.json at the root level. And yes I know this will be set for all projects but is that a problem? We only have security headers there.

mkucmus:
> **Conclusion:**
> 
> * Only vercel.json at the root level helps to correctly set the needed security headers

It's probably because of how vercel builds the project (almost sure that the output is fully static). And then the `vercel.json` is applied in some upper layer like LB to expose the headers before static site 🤔

apps/docs/.vitepress/config.ts
```diff
@@ -324,6 +325,7 @@ export default defineConfigWithTheme<ThemeConfigExtended>({
       stringify: true,
     },
     plugins: [
+      SearchPlugin(),
```
mkucmus, line 328: what plugins do require this add-on? 🕵🏼

BrocksiNet: Moved from deleted vite.config.ts file :scream_cat:

patzick (commented):
@BrocksiNet i fixed building so proper vercel file is taken but it is not applying to all responses yet

I see them only on font responses `.woff` files:
image

other requests does not contain these headers

BrocksiNet:
@patzick I see them also on other requests like this:
image

Do not know how Vercel is doing this but maybe it is related to some cache.

patzick (approved):
Works well on internal navigation. The only time I don't see these headers is on first render of the page

PR #431, opened 2023-10-18
feat(composables): add tests

### LINKED ISSUE #419: [UNITTESTS] useProductAssociations.ts, useProductConfigurator.ts, useProductPrice.ts
### Description

**Current state:**

File                         | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s                                                                                                   
-----------------------------|---------|----------|---------|---------|---------------------------------------------------------------------------------------------------------------------                                                                                    
 useProductAssociations.ts  |   38.61 |      100 |       0 |   38.61 | 40-101                                                                                                              
  useProductConfigurator.ts  |   29.16 |      100 |       0 |   29.16 | 43-144                                                                                                              
  useProductPrice.ts         |   39.65 |      100 |       0 |   39.65 | 47-116   

### Use Case

**Package:** @shopware-pwa/composables-next

We should aim for better unit-test coverage.

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

I assume the new API client will only change the logic inside the components but not the input/output. Also, it is better to have a higher coverage and then adapt the unit tests instead of making everything later when the new API client is ready.

mdanilowicz:
### Description

closes #419 

### Type of change


packages/composables/src/mocks/Product.ts
```diff
@@ -466,98 +466,6 @@ const product: Product = {
   },
   isNew: false,
   calculatedMaxPurchase: 100,
-  //   seoCategory: {
```
mkucmus, line 469: please remove the commented code if it's not relevant

mkucmus (changes requested):
one comment and we can go with merging it 👍🏼

packages/composables/src/mocks/Product.ts
mdanilowicz: @mkucmus This is a `main` branch column :D

mkucmus (approved):
It's good to see it working 💪🏼 
we will discuss on how the shape of composables test can look like in another topic.

PR #436, opened 2023-10-24
feat(examples): Quick Order (from Commercial Features)

mkucmus:
### Description

Exemplary implementation of Quick Order from [Commercial Features](https://docs.shopware.com/en/shopware-6-en/commercial-features/b2b-components)

TODO: 
- [x] ability to change a quantity of a  chosen item
- [x] add "add to shopping cart" button - that adds every chosen item to the cart
- [x] add "clear list" button - that clears chosen items from the cache
- [x] uploader for CSV file 

### How to test it
despite of the info from [README](https://github.com/shopware/frontends/blob/feat/commercial-quick-order-example/examples/commercial-quick-order/README.md):

- rename `.env.template` into `.env`
- adjust credentials (currently, the swstage env is being used, **with no default customer with quick order feature activated, so it must be added beforehand**)
- run it

[Open in StackBlitz](https://stackblitz.com/edit/github-4zir4k?file=README.md)

### Type of change

(_I have a dark mode enabled_)

Added items view
Screenshot from 2023-10-24 17-19-38

Suggest search view
Screenshot from 2023-10-24 17-19-33

### Additional context

file upload can be tested in that format:
```
product_number,quantity
SW10096.3,2
SW10079,5
```

example file:
[products.csv](https://github.com/shopware/frontends/files/13120887/products.csv)


examples/commercial-quick-order/.env.template
```diff
@@ -0,0 +1,4 @@
+VITE_SHOPWARE_API="https://201023-quick-order.swstage.store/"
```
mdanilowicz, line 1: "Version is flagged for deletion".
Adding a temporary instance to the example is not a good idea. We should use our demo or "force" users to add their own instance.

mkucmus: thanks for pointing that! I changed the instance's URL

PR #443, opened 2023-10-27
feat(api-client): invocation params no longer mandatory

### LINKED ISSUE #294: not existing params required to be passed in the invoke action payload
### Description

Some of actions for `invoke` method don't have any payload required to be sent within a request.

image

### Use Case

to not pass unnecessary parameters if not needed

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

patzick:
### Description

closes #294 

### Type of change

Improvement for API client params invocation. Params object is no longer mandatory when no params are defined for the endpoint.


mdanilowicz (approved):

mkucmus (approved):
behold the magic of TS 🧙🏼

I added one exception for path to query transformer, where it tries to iterate over parameters even when they weren't passed.

PR #444, opened 2023-10-28
fix(composables): toggle relative url slash

### LINKED ISSUE #442: [BUG] Corrupted CmsElementImage url

### Current Behavior

Configured CmsElementImage to have link type category in Admin Cms builder.
Screenshot from 2023-10-27 18-08-33

Url is passed to Fe with structure "/Namams/"

useCmsElementImage checks if url is absolute and if not, adds another "/" to the url start.
https://github.com/shopware/frontends/blob/main/packages/composables/src/cms/useCmsElementImage.ts#L61

Result:
Screenshot from 2023-10-27 18-13-06

Double // cancel each other out and browser redirects to url "namams" without domain after click.

### Expected Behavior

useCmsElementImage should check if url already have "/" at the starts, and only add it if its absent.

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
- Node:
- pnpm:
```

### Anything else?

_No response_

mdanilowicz:
### Description

 closes #442 

### Type of change


BrocksiNet:
What is the difference to [this](https://github.com/shopware/frontends/commit/f636e5a648d2d13b83589b34e377f1d3fccc1cf7)?

Could this be unified?

mdanilowicz:
> What is the difference to [this](https://github.com/shopware/frontends/commit/f636e5a648d2d13b83589b34e377f1d3fccc1cf7)?
> 
> Could this be unified?

Those are 2 different packages, we should not use composables helpers in the CMS package. I'm also not sure that we need the same helper there.

mkucmus:
> > What is the difference to [this](https://github.com/shopware/frontends/commit/f636e5a648d2d13b83589b34e377f1d3fccc1cf7)?
> > Could this be unified?
> 
> Those are 2 different packages, we should not use composables helpers in the CMS package. I'm also not sure that we need the same helper there.

If there are two different packages, but the helper function's purpose is the same, please move the function to the shared package named `@shopware-pwa/helpers-next` and use it in two places

mkucmus (approved):
🥇

PR #448, opened 2023-11-02
feat: active class top nav

### LINKED ISSUE #405: [BUG] router-active-class not set on parentNode

### Current Behavior

Currently, the active class is not set on the parentNode when you visit a child link.

When the first level is active:
image

When a child link from the first level is active:
image

Related Docs:
- https://nuxt.com/docs/api/components/nuxt-link
- https://router.vuejs.org/api/interfaces/RouterLinkProps.html#activeClass

There is also something in our nuxt config:
```
  router: {
    options: {
      linkExactActiveClass: "text-brand-primary",
    },
  },
```

Related slack discussion:
https://shopwarecommunity.slack.com/archives/C050L6NCMGQ/p1695566531978429

### Expected Behavior

When you click on a child link in the top navigation the active class is also set on the parent.
This means the child should have an exact active class and parent active class.
This gives the possibility to highlight parentNode when a child is active (currently this is not possible).

### Steps To Reproduce

You can see this in current demo store.

BrocksiNet:
### Description
closes #405

### Type of change

- [X] Bug fix (non-breaking change that fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

image

### Additional context
This solution feels a bit like loops that should not be necessary. 

- No additional requests to get the currently active class from API (always using the cached object)
- Currently, this will only work with the main navigation of two levels
- Work after hard reload with onMounted
- Works during client site navigation with click events
- Reset the active class when the route path changes (watch)
- The click events have a timeout of 100ms to be executed after the watch (sadly)

Maybe we will leave this just as a reference here and not merge it into the demo store :grey_question:

Ideas for improvements :bulb:


templates/vue-demo-store/components/layout/LayoutTopNavigation.vue
```diff
@@ -4,14 +4,92 @@ import {
...
+      }
+      if (navigationElement) {
+        break;
+      }
+    }
+  }
+
+  return navigationElement;
+};
+// only works with 2 level navigation, timeout needed to be executed after watch
```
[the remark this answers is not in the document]
BrocksiNet: Nice :+1: good idea, will try

templates/vue-demo-store/components/layout/LayoutTopNavigation.vue
```diff
@@ -4,14 +4,92 @@ import {
...
+              children[ci].activeClass = true;
+            }
+          }
+        }
+      }
+    }
+  }, timeout ?? 0);
+};
+// reset when route.path changes
+watch(
```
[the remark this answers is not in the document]
BrocksiNet: Yes, exactly that. We can not know the correct way and products can be in multiple categories (hard reload will maybe show the wrong active category). So currently we do not show active at all, we reset it.

templates/vue-demo-store/components/layout/LayoutTopNavigation.vue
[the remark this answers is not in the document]
BrocksiNet: I used a ref to decide if we need to reset or not. Much easier.

packages/types/shopware-6-client/models/content/category/Category.d.ts
```diff
@@ -24,6 +24,7 @@ export type Category = Entity & {
   breadcrumb: string[];
   level: number;
   active: boolean;
+  activeClass: boolean;
```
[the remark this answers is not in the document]
BrocksiNet: Totally correct. So extending the type in the template would be your suggestion?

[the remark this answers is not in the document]
BrocksiNet: Removed it.

templates/vue-demo-store/components/layout/LayoutTopNavigation.vue
```diff
@@ -4,14 +4,92 @@ import {
...
+  const currentNaviagtionElement = findNavigationElement(route.path.slice(1));
+  if (currentNaviagtionElement) {
+    updateActiceClass(
+      currentNaviagtionElement.id,
+      currentNaviagtionElement.parentId,
+    );
+  }
+});
+// only works with 2 level navigation
+const findNavigationElement = (routePath: string): Category | undefined => {
```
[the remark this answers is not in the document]
BrocksiNet: Solved it but without deep merge.

BrocksiNet:
The last changes were:

- Adding some TailwindCss Loading Indicator (reduce jumping, tells customer something is still in progress)
- You now have the possibility to Choose between the old logic and to only display the Sub-Categories of the current category (subCategories yes/no)

packages/cms-base/components/public/cms/element/CmsElementCategoryNavigation.vue
```diff
@@ -1,33 +1,65 @@
...
 };
 
 const globalTranslations = getTranslations();
 translations = deepMerge(translations, globalTranslations) as Translations;
 
 const { category: activeCategory } = useCategory();
-const { loadNavigationElements, navigationElements } = useNavigation();
-const navigations = computed(() => {
+const loading: Ref<boolean> = ref(true);
+const useSubCategories: Ref<boolean> = ref(true); // could be passed maybe as a prop in the future
```
patzick, line 39: it doesn't not change, no need for ref. Also useX is implicating composable, my proposition
```suggestion
const flagAllowSubcategories = true;
```

packages/cms-base/components/public/cms/element/CmsElementCategoryNavigation.vue
```diff
@@ -1,33 +1,65 @@
...
+const categoryNavigation: Ref<Schemas["Category"][] | null> = ref([]);
```
patzick, line 40: no need for null
```suggestion
const categoryNavigation: Ref<Schemas["Category"][]> = ref([]);
```

packages/cms-base/components/public/cms/element/CmsElementCategoryNavigation.vue
```diff
@@ -1,33 +1,65 @@
...
+const cmsCategoryNavigation = async (depth: number) => {
+  return loadMainNavigationElements({ depth: depth });
+};
+
+const loadCategories = async (depth: number) => {
+  try {
+    categoryNavigation.value = (await cmsCategoryNavigation(depth)) ?? [];
+  } catch (e) {
+    const err = e as ClientApiError;
+    console.error("[CmsElementCategoryNavigation]", err.messages);
```
patzick, line 56: try/catch is not needed here, as `loadNavigationElements` is handling the error

packages/cms-base/components/public/cms/element/CmsElementCategoryNavigation.vue
```diff
@@ -1,33 +1,65 @@
...
+const { loadNavigationElements: loadMainNavigationElements } = useNavigation({
+  type: type,
+});
+const cmsCategoryNavigation = async (depth: number) => {
+  return loadMainNavigationElements({ depth: depth });
+};
+
+const loadCategories = async (depth: number) => {
+  try {
+    categoryNavigation.value = (await cmsCategoryNavigation(depth)) ?? [];
```
patzick, line 53: Why not to use directly `loadMainNavigationElements`? It returns empty array by default
```suggestion
    categoryNavigation.value = await loadNavigationElements({depth});
```

patzick (changes requested):
few comments to simplify logic a bit

packages/cms-base/components/public/cms/element/CmsElementCategoryNavigation.vue
```diff
@@ -1,33 +1,55 @@
...
-const navigations = computed(() => {
+const loading: Ref<boolean> = ref(true);
+const flagAllowSubcategories: boolean = true; // could be passed maybe as a prop in the future
+const categoryNavigation: Ref<Schemas["Category"][]> = ref([]);
+
+const currentCategoryId = activeCategory.value?.id ?? "main-navigation";
+const type = flagAllowSubcategories ? currentCategoryId : "main-navigation";
+const { loadNavigationElements: loadMainNavigationElements } = useNavigation({
+  type: type,
+});
```
patzick, line 45:
```suggestion
const { loadNavigationElements } = useNavigation({
  type: type,
});
```

packages/cms-base/components/public/cms/element/CmsElementCategoryNavigation.vue
```diff
@@ -1,33 +1,55 @@
...
+const cmsCategoryNavigation = async (depth: number) => {
+  return loadMainNavigationElements({ depth: depth });
+};
+
```
patzick, line 49:
```suggestion
```

packages/cms-base/components/public/cms/element/CmsElementCategoryNavigation.vue
```diff
@@ -36,35 +58,116 @@ const navigations = computed(() => {
...
+};
 
 onMounted(async () => {
-  try {
-    await loadNavigationElements({ depth: 2 });
-  } catch (e) {
-    console.error("[SwBottomMenu]", e);
+  // depth 0 means, we load only first level of categories, depth 1 means we load first and second level of categories ...
+  const depth = flagAllowSubcategories ? 0 : 1;
+  categoryNavigation.value = await cmsCategoryNavigation(depth);
```
patzick, line 66:
```suggestion
  categoryNavigation.value = await loadNavigationElements({depth});
```

patzick (commented):
omitted one suggestion to use `loadNavigationElements` directly

Issue #209, opened 2023-05-16
[FEATURE] Custom products presentation


mkucmus:
### Description

Currently `Custom products` (provided by built-in cloud extension) aren't displayed properly. See [the link](https://frontends-demo.vercel.app/YORK-2/SW10158).

![image](https://github.com/shopware/frontends/assets/5596960/91e92244-0665-4e33-8aa5-2a60d1ff004a)

SW6 docs: https://docs.shopware.com/en/shopware-6-en/extensions/customproducts

### Use Case

Handling another type of a product. In comparison to other ecommerce platform the look and feel is close to `configurable` products in Magento.

### Proposed Solution

Display product configurator comparable to the standard storefront (it's about mechanism):

how it's done on storefront: https://demo-frontends.shopware.store/YORK-2/SW10158

### Alternatives Considered

because the solution is shipped by a plugin/app on cloud side and some instance may don't have it installed, maybe it's better to prepare the implementation as an external module, or abandon the idea and put the proper information in the docs.

PR #451, opened 2023-11-03
feat(vue-demo-store): customized product example

mkucmus:
### Description

adds a simplified example of a custom implementation of `customized product` feature.

closes: #209 

see the docs: https://github.com/shopware/frontends/blob/79981745fc832bc20ec893085b0b2ddbcd2a24ac/apps/docs/src/resources/integrations/custom-products.md

### Type of change

### ToDo's
- [ ] - show selected option in CartItem of Minicart view

image


packages/composables/src/useCart.ts
```diff
@@ -189,9 +189,7 @@ export function useCartFunction(): UseCartReturn {
   const count = computed(() => {
     return cartItems.value.reduce(
       (accumulator: number, lineItem: LineItem) =>
-        lineItem.type === "product"
-          ? lineItem.quantity + accumulator
-          : accumulator,
+        lineItem.good === true ? lineItem.quantity + accumulator : accumulator,
```
mdanilowicz, line 192: type - Type refers to the entity type of an item whether it is product or promotion for instance.

good - When set to true, it indicates the line item is physical else it is virtual.


Are you sure about this change? Looks like the virtual product will be omitted

templates/vue-demo-store/components/product/ProductAddToCart.vue
```diff
@@ -9,9 +9,17 @@ const { product } = toRefs(props);
...
 const { addToCart, quantity } = useAddToCart(product);
+const {
+  addToCart: customizedProductAddToCart,
+  isActive: isCustomizedProductActive,
+} = useProductCustomizedProductConfigurator();
 
 const addToCartProxy = async () => {
-  await addToCart();
+  if (isCustomizedProductActive.value) {
+    await customizedProductAddToCart();
```
mdanilowicz, line 19: Can we add this logic to the `addToCart`? `useAddToCart` composable should be more flexible

packages/composables/src/useCart.ts
mkucmus: thanks for hinting that! I've created an issue regarding it IMO in a complex way: https://github.com/shopware/frontends/issues/456

regarding the `good` thing ;) do we have such thing like virtual product? I know the digital one, and the promotion which is now displayed when there is no product in the cart. do we maybe want to show any promotion anyway?

templates/vue-demo-store/components/product/ProductAddToCart.vue
mkucmus: I'm here to discuss about it because It's not clear yet where to put the example. Currently it's a part of the template only, and the composable itself is located here, not as a part of composables package. The reason is that 
the composables are designed to be more "core"-ish,  would be for a core-related features only, and the conditionally enabled feature like "custom product" would be easy to delete/change in the template level. WDYT?

apps/docs/.vitepress/config.ts
```diff
@@ -213,6 +213,10 @@ export const sidebar = [
             link: "/resources/integrations/multi-cms",
           },
           { text: "Strapi", link: "/resources/integrations/strapi/" },
+          {
+            text: "Custom Products extension",
+            link: "/resources/integrations/custom-products",
```
BrocksiNet, line 218: I would put it below:
https://frontends.shopware.com/getting-started/features/

Because this is not really an integration.

mkucmus: I've moved it to the mentioned chapter 👍🏼

BrocksiNet: Nice then remove the old link from the sidebar also.

BrocksiNet:
Is it planed to show the selected options in a cart/mini-cart or somewhere after the customer adds the product?

mkucmus:
> Is it planed to show the selected options in a cart/mini-cart or somewhere after the customer adds the product?

What about implementing it during https://github.com/shopware/frontends/issues/456 ? because it seems to be more complex task to avoid a lot of if statements for just one use case

BrocksiNet:
Sure we can do it in a separate issue/task.

BrocksiNet:
I guess we need to update the branch here :grinning:

packages/composables/src/useCartItem.ts
```diff
@@ -37,6 +37,10 @@ export type UseCartItemReturn = {
    * Determines if the current item is a promotion
    */
   isPromotion: ComputedRef<boolean>;
+  /**
+   * Determines if the current item can be removed from cart
+   */
+  isRemovable: ComputedRef<boolean | undefined>;
```
patzick, line 43: we can do this as boolean as it's used in v-if either way
```suggestion
  isRemovable: ComputedRef<boolean>;
```

packages/composables/src/useCartItem.ts
```diff
@@ -111,6 +115,8 @@ export function useCartItem(
 
   const isPromotion = computed(() => cartItem.value.type === "promotion");
 
+  const isRemovable = computed(() => cartItem.value.removable);
```
patzick, line 118:
```suggestion
  const isRemovable = computed(() => !!cartItem.value.removable);
```

templates/vue-demo-store/components/product/ProductCard.vue
```diff
@@ -1,5 +1,8 @@
 <script setup lang="ts">
-import type { BoxLayout, DisplayMode } from "@shopware-pwa/composables-next";
+import {
```
patzick, line 2: Let's stay with the previous imports and stick with the convention

patzick (changes requested):
let's move this to examples to avoid demo template complications

templates/vue-demo-store/components/product/ProductCard.vue
mkucmus: done

mkucmus:
> let's move this to examples to avoid demo template complications

ok, it's moved now.

examples/commercial-customized-products/apiSchema-6.4.18.1.json
BrocksiNet, line 1: Do we need the apiSchema files in the example?

BrocksiNet: Should we not have them in api client package?

packages/api-client-next/src/transformPathToQuery.ts
```diff
@@ -68,7 +70,13 @@ export function transformPathToQuery<T extends Record<string, unknown>>(
...
       Reflect.set(returnOptions.body, key, params[key]);
     }
-  });
+  }
+
+  // Exception for unknown type of FormData - multipart/form-data MIME type
+  // pass as it is
+  if (params instanceof FormData) {
+    returnOptions.body ??= params as T;
+  }
```
patzick, line 79: we need to add tests for that ad I'm not sure in which cases it should be considered

packages/api-client-next/src/transformPathToQuery.ts
```diff
@@ -55,11 +57,11 @@ export function transformPathToQuery<T extends Record<string, unknown>>(
     body?: Partial<T>;
   };
 
-  if (!params) {
+  if (!params || ["head", "get"].includes(method)) {
```
patzick, line 60: we need to add tests for this with the use-case

packages/api-client-next/src/index.ts
```diff
@@ -123,7 +136,9 @@ export function createAPIClient<
   ): Promise<RequestReturnType<OPERATION_NAME, OPERATIONS>> {
     const [requestPath, options] = transformPathToQuery(
       pathParam,
-      params?.[0] as Record<string, string>,
+      Array.isArray(params)
+        ? (params?.[0] as Record<string, string>)
+        : (params as Record<string, string>),
```
patzick, line 141: in which case do we have it? Tests needed

patzick (changes requested):
Looks good. We need tests coverage (api-client) with real use cases and explanation what should happen and why.

Also for example itself:
- [ ] adding custom product is not working
- [x] uploading file
- [ ] we could think of simplifying overriding structure more

packages/api-client-next/src/transformPathToQuery.ts
mkucmus: it's the only case when we pass a FormData (multipart/form-data content type), it cannot be split into smaller pieces or destructured. so it's always a one compact object. 
I added a test-case for that.

packages/api-client-next/src/index.ts
mkucmus: you know what question to ask. I removed it, and probably it added it due some not related problems I had. cleaned up!

patzick (approved):

PR #452, opened 2023-11-06
feat!: override components and composables

patzick:
## List of changes

Info: signature ********************************breaking change********************************: means that these changes might be breaking and require some migration steps on the project side

- changed `composables` and `cms-base` packages from nuxt modules to nuxt layers - up to discussion if we want to introduce it as separate packages not to have breaking chane. Braking change would be very minor - 2 lines in nuxt config file.
- changed `composables` package to export layer module with `nuxt.config.ts` as main file, built version with no possibility to override) is accessible under `/dist` for now (added as discussion point below)
- **breaking change**: imports now are available under `#imports` with all dynamic
- **breaking change**: types are now imported in a matter that can be overriden as well
- possibility to override API types

## PR changes to finish

- [x]  add migration step to template changelogs
    - [x]  changing direct imports to `import {} from "#imports"`
    - [x]  changing types imports to `import type { Schemas } from "#shopware"`
- [x]  replace `types` definitions with imports from `#shopware` to ensure that types comes from API, add deprecation message on types in `types` package. This can help avoid breaking change and allow for smoother transition towards new types definitions
- [x]  improve importing types from `composables` package, currently it’s `import type { SliderElementConfig } from "@shopware-pwa/composables-next/composables";`
- [x]  expose `_useContext` composable as `useContext` in package

### Issues to create after this PR

- [ ]  address openAPI schema TODOs and create issues + generator fixes for them until resolved
- [ ]  adding more examples on how to manually override API types
- [ ]  rewrite `helpers` to use generic types instead of `types`
- [ ]  track all `types` package usage and progress with deprecation and removal process
- [ ]  prepare template for comsposable unit testing
- [ ]  guide to create types from custom instance and override them manually

### Things to consider and decide

- [x]  move composables and cms to layers, do we still need nuxt module?
    - might still be useful for example for devtools
- [x]  check if plugins by default could be used inside layers and override it inside project if needed
- [x]  imports for builded version (currently work version) `/composables/**dist**`
    - experiment with other possibilities for exports
    - consider another package for build package (no overrides)


mkucmus (approved):
like a christmas gift 🎅

PR #458, opened 2023-11-08
fix: import nuxt composables from #imports

danielroe:
### Description

This is a DX improvement when developing - we can avoid loading the entire barrel file at `#app` by using the new granular imports merged in https://github.com/nuxt/nuxt/pull/23951.

### Type of change


patzick (approved):
This is nice! Thank you @danielroe 🙌

PR #463, opened 2023-11-14
Fix/gh 461

mdanilowicz:

### DIFF
--- a/templates/vue-demo-store/components/checkout/CheckoutCartItem.vue
+++ b/templates/vue-demo-store/components/checkout/CheckoutCartItem.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import { getSmallestThumbnailUrl } from "@shopware-pwa/helpers-next";
-import type { LineItem } from "@shopware-pwa/types";
+import type { LineItem, ClientApiError } from "@shopware-pwa/types";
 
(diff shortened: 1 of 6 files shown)


templates/vue-demo-store/components/checkout/CheckoutCartItem.vue
```diff
@@ -37,9 +37,15 @@ const updateQuantity = async (quantityInput: number | undefined) => {
...
-  const response = await changeItemQuantity(Number(quantityInput));
-  // Refresh cart after qty update
-  await refreshCart(response);
+  try {
+    const response = await changeItemQuantity(Number(quantityInput));
+    // Refresh cart after qty update
+    await refreshCart(response);
+  } catch (error) {
+    const e = error as ClientApiError;
+    const { errors } = useApiErrorsResolver(e.messages);
```
patzick, line 46: composables shouldn't be invoked in the runtime functions, proposed API:

```
const {resolveApiErrors} = useApiErrorsResolver();
```

and then in the function's block
```
 } catch (error) {
    const errors = resolveApiErrors(e.messages);
}
```

patzick (changes requested):

kstala (approved):

patzick (approved):

PR #484, opened 2023-12-04
fix(demo): search and filters after major update

BrocksiNet:
### Description

After the major update the search result page and some filters (manufacturer) where not working anymore.

### Type of change

One of the problems (currently on demo):
image

Fixed on local:
image

### Additional context
Rendering manufacturer was not possible because of some types issue.


templates/vue-demo-store/components/listing-filters/ListingFiltersProperties.vue
```diff
@@ -1,21 +1,16 @@
...
-  "
->
+<script setup lang="ts">
 import { inject, ref } from "vue";
 import { getTranslatedProperty } from "@shopware-pwa/helpers-next";
 import type { Schemas } from "#shopware";
 
 defineProps<{
-  filter: ListingFilter;
+  filter: {
```
patzick, line 7: Why have you switched from generic? It might cause type problems if anyone would like to use this component directly.

patzick (changes requested):
One comment

patzick (approved):

PR #503, opened 2023-12-13
fix(demo): click on label bug

BrocksiNet:
### Description
Related to PR https://github.com/shopware/frontends/pull/484

### Type of change

### Additional context

After the fieldset was introduced for accessibility the click event was not working on the label.
Currently, the event is fired twice and also not working for labels only for the input.
With these changes, it works on category and search the same way.

### DIFF
--- a/packages/cms-base/components/listing-filters/SwFilterProperties.vue
+++ b/packages/cms-base/components/listing-filters/SwFilterProperties.vue
@@ -59,7 +59,7 @@ const toggle = () => {
             v-for="option in props.filter.options || props.filter.entities"
             :key="`${option.id}-${selectedOptionIds?.includes(option.id)}`"
             class="flex items-center"
-            @click="
+            @click.once="
               emits('select-value', {
                 code: props.filter.code,
                 value: option.id,
@@ -100,12 +100,6 @@ const toggle = () => {
             <label
               :for="`filter-mobile-${props.filter.code}-${option.id}`"
               class="ml-3 text-gray-600"
-              @click="
-                emits('select-value', {
-                  code: props.filter.code,
-                  value: option.id,
-                })
-              "
             >
               {{ getTranslatedProperty(option, "name") }}
             </label>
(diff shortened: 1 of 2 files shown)


packages/cms-base/components/listing-filters/SwFilterProperties.vue
```diff
@@ -59,7 +59,7 @@ const toggle = () => {
             v-for="option in props.filter.options || props.filter.entities"
             :key="`${option.id}-${selectedOptionIds?.includes(option.id)}`"
             class="flex items-center"
-            @click="
+            @click.once="
```
patzick, line 62: div's should not be clickable, it creates problem with accessibility. Could you try to make it on `change` event inside `input`?

templates/vue-demo-store/components/listing-filters/ListingFiltersProperties.vue
```diff
@@ -53,12 +53,13 @@ const toggle = () => {
...
       <div v-show="isFilterVisible" id="filter-section-0" class="pt-6">
-        <div class="space-y-4">
+        <fieldset class="space-y-4">
+          <legend class="sr-only">{{ props.filter.name }}</legend>
           <div
             v-for="option in props.filter.options || props.filter.entities"
             :key="`${option.id}-${selectedOptionIds?.includes(option.id)}`"
             class="flex items-center"
-            @click="
+            @click.once="
```
patzick, line 62: here as well

patzick (changes requested):

packages/cms-base/components/listing-filters/SwFilterProperties.vue
mdanilowicz: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role will resolve accessibility problem

patzick: or we can just react to `change` event of the input field and not have workarounds :)

patzick (approved):

PR #514, opened 2023-12-19
feat(templates): i18n module upgrade for vue-demo-store

mkucmus:
### Description

Upgrades @nuxjs/i18n module to the latest rc.11 version.

- removes JIT warning (`compile.jit: false` config is not required anymore)
- `jsTsFormatResource` optional config is not required anymore to read translations properly from .ts files

### DIFF
--- a/.changeset/fifty-bats-type.md
+++ b/.changeset/fifty-bats-type.md
@@ -1,0 +1,5 @@
+---
+"vue-demo-store": patch
+---
+
+@nuxtjs/i18n module upgrade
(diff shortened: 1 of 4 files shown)


mkucmus:
@kstala please have a look also on translations within CMS elements imported from `cms-base` package. I think the problem might be general and we need to find out how to cover the imported packages with dynamic translations coming from app context (nuxt app). If it's not working (check also the canary version), please create a ticket with the details.

mdanilowicz:
Screenshot 2023-12-22 at 10 45 41

Not everything is translated. Can you add the missing translations? Or check why it is not working?

patzick (approved):

PR #517, opened 2023-12-22
feat(demo): adding new colors definition

BrocksiNet:
### Description
Added all colors to uno.config.ts to show that they can be changed via config.

Changes in detail:

- blue is now covered by primary.
- gray and slate are now covered by secondary.
- brand-primary is now primary (default).
- brand-light is now light (default).
- brand-dark is now dark (default).
- Added other colors to the config like yellow, green, indigo, and white.

### Type of change
Breaking change for cms-package (if some projects depend on our color definitions)

image

### Additional context
Showing all the colors in the config will help people to see how they can be changed easily.


mdanilowicz (approved):
Nice!

PR #520, opened 2023-12-31
feat: add getProductCrossSelling with endpoint

### LINKED ISSUE #519: [FEATURE] Cross-selling Endpoint and api client method
### Description

I m currently working on a headless demo shop and noticed that there is a endpoint missing for fetch the cross selling products.

### Use Case

Fetch Cross Selling Products by product id

### Proposed Solution

Adding the Endpoint with a method to fetch the data by product id.

### Alternatives Considered

_No response_

### Additional Context

_No response_

jonathan-martz:
Solution for Issue closes #519 

Is there anything missing to merge it into main ?
Is there already a method which solves the problem to fetch cross selling products ?


patzick (approved):
I encourage to use new api client - https://www.npmjs.com/package/@shopware/api-client but until it's fully stable we can add this endpoint here :)

PR #524, opened 2024-01-09
feat(cms-base): export helpers, rename getTranslations

BrocksiNet:
### Description

Rename getTranslations to getCmsTranslations.  
Export some helpers for external use e.g. in demo-store:

```
import { getCmsTranslations, deepMerge } from "@shopware-pwa/cms-base/helpers";
```

### Type of change

### ToDo's

- [X] Changeset file provided

### Additional context

Currently, the helpers need to be copied to the project folder when you want to overwrite some elements because you can not use them from the original CMS base. With that change, you can use most of the helpers (not all).

We should think about to allow overwriting `SwProductCard.vue` and so on. Because currently, you have to change every element that is using `SwProductCard` and then replace it with your own `SwProductCard`. And that's a lot of useless code.

What do you think @mkucmus @patzick @mdanilowicz :hugs:


patzick (changes requested):
I wouldn't go this way to open package API. Initially, these were internal helpers; they are not properly tested nor documented and would create more maintenance problems, especially things like `deepMerge`.

My proposition for exported helpers:
- `buildUrlPrefix` - move to helpers package, test and document
- `deepMerge` - remove this helper and use [defu](https://www.npmjs.com/package/defu) instead, well tested, more powerful and small
- `getCmsTranslations ` - move to composables as `useCmsTranslations`
- `getUrlPrefix` - move to composables as  method in `useUrlResolver`
- `resolveUrl` - move to composables as  method in `useUrlResolver`

this way we have cleaner areas of responsibilities and APIs. WDYT?

packages/composables/package.json
```diff
@@ -60,6 +60,7 @@
     "@shopware-pwa/helpers-next": "workspace:*",
     "@shopware/api-client": "workspace:*",
     "@vueuse/core": "^10.7.1",
+    "defu": "^6.1.3",
```
mdanilowicz, line 63: Why do we need a whole package?
There was a file for deep merging translations.

BrocksiNet: Comment from Patryk before:
`remove this helper and use [defu](https://www.npmjs.com/package/defu) instead, well tested, more powerful and small`

patzick: Yeah, mostly as it was untested code and maintenance burden 👌

patzick (approved):

Issue #139, opened 2023-04-13
[BUG] Category breadcrumbs path is missing


StanislovasRamanauskas:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently useBreadcrumbs() composable only returns names array for category breadcrumbs. Paths are missing so link to category cannot be created.

E.G.
```js
{
  "name": "Food"
}
```

Helper function to get breadcrumbs getCategoryBreadcrumbs() also gather only names:
```js
export function getCategoryBreadcrumbs(
  category: Category | null | undefined,
  options?: {
    /**
     * Start at specific index if your navigation
     * contains root names which should not be visible.
     */
    startIndex?: number;
  }
) {
  const breadcrumbs =
    category?.translated?.breadcrumb || category?.breadcrumb || [];
  const startIndex = options?.startIndex || 0;
  if (breadcrumbs.length <= startIndex) return [];
  return breadcrumbs.slice(startIndex).map((element) => {
    return {
      name: element,
    };
  });
}
```

Breadcrumbs in product.seoCategory / category API response also contain only names. So I think this is issue both in BE and FE parts:
![Screenshot from 2023-04-14 00-07-12](https://user-images.githubusercontent.com/38095299/231883067-4f2a619d-81c8-4099-b3da-b3bd7e05c490.png)

### Expected Behavior

As declared in useBreadcrumbs return type I expect category breadcrumb with path returned:
```
type Breadcrumb = {
    name: string;
    path?: string;
};
```

### Environment

```markdown
- OS:
- Node:
- npm:
```

BrocksiNet:
Thx for submitting this already known issue. We have an discussion already with the product area that is responsible to return the path via API. Also good to have this issue here to do not forget it. Will ping you when I have news about this.

BrocksiNet:
@patzick could we use the same technic we used for "duplicated" seoURL''s calls to solve this issue? Like looking in the list if we have this seoURL already and then adding the link instead of waiting for the changed response? Just an idea.

BrocksiNet:
@StanislovasRamanauskas here is an example of how to solve this in frontends (but this is just a workaround):
https://github.com/shopware/frontends/pull/258/files

And I do not know if we will merge this into the demo-store.

BrocksiNet:
For reference: NEXT-24954

BrocksiNet:
[...]

PR #531, opened 2024-01-16
feat(vue-demo): JSON-LD example

### LINKED ISSUE #530: [FEATURE] How to use JSON-LD
### Description

Add example of how to use JSON-LD in the Frontends app

### Use Case

-

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

mdanilowicz:
### Description

closes #530


BrocksiNet:
I like this PR a lot. Thx. :+1: 

Only two comments:
- Description has HTML inside and should only be text
- With JSON LD you can have multiple entries. Not only Product. A classic example would be Product + Breadcrumb in JSON LD. Should our small example support that possibility?

Also nice reference [here](https://developers.google.com/search/docs/appearance/structured-data/product).

mdanilowicz:
> I like this PR a lot. Thx. 👍
> 
> Only two comments:
> 
> * Description has HTML inside and should only be text
> * With JSON LD you can have multiple entries. Not only Product. A classic example would be Product + Breadcrumb in JSON LD. Should our small example support that possibility?
> 
> Also nice reference [here](https://developers.google.com/search/docs/appearance/structured-data/product).

Ok I removed tags.
Regarding the breadcrumbs, there are no links in response so we should wait for the https://github.com/shopware/frontends/issues/139

The main goal of this task is to show - HOW. That is why there is a possibility of overriding the JSON by the additional param. Also, this is a template composable, so each app should have its own implementation of it.

BrocksiNet:
Sure, we can also create a follow-up issue after merging this PR.
To improve the example when breadcrumbs are done. Like it. :+1:

BrocksiNet (approved):

patzick (approved):

PR #540, opened 2024-01-24
fix(cms-base): getOptionsFromNode

BrocksiNet:
### Description

There was an error thrown for a customer project.
The error message was like "undefined can not converted to object".

### Type of change

### ToDo's

### Additional context

In old versions, attr was outputted to the HTML DOM.
Current testing no error in console and no empty attributes in HTML DOM.


patzick (changes requested):
thanks for showing the problem and inspiration, please check #542 as I did refactor a bit differently and tested whole method

PR #544, opened 2024-01-26
fix(vue-demo-store): type deprications in vue-demo-store

jhony1104:
### Description

Fixed type based deprications in the __vue-demo-store__ by using `import type { Schemas } from "#shopware"`.

### Type of change


patzick (approved):
👌 great, thank you!

PR #545, opened 2024-01-29
fix(vue-demo-store): visibility classes

itscark:
### Description

Fixes element visibility.

### Type of change


patzick (approved):
Great work, thanks! 🙌 
I added tests for that and follow up in #549 to export all classes and simplify adding it in theme.

PR #550, opened 2024-01-30
refactor: add biome linter to helpers package, improve types and force full code coverage

patzick:
### Description

- added biome and replaced linter in `helpers` package for the first try
- improved linter errors, almost all `any` types removed (left only in deprecated methods which will be removed)
- updated vitest code coverage to fail when 100% is not reached and added missing tests


mkucmus (approved):
hey world, we also have a biome!

PR #551, opened 2024-01-31
feat: adjust reviews type

mdanilowicz:


patzick:
we won't update it manually, let's wait for the RC version

Issue #553, opened 2024-01-31
[FEATURE] api-next: allow passing directly the oauth credentials instead of session data


shyim:
### Description

Right now when you create an admin client, you have to pass the "sessionData" which is the oauth API response. It's annoying to craft this by own. There should be an optional field where I can put the password or client credentials grant data, and it fetches his token

### Use Case

Make easier to use the SDK

### Alternatives Considered

```
const auth = (await (
  await fetch(`${process.env.BASE_URL}/api/oauth/token`, {
    method: "POST",
    headers: {
      "Content-Type": "application/json",
    },
    body: JSON.stringify({
      client_id: "administration",
      grant_type: "password",
      username: "admin",
      password: "shopware",
    }),
  })
).json()) as {
  access_token?: string;
  refresh_token: string;
  expires_in: number;
};
```

patzick:
hey @shyim, `sessionData` is not required if you don't have it on creation you simply don't pass it. In example we do use `cookies` for storing session data, but you can use other storage or just keep it memory (default behaviour, then you don't even need to pass `onAuthChange` behaviour).

Then you need to log in, depending on the grant type there are different approaches, but the simplest one with password:
```ts
await client.invoke("token post /oauth/token", {
      grant_type: "password",
      client_id: "administration",
      scopes: "write",
      username: "USERNAME",
      password: "PASSWORD",
    });
```

and from now on (unless your credentials are correct) you can access all other endpoints. Let me know if you need any further clarification

shyim:
That's kinda exactly what I did and I think it's bad for the Client to do that manually. My expectation would be, I can pass the credentials optionally to the object in the createAdminClient. and then when the sessionData is empty, you authorize it for me.

something like this:

```
export const adminApiClient = createAdminAPIClient<operations, operationPaths>({
  baseURL: `${process.env.SHOP_URL}/api`,
  credentials: {
      grant_type: "password",
      client_id: "administration",
      scopes: "write",
      username: "USERNAME",
      password: "PASSWORD",
    }
});
```

and then I can invoke directly without further thinking the API

shyim:
maybe a real world example :) https://github.com/shopwareLabs/k6-shopware/blob/main/fetch-fixtures.ts

patzick:
thanks @shyim, that's a good use-case for scripting and not managing the session during runtime. I created PR back to you with improvements in that area. Looking forward for more feedback! 🙌

PR #560, opened 2024-02-02
feat(api-client): provide optional `credentials` field for automatic authentication

patzick:
### Description

closes: #553 

### Type of change

Replaced additional only token-based `clientData` with `optional `credentials` field.

This way, scripts or apps can provide admin authentication of their choice to be used automatically before invoking any endpoint.


mkucmus (approved):
image

PR #564, opened 2024-02-05
feat(api-gen): added `apiType` option to `loadSchema` command

patzick:
### Description

Adds the possibility to pick API type on schema generation plus uses API client to authenticate and load schema using our apiClient.

### DIFF
--- a/packages/api-gen/README.md
+++ b/packages/api-gen/README.md
@@ -52,8 +52,8 @@ pnpx @shopware/api-gen loadSchema --apiType=admin --filename=adminApiSchema.json
 Remember to add `.env` file in order to authenticate with Shopware instance.
 
 ```bash
-OPENAPI_JSON_URL="https://demo-frontends.shopware.store"
-OPENAPI_ACCESS_KEY="SWSCBHFSNTVMAWNZDNFKSHLAYW"
+OPENAPI_JSON_URL="https://your-shop-instance.shopware.store"
+OPENAPI_ACCESS_KEY="YOUR_STORE_API_ACCESS_KEY"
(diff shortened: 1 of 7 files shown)


packages/api-gen/README.md
```diff
@@ -35,6 +41,21 @@ options:
...
+
+# load schema from admin API
+pnpx @shopware/api-gen loadSchema --apiType=admin --filename=adminApiSchema.json
+```
+
+Remember to add `.env` file in order to authenticate with Shopware instance.
+
+```bash
+OPENAPI_JSON_URL="https://demo-frontends.shopware.store"
+OPENAPI_ACCESS_KEY="SWSCBHFSNTVMAWNZDNFKSHLAYW"
```
mdanilowicz, line 56: We should use placeholders instead of real data, even if they are just test data. It can lead to unnecessary questions

packages/api-gen/src/commands/loadSchema.ts
```diff
@@ -2,28 +2,99 @@ import { writeFileSync } from "node:fs";
...
+const SCHEMA_ENDPOINT = "_info/openapi3.json";
+const STORE_API_ENDPOINT = `/store-api/${SCHEMA_ENDPOINT}`;
+const ADMIN_API_ENDPOINT = `/api/${SCHEMA_ENDPOINT}`;
+
+export async function loadSchema(args: {
+  cwd: string;
+  filename: string;
+  apiType: string;
+}) {
+  if (args.apiType !== "store" && args.apiType !== "admin") {
```
mdanilowicz, line 27:
```suggestion
  if (!["store", "admin"].includes(args.apiType)) {
```

mdanilowicz (approved):

PR #569, opened 2024-02-08
fix(composables): available shipping methods

itscark:
### Description
Fix loading only available shipping methods


.changeset/curvy-mice-give.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-pwa/composables-next": minor
```
patzick, line 2:
```suggestion
"@shopware-pwa/composables-next": patch
```

patzick (approved):

PR #615, opened 2024-02-19
feat(docs): 503 error devenv

BrocksiNet:
### Description

Add new troubleshooting hint about telemetry, devenv and, frontends.

### Type of change

Only docs.

### DIFF
--- a/apps/docs/src/resources/troubleshooting.md
+++ b/apps/docs/src/resources/troubleshooting.md
@@ -63,3 +63,7 @@ The HTTP status code 412 (Precondition Failed) usually means in the Shopware `st
 ## Access from origin 127.0.0.1:3000 has been blocked by CORS policy
 
 Depending on your server, you may need to set the `Access-Control-Allow-Origin` header to access your server from an external origin. And yes, your local development server is also an external origin in this case. Also, have a look at this [documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSMissingAllowOrigin) from MDN.
+
+## Local frontends used in devenv only show 503 errors on the network tab?
+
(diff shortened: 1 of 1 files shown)


apps/docs/src/resources/troubleshooting.md
```diff
@@ -63,3 +63,7 @@ The HTTP status code 412 (Precondition Failed) usually means in the Shopware `st
 ## Access from origin 127.0.0.1:3000 has been blocked by CORS policy
 
 Depending on your server, you may need to set the `Access-Control-Allow-Origin` header to access your server from an external origin. And yes, your local development server is also an external origin in this case. Also, have a look at this [documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSMissingAllowOrigin) from MDN.
+
+## Local frontends used in devenv only show 503 errors on the network tab?
```
fschmtt, line 67: Hmm I'm not sure if it's only related to devenv? 🤔 

I would guess it's a general thing, maybe you have not encountered it, because you have disabled this globally in some `.nuxt` config directory in your system or so 😄

patzick (changes requested):
@BrocksiNet  let's add it in our composables layer `nuxt.config` (other nuxt.configs also fine) and this way that line is not needed

apps/docs/src/resources/troubleshooting.md
```diff
@@ -63,3 +63,7 @@ The HTTP status code 412 (Precondition Failed) usually means in the Shopware `st
...
+
+If you encounter this problem, try to set the option `NUXT_TELEMETRY_DISABLED=1` in your local .env file for your template/project root folder, see also [Nuxt telemetry module](https://github.com/nuxt/telemetry).
```
shyim, line 69: It would be nice to explain why it happens. I would personally disable telemetry always 😅

BrocksiNet:
So we will not add this to docs and just set the default to false for telemetry (see https://github.com/shopware/frontends/pull/619).

PR #616, opened 2024-02-19
feat: Add devenv

fschmtt:
### Description

Add [devenv](https://devenv.sh/) support to enable developers using devenv to easily get started:
* No manual installation of dependencies.
* Reproducible environments for all developers.
* Environment configured properly OOTB for running the *Getting started* section.

### Type of change

New feature

### ToDo's

- [ ] Check versions of installed packages (e.g. pnpm, node etc.)

### DIFF
--- a/devenv.nix
+++ b/devenv.nix
@@ -1,0 +1,16 @@
+{ pkgs, lib, ... }:
+
+{
+  # Packages
+  packages = [
+    pkgs.nodePackages.pnpm
+  ];
+
+  # Languages
+  languages.javascript.enable = lib.mkDefault true;
+  languages.javascript.package = lib.mkDefault pkgs.nodejs_18;
+
+  # Environment
+  env.NODE_TLS_REJECT_UNAUTHORIZED = lib.mkDefault "0";
+  env.NUXT_TELEMETRY_DISABLED = lib.mkDefault "1";
+}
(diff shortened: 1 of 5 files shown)


devenv.nix
```diff
@@ -0,0 +1,16 @@
...
+
+{
+  # Packages
+  packages = [
+    pkgs.nodePackages.pnpm
+  ];
+
+  # Languages
+  languages.javascript.enable = lib.mkDefault true;
+  languages.javascript.package = lib.mkDefault pkgs.nodejs_18;
```
fschmtt, line 11: We can pin the node (major) version - is 18 correct?

devenv.nix
```diff
@@ -0,0 +1,16 @@
+{ pkgs, lib, ... }:
+
+{
+  # Packages
+  packages = [
+    pkgs.nodePackages.pnpm
```
fschmtt, line 6: This will use the latest version, is this OK?

devenv.nix
```diff
@@ -0,0 +1,16 @@
...
+  ];
```
fschmtt, line 7: Are additional packages required?

For me it was sufficient ot only add `pnpm`.

patzick (changes requested):
Hey @fschmtt 
Please write more about your motives and what you are trying to achieve. Currently I see no reason for adding this to the monorepo.

devenv.nix
shyim: enable ```
languages.javascript.corepack.enable = true;
```

and remove pnpm as package. so it uses the pinned version in the package.json

fschmtt:
Hey @patzick, for me as someone mainly using devenv for running all sorts of projects locally, it's just super nice to not have a "break" in DX 😄 

I updated the description of the PR acordingly.

I don't insist on merging this, if you feel like it's not contributing to the project, though 🙂

PR #619, opened 2024-02-19
fix: opt-out telemetry, caused 503

BrocksiNet:
### Description
Default telemetry set to false after discussion (https://github.com/shopware/frontends/pull/615).

### Type of change


BrocksiNet:
Now covered by #620 :joy:

PR #623, opened 2024-02-19
fix: do not change context token if no context-token header in api re…

### LINKED ISSUE #622: New API-client: context gets resetted unneccessarily
Because not every API-Call from Shopware returns a new Context-Token via the header, it gets resetted, if no header is sent.
https://developer.mozilla.org/en-US/docs/Web/API/Headers/get 

If no header is set, `headers.get` returns null which makes the following condition truthy.
https://github.com/shopware/frontends/blob/dc9c3d46180c25e41fd79db3215f8b0cb24612d9/packages/api-client-next/src/index.ts#L109

There has to be a check, if the header exists in the response first.

niklaswolf:
### Description

closes #622 

### Type of change
 do not change the context token if no context-token header was provided in the API response

### DIFF
--- a/packages/api-client-next/src/index.ts
+++ b/packages/api-client-next/src/index.ts
@@ -109,8 +109,9 @@ export function createAPIClient<
         defaultHeaders["sw-context-token"] !==
           context.response.headers.get("sw-context-token")
       ) {
-        const newContextToken =
-          context.response.headers.get("sw-context-token") || "";
+        const newContextToken = context.response.headers.get(
+          "sw-context-token",
+        ) as string;
         defaultHeaders["sw-context-token"] = newContextToken;
         params.onContextChanged?.(newContextToken);
       }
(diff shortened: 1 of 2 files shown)


packages/api-client-next/src/index.ts
```diff
@@ -105,8 +105,9 @@ export function createAPIClient<
     // async onRequestError({ request, options, error }) {},
     async onResponse(context) {
       if (
+        context.response.headers.has("sw-context-token") &&
         defaultHeaders["sw-context-token"] !==
-        context.response.headers.get("sw-context-token")
+          context.response.headers.get("sw-context-token")
```
BrocksiNet, line 110: space needed?

niklaswolf: This was prettier `pnpm format`, not me ;)

patzick (approved):

PR #635, opened 2024-02-19
fix: ensure api client types does not resolve to `any`, refactor schema generic

patzick:
### Description

Fixes problem, when default types resolves to `any` as generic name in types was redefining another type causing problem like this:
image

fix required also to improve some tests which started showing type improvement from resolved to any


packages/api-client-next/src/tests/apiClient.test-d.ts
```diff
@@ -0,0 +1,17 @@
...
+describe("createApiClient type checks", () => {
+  it(`should make sure that defaut types are not resulving to "any"`, async () => {
+    const apiInstance = createAPIClient<operations, operationPaths>({
+      baseURL: "",
+      accessToken: "",
+    });
+    const result = await apiInstance.invoke("readContext get /context");
+    assertType<typeof result>({
+      // @ts-expect-error should not allow random property, and prevent `any` type
+      notExisting: "notExisting",
```
mkucmus, line 14: tough rules!

mkucmus (approved):
image

PR #625, opened 2024-02-19
feat: add possibility to apply language-id header on api-calls

niklaswolf:
### Description

closes #624 

### Type of change
New feature (non-breaking change which adds functionality)

### ToDo's

- [x] Changeset file provided [read more](https://github.com/shopware/frontends/blob/main/CONTRIBUTION.md#changelog-preparation)
- [ ] Documentation added/updated 
- [x] Unit-Tests added/updated
- [ ] E2E-Tests added/updated
- [x] Related Issue updated

PR #639, opened 2024-02-20
feat(api-client): headers management for api clients

### LINKED ISSUE #624: [FEATURE] new api client: possibility to set language-id
### Description

A possibility is needed to set the header `sw-language-id` for all API-requests.

### Use Case

A Shopware-Instance where you have a saleschannel with multiple languages.
By default, no language header is applied, so the default-language of the saleschannel is used.

### Proposed Solution

add an optional parameter `languageId` to `createAPIClient`

### Alternatives Considered

_No response_

### Additional Context

_No response_

patzick:
### Description
closes #624
extended functionality of #625 

Possibility of setting, displaying and changing default headers, which are added to every request within the API client.

### DIFF
--- a/packages/api-client-next/src/defaultHeaders.ts
+++ b/packages/api-client-next/src/defaultHeaders.ts
@@ -1,0 +1,71 @@
+/**
+ * Source: https://github.com/shopware/shopware/blob/trunk/src/Core/PlatformRequest.php#L16
+ */
+type RequestHeaderName =
+  | "sw-context-token"
+  | "sw-access-key"
+  | "sw-language-id"
+  | "sw-currency-id"
+  | "sw-inheritance"
+  | "sw-version-id"
+  | "sw-include-seo-urls"
+  | "sw-skip-trigger-flow"
+  | "sw-app-integration-id"
+  | "indexing-behavior"
+  | "indexing-skip";
+
+export type ClientHeaders = Record<RequestHeaderName | string, string>;
+
+export type ClientHeadersProxy = ClientHeaders & {
+  /**
+   * Set default headers for the client.
+   * Default headers are added to every request.
+   * If the header value is falsy, it will be removed from the headers.
+   *
+   * @example
+   * ```ts
+   * apiClient.defaultHeaders.apply({
+   *  "sw-language-id": "my-language-id",
+   * });
+   * ```
+   */
+  readonly apply: (headers: ClientHeaders) => void;
+};
+
+export function createHeaders(init: ClientHeaders): ClientHeadersProxy {
+  const _headers: ClientHeaders = {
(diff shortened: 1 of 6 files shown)


niklaswolf:
From my side, this is looking good and also covering the usecase from #624 / #625

packages/api-client-next/src/defaultHeaders.ts
```diff
@@ -0,0 +1,71 @@
...
+  | "sw-currency-id"
+  | "sw-inheritance"
+  | "sw-version-id"
+  | "sw-include-seo-urls"
+  | "sw-skip-trigger-flow"
+  | "sw-app-integration-id"
+  | "indexing-behavior"
+  | "indexing-skip";
+
+export type ClientHeaders = Record<RequestHeaderName | string, string>;
```
mkucmus, line 17: I like the way we provide built-in Shopware headers 👍🏼 

Also, I would like to have consistent way of building request headers as `ofetch` is build upon. So my proposition is to combine https://developer.mozilla.org/en-US/docs/Web/API/Headers/Headers constructor together with the headers that Shopware 6 supports (union) - you already defined in `RequestHeaderName` type.

thanks to this, working with header will be more natural for not only ofetch users, but fetch itself - that's a standard supported by all browsers at the moment.

mkucmus (commented):
list of supported headers finally here! 
one small question included :)

packages/api-client-next/src/defaultHeaders.ts
patzick: That could be the next step to support whole HeadersInit object (that could be also a tuple) so I went for a simpler object at this time and we can see if the usecases will require extending that

mkucmus: alright, time will tell 👍🏼

packages/api-client-next/src/index.ts
```diff
@@ -88,16 +91,14 @@ export function createAPIClient<
...
+  const defaultHeaders = createHeaders({
     "sw-access-key": params.accessToken,
     Accept: "application/json",
-  };
-
-  // protection from setting "null" or "undefined" as a token in API side
-  if (params.contextToken) {
-    defaultHeaders["sw-context-token"] = params.contextToken;
-  }
+    ...(params.contextToken && { "sw-context-token": params.contextToken }),
```
mkucmus, line 99: you've removed "protection from setting null or underined" above:

- what value will be sent in a very first request for `sw-context-token` when value it's not provided?
we need to ensure if it's not serialized to the empty string `""` because in that case we'd borrow a session under this context token which accepts every kind of string. maybe it's worth considering a backend issue to prevent this one edge case.

mkucmus (approved):
👌🏼

PR #642, opened 2024-02-20
feat(demo): aggregations, filters and limit for search

BrocksiNet:
### Description
- Changes how the filters work on the Search page
- Add new Aggregations for the count to the Search page
- Adds the count from the aggregations to properties or manufacturers in Template
- Use filter instead of "global" properties and manufacturer (aggregations are now updated)
- If you have properties or manufacturers in the URL the filters are opened automatically
- Added some select to set products (limit) per page
- Added ProductCardSkeleton on Search and Category Pages

### Type of change

### ToDo's
- [ ] Clean up and remove the functions from the search
- [x] Integrate the same logic into category
- [x] Create new e2e tests for all the new features

### Additional context
Sometimes the count does not match the products the customer sees.
Currently, we have a problem getting the aggregations into CMSPage listing elements. :warning:


packages/cms-base/components/public/cms/element/CmsElementProductListing.vue
```diff
@@ -72,12 +103,53 @@ setInitialListing(
...
+        <div
+          class="grid grid-cols-1 gap-4 md:gap-6 lg:gap-8 p-4 md:p-6 lg:p-8"
+          :class="{ 'lg:grid-cols-2': getTotalPagesCount > 1 }"
+        >
           <SwPagination
             :total="getTotalPagesCount"
             :current="Number(getCurrentPage)"
             @changePage="changePage"
           />
+          <div class="text-center place-self-center mt-2 lg:mt-0">
```
patzick, line 115: maybe let's show it only when there is more than minimal limit on page?

For example there is https://frontends-demo-git-feat-filter-count-8b03c7-shopware-frontends.vercel.app/CMS/ with 4 products and it looks unnecessary as pagination is also not visible 
image


or as alternative - show pagination with 1 page then, for consistency

packages/cms-base/components/public/cms/element/CmsElementProductListing.vue
```diff
@@ -35,19 +41,44 @@ const {
...
+  });
+  await changeCurrentPage(
+    1,
+    route.query as unknown as RequestParameters<"searchPage">,
+  );
+};
+
 const isProductListing = computed(
   () => props.content?.type === "product-listing",
 );
```
patzick, line 84: maybe we'd compare initialListing with currentListing (router one) and if there is a mismatch invoke search? That would at leat be decent workaround for Vercel query params cache issue. WDYT?

packages/cms-base/components/public/cms/element/CmsElementProductListing.vue
BrocksiNet: Pagination is now always shown also when there is only one page.

packages/cms-base/components/public/cms/element/CmsElementProductListing.vue
BrocksiNet: Implemented some ugly workaround. :sob:

patzick (commented):
So overall, it works well; I wouldn't make it bigger anymore. The last thing to fix would be blinking when the old result shows and the new one still needs to load. It would be better to show the loading skeleton there instead of bad results from cached SSR

BrocksiNet:
@patzick, @mkucmus, and @mdanilowicz I am done with that. 
Also, PR is getting a bit too big. So please check and let's merge that.

:rocket:

patzick:
image
there are some bugs in preview, I'll try to see what is the problem

packages/composables/src/useListing.ts
```diff
@@ -253,7 +254,7 @@ export function createListingComposable({
   //   ? `${contextName}(cms-${cmsContext})`
   //   : contextName;
 
-  const loading = ref(false);
+  const { loading } = useListingLoading();
```
patzick, line 257: we can have multiple listings, for example listing and preview with another listing, cannot use the same loading state for both. `createListingComposable` is a factory so creates shared listing state composable for dedicated purpose. That would break some flows. What is the reason behind it?

packages/composables/src/useListing.ts
```diff
@@ -4,6 +4,7 @@ import { getListingFilters } from "@shopware-pwa/helpers-next";
 import { useShopwareContext, useCategory } from "#imports";
 import ContextError from "./helpers/ContextError";
 import type { Schemas, RequestParameters } from "#shopware";
+import { useListingLoading } from "./useListingLoading";
```
patzick, line 7: direct imports prevents overriding, that should be imported by `#imports`. But see the comment below, as `useListingLoading` probably should not exist

patzick (changes requested):
We can work around the listing blinking later to not make this PR even bigger. The last thing that is to improve is to get rid of `useListingLoading` or discuss its usage further

PR #671, opened 2024-02-25
feat(demo): add payment form

itscark:
### Description

PR for [568](https://github.com/shopware/frontends/discussions/568)

### DIFF
--- a/examples/unzer-payment/components/CheckoutPaymentUnzerCreditCardPaymentHandler.vue
+++ b/examples/unzer-payment/components/CheckoutPaymentUnzerCreditCardPaymentHandler.vue
@@ -1,0 +1,202 @@
+<script setup lang="ts">
+import type { Schemas } from "#shopware";
+import { defu } from "defu";
+import { useCmsTranslations } from "@shopware-pwa/composables-next";
+
+type Translations = {
+  UnzerPayment: {
+    frame: {
+      creditCard: {
+        choose: string;
+        new: string;
+        number: string;
+        expiry: string;
+        cvc: string;
+        remember: string;
+      };
+    };
+  };
+};
+
+let translations: Translations = {
+  UnzerPayment: {
+    frame: {
+      creditCard: {
+        choose: "Choose a saved credit card",
+        new: "Use a new credit card",
+        number: "Card number",
+        expiry: "Expiry date",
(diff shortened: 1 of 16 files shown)


examples/unzer-payment/components/CheckoutPaymentUnzerCreditCardPaymentHandler.vue
```diff
@@ -0,0 +1,202 @@
...
+  "orderButtonDisabled",
+  ref(true),
+);
+
+async function createResource(): Promise<void> {
+  console.debug("[unzer] create:resource", creditCard.value);
+
+  if (creditCard.value !== null) {
+    try {
+      const response: { id: string } = await creditCard.value.createResource();
```
mkucmus, line 49: image

examples/unzer-payment/composables/useUnzer.ts
```diff
@@ -0,0 +1,70 @@
...
+
+      const { $i18n } = useNuxtApp();
+      unzerInstance.value = new unzer(publicKey, {
+        locale: $i18n.locale.value,
+      });
+    }
+  }
+
+  onBeforeMount(() => {
+    useHead({
```
mkucmus, line 50: attaching the unzer.js script is too lazy. It's not being loaded in the test env: https://stackblitz.com/edit/github-c9t5on?file=nuxt.config.ts (backend is temporary, we can talk about it in DM later on)

it causes the error in here: https://github.com/shopware/frontends/pull/671/files#diff-a73f8d424b5c2834bf69de34a0cd276893eb9a66a9a84ef96553079f816ab0e1R49

I can see two solutions for that:
- I would wrap useHead things in a function and export it in the composable to be used directly in the setup function of the component - then run it in `onBeforeMount` lifecycle hook in a component.
- use a plugin (server + client) to register a script for specific page, in client plugin initialize the unzer instance

wdyt?

mkucmus (changes requested):
Hey! thank you very much for a great PR!

I've checked twice and ran into a problem only with missing unzer instance (or sth related) so I suggested two solutions for that. 

anyway, the overall is really great and using a dynamic import for component relying on its name in the backend is a good move 💪🏼 
I was able to finish the flow and it worked 💙

PR #699, opened 2024-03-03
feat: b2b quotes example

### LINKED ISSUE #543: [FEATURE] Quote Management - B2B Components
### Description

We should provide Examples and Composables of how to use the **Quote Management** Feature for B2B Components.

### Use Case

Feature Quote Management of B2B Components should be usable with Composable Frontends.

### Proposed Solution

Similar to the **Quick Order** Example [here](https://github.com/shopware/frontends/pull/436). But, it is better to have at least account management and adding to cart implemented, so maybe it's worth to extend the `vue-temo-store` template to give a whole experience.

### Acceptance criteria:

**The best benchmark is our demo instance, where any customer can have enabled "quote management" feature.**

1. customer with quote management enabled can see "request quote" button in the minicart, when something is already added to cart (find out how to distinguish such customers from others - extension? customer object extended?).
2. sending a request is possible with optional message included - message to the shop managers/admins.
3. In "my account" view, there is another tab/sub page showing the list of requested quotes, with status, description, and action buttons (edit/delete/see details) 
4. detail view of quote request must have displayed useful data (also like message history) and actions to proceed (like place order on specific state)
5. there is also possible to generate documents, so downloading them must be also possible

### Alternatives Considered

_No response_

### Additional Context

More description you can find [here](https://shopware.atlassian.net/wiki/spaces/PRODUCT/pages/19527827759/B2B+Component+Quote+Management). :lock:

mdanilowicz:
### Description

closes #543


packages/helpers/src/index.ts
```diff
@@ -14,3 +14,4 @@ export * from "./checkout";
 export * from "./languages";
 export * from "./relativeUrlSlash";
 export * from "./urlIsAbsolute";
+export * from "./b2bQuote Management/canUseQuoteActions";
```
BrocksiNet, line 17: Why do we have a space here? 
Looks strange also in `helpers/src/b2bQuote Management`.
I would suggest removing the space.

mdanilowicz: by mistake, fixed

patzick (approved):
Ok to merge after fixing tests 👌

PR #757, opened 2024-03-14
fix: workflows fix for failed job check

kstala:
Unification of tests for Stackblitz projects (running tests for stackblitz templates in the same way as for examples) + 
fixes for failed-job-checker for new workflow names
image


.github/workflows/check-stackblitz-templates.yml
```diff
@@ -1,28 +1,42 @@
 name: Stackblitz templates
-run-name: Playwright stackblitz templates 🚀
-on: 
-  workflow_dispatch:
-  schedule:
-    - cron: '59 23 * * *'
+
+
+run-name: Playwright is testing stackblitz templates 🚀
```
patzick, line 4:
```suggestion
run-name: Playwright stackblitz templates 🚀
```
let's stay with the name for consistency

patzick (approved):

PR #795, opened 2024-03-23
fix(cms-base): push errors from add to cart API call

### LINKED ISSUE #782: [BUG] errors during adding a product to cart aren't displayed

### Current Behavior

When I add a product to cart with some impossible quantity I'm able to add only one of it, but the message from the response is not displayed
image

### Expected Behavior

the errors from the response are displayed, so I know what happened and why I don't have exact number of items in my cart

### Steps To Reproduce

1. go to `/detail/3c40bbdab5a94212b7834f879ab34539` route
2. add 9999 quantity
3. see the only one product was added
4. I can see errors object with proper warning attached - in `line-item` endpoint response

### Environment

```markdown
- OS:
- Node:
- pnpm:
```

### Anything else?

_No response_

florianliebig:
### Description

closes #782 

### Type of change

### ToDo's

- [ ] Changeset file provided [read more](https://github.com/shopware/frontends/blob/main/CONTRIBUTION.md#changelog-preparation)

- [ ] E2E-Tests added/updated

2024-03-23 at 19 06

### DIFF
--- a/packages/cms-base/components/SwProductAddToCart.vue
+++ b/packages/cms-base/components/SwProductAddToCart.vue
@@ -5,7 +5,7 @@ import { toRefs } from "vue";
 import { defu } from "defu";
 import type { Schemas } from "#shopware";
 
-const { pushSuccess } = useNotifications();
+const { pushSuccess, pushError } = useNotifications();
 const props = defineProps<{
   product: Schemas["Product"];
 }>();
@@ -33,7 +33,14 @@ const { product } = toRefs(props);
 const { addToCart, quantity } = useAddToCart(product);
 
 const addToCartProxy = async () => {
-  await addToCart();
+  const res = await addToCart();
+
+  if (res.errors) {
+    const errors = Object.values(res.errors).map((error) => error.message);
+    pushError(errors.join(". "));
+    return;
+  }
+
   pushSuccess(
     `${props.product?.translated?.name} ${translations.product.addedToCart}`,
   );
--- a/.changeset/sour-lobsters-bow.md
+++ b/.changeset/sour-lobsters-bow.md
@@ -1,0 +1,5 @@
+---
+"@shopware-pwa/cms-base": minor
+---
+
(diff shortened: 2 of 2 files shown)


packages/cms-base/components/SwProductAddToCart.vue
```diff
@@ -33,7 +33,15 @@ const { product } = toRefs(props);
 const { addToCart, quantity } = useAddToCart(product);
 
 const addToCartProxy = async () => {
-  await addToCart();
+  const res = await addToCart();
+  const errors = Object.values(res.errors);
```
mdanilowicz, line 37: Cart errors should be transformed and translated. Please check how it's done in the demo store 

https://github.com/shopware/frontends/blob/main/templates/vue-demo-store/components/product/ProductAddToCart.vue#L12

mdanilowicz (changes requested):
https://github.com/shopware/frontends/pull/795/files#r1538635825

packages/cms-base/components/SwProductAddToCart.vue
florianliebig: Ah nice thanks! Do we also need to handle any translations or can we rely that the API Response is already translated?

BrocksiNet: Hello :wave:

I have created a new Branch from your PR, and changed the part @mdanilowicz mentioned.

Look here: https://github.com/shopware/frontends/commit/d8e28c8b0eb63976194b4e63e8679c677a34de64

But to be honest I do not like the part about the quantity. Currently, we would need to search in the line items for the quantity that was added to the cart. Maybe a more generic error message would be easier.

mdanilowicz: > Ah nice thanks! Do we also need to handle any translations or can we rely that the API Response is already translated?


Unfortunately, all API errors are in English, even if the context language is different.  So, we have to translate it "locally"

Issue #799, opened 2024-03-25
[BUG]  Incorrect format of images on the product card


kstala:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

For some part of the product card there is a problem with displaying product images, like 
https://frontends-demo.vercel.app/detail/05a7298bc7314ca5b9011c5cabfac217
<img width="601" alt="image" src="https://github.com/shopware/frontends/assets/30267600/283bc40b-3d3e-4c59-a383-3c63097621ff">

### Environment

```markdown
- OS:
- Node:
- pnpm:
```

PR #810, opened 2024-03-27
fix(cms): more beauty image gallery

BrocksiNet:
### Description

After a hard reload the image gallery was nice but with client navigation, it often happens that the images on PDP looked ugly. With those changes, this is fixed and always "beautiful". :joy: 

closes: #799

### Type of change

This (**current**):
image

vs

This (**fixed**):
image


patzick (approved):
NIce! 💅

PR #857, opened 2024-04-08
feat(api): maintenance mode

### LINKED ISSUE #469: [FEATURE] Maintenance mode
### Description

Now that maintenance mode is supported via store api (since Shopware version 6.5.7.0), it would be nice to have an example of how best to handle it in Frontends apps. Maybe to add different layout if its maintenance mode?

Also, not sure how maintenance mode should work with these lines uncommented:
https://github.com/shopware/frontends/blob/main/templates/vue-demo-store/app.vue#L24
Because then error response code is always 500

### Use Case

Maintenance mode in Frontends app

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

mdanilowicz:
### Description

 closes #469


packages/nuxt3-module/plugin.ts
```diff
@@ -36,6 +39,16 @@ export default defineNuxtPlugin((NuxtApp) => {
         sameSite: "lax",
       });
     },
+    onErrorHandler(response) {
+      const error = isMaintenanceMode(response._data?.errors ?? []);
+      console.log("aaa", response._data?.errors);
```
mkucmus, line 44: you mean "American Automobile Association" by this? :) please remove

patzick: @mkucmus easier to add suggestion, we can commit without checking out the code :)
```suggestion
```

patzick (approved):
Error and success callbacks with became hooks with #871, but we'll adjust it there :)

Issue #767, opened 2024-03-19
[BUG] bfcache Issue


BrocksiNet:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

**That's the problem**
![image](https://github.com/shopware/frontends/assets/3763023/7650918d-8a12-4c7d-86e3-ef1e54c16529)

**That could be a solution**
https://github.com/nuxt/nuxt/pull/23667

### Expected Behavior

The Issue is not shown anymore in Lighthouse for the demo-store.
The demo-store works like before without any errors.

### Anything else?

Another related https://github.com/shopware/frontends/issues/539

InstantlyMoist:
Currently having the same issue, no dependencies seem to interfere with it.

BrocksiNet:
I was researching a lot today about that issue. :mag: 

I assume that the issue comes from [nuxt i18n module](https://v8.i18n.nuxtjs.org/getting-started/setup).

So I created a empty nuxt/vite project with that command:
`pnpm dlx nuxi@latest init <project-name>`

Testing this in **preview mode** is important because Vite is opening a WebSocket in dev mode, which can also cause an issue with bfcache.

So the steps for **testing** are:
1. pnpm build
2. pnpm preview
3. Open the Page in a Chrome Browser
4. Open Dev Tools -> Application -> Back/Forward Cache -> Click on Test Button

**Result:**
- Blank Nuxt/Vite Project -> No bfcache Issue :heavy_check_mark: 
- Blank Nuxt/Vite Project + i18n module -> bfcache Issue :x: 

Somehow I assume it has something to do with `detectBrowserLanguage` function. But not sure about that. :confused:

BrocksiNet:
Today, I tested again with the `Blank Nuxt/Vite Project + i18n module` and can not reproduce the `bfcache` Issue anymore. This is great 🎉 

So, it would be super cool to eliminate that problem in the **demo-store** template. 💡 

Currently, we are using `@nuxtjs/i18n` in version `8.3.1`, and I guess we need to update package.json to `^8.5.5`.

But when I execute this command:
`pnpm run generate && pnpm run preview`

I got a lot of warnings and errors. So, we need to fix the generate command first.

BrocksiNet:
So when  I update @nuxtjs/i18n and comment out this part in app.vue:
`//useBroadcastChannelSync();`

We do not have any `bfcache` Issue anymore:
[...]

PR #866, opened 2024-04-11
feat(vue-demo): broadcasting

### LINKED ISSUE #788: [BUG]  logout not logout from whole page

### Current Behavior

We open two browser tabs, one for the home page and one for the shopping cart. If we are on the basket tab and click log out in the profile, we are still logged in on the second tab. We can then go to the profile, make changes etc

### Expected Behavior

After clicking logout, we should be logout from all page 

### Steps To Reproduce

1 log in to the site
2, open e.g. the home page on the first tab, then open the second tab with the next subpage (cart/product, whatever)
3 on the first tab, go to your profile and click log out
4 on the second tab, we are still logged in, refreshing does nothing

### Environment

```markdown
- OS: linux/windows 10, I tested it on a few systems
- Node: 20
- pnpm: -
```

### Anything else?

I don't know if this is a bug in your application or a bug in my project, but I haven't seen any changes in my project's code related with logout function

mdanilowicz:
### Description

 closes #788

### DIFF
--- a/templates/vue-demo-store/nuxt.config.ts
+++ b/templates/vue-demo-store/nuxt.config.ts
@@ -5,6 +5,9 @@ export default defineNuxtConfig({
     "@shopware-pwa/cms-base",
   ],
   runtimeConfig: {
+    public: {
+      broadcastChannelName: "shopware-composable-frotends",
+    },
     shopware: {
       /**
        * SSR Shopware Endpoint
--- a/templates/vue-demo-store/app.vue
+++ b/templates/vue-demo-store/app.vue
@@ -14,7 +14,11 @@ useHead({
     lang: "en",
   },
 });
+const { consume } = useBroadcastConsumer();
 
+const {
+  public: { broadcastChannelName },
+} = useRuntimeConfig();
 const { apiClient } = useShopwareContext();
 const sessionContextData = ref<Schemas["SalesChannelContext"]>();
 sessionContextData.value = await apiClient.invoke("readContext get /context");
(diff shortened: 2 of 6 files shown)


BrocksiNet:
After logging out, the mini-cart is not updated. :thinking: 

Also, I am asking myself if this will make that issue https://github.com/shopware/frontends/issues/767 even worse, so we break the bfcache with that new BroadCastingChannel "again".

mdanilowicz:
> After logging out, the mini-cart is not updated. 🤔

Same behaviour we have on the main. It looks like a bug. I created a new task for that

> Also, I am asking myself if this will make that issue #767 even worse, so we break the bfcache with that new BroadCastingChannel "again".

Why? It is not related to the navigation

templates/vue-demo-store/nuxt.config.ts
```diff
@@ -5,6 +5,9 @@ export default defineNuxtConfig({
     "@shopware-pwa/cms-base",
   ],
   runtimeConfig: {
+    public: {
+      broadcastChannelName: "shopware-composable-frotends",
+    },
```
patzick, line 10: isn't that complicating config? It's an internal template mechanism; maybe it would be simpler to keep it with the mechanism itself?

patzick (commented):
What about not refreshing the session context in every tab but passing in the current context instead? This way, every tab shares the same context and no need to invoke refresh X times.

mdanilowicz:
> What about not refreshing the session context in every tab but passing in the current context instead? This way, every tab shares the same context and no need to invoke refresh X times.

Done

templates/vue-demo-store/nuxt.config.ts
mdanilowicz: We can have many broadcast channels. I think this is the easiest way to change the name by the config

patzick (approved):

PR #893, opened 2024-04-21
feat: added Pagination for Wishlist

khanSoliheen:


mkucmus (changes requested):
hey, looks really good 🙌🏼  

👉🏼 Could you please adjust the tests because they are failing

BrocksiNet:
@mkucmus, @kstala, and @khanSoliheen, I fixed the Test for the `useWishlist` composable.

Now, we have 100% coverage for the `useWishlist` composable. :tada:

packages/composables/src/useCustomerOrders.ts
```diff
@@ -13,11 +13,22 @@ export type UseCustomerOrdersReturn = {
...
+    limit: number | string,
+  ): Promise<void>;
   /**
    * Fetches the orders list and assigns the result to the `orders` property
    */
   loadOrders(parameters?: Schemas["Criteria"]): Promise<void>;
+  /**
+   * Current page number
+   */
+  getCurrentPage: ComputedRef<number>;
```
patzick, line 27: it's computed property, not getter function, we should rename it
```suggestion
  currentPage: ComputedRef<number>;
```

packages/composables/src/useCustomerOrders.ts
```diff
@@ -13,11 +13,22 @@ export type UseCustomerOrdersReturn = {
...
+  /**
+   * total pages count
+   */
+  getTotalPagesCount: ComputedRef<number>;
```
patzick, line 31: same here
```suggestion
  totalPagesCount: ComputedRef<number>;
```

packages/composables/src/useSyncWishlist.ts
```diff
@@ -27,10 +30,11 @@ export type UseSyncWishlistReturn = {
   /**
    * Wishlist items count
    */
-  count: ComputedRef<number>;
+  totalWishlistItemsCount: Ref<number>;
```
patzick, line 31: I'm afraid that is an unnecessary breaking change; why do you feel the need to change it?

Also it should be a computed property

packages/composables/src/useSyncWishlist.ts
```diff
@@ -89,14 +96,13 @@ export function useSyncWishlist(): UseSyncWishlistReturn {
   }
 
   const items = computed(() => _wishlistItems.value);
-  const count = computed(() => items.value.length);
```
patzick, line 92: Count could just show `totalWishlistItemsCount` value here

packages/composables/src/useWishlist.test.ts
```diff
@@ -1,37 +1,159 @@
...
-    },
-  }));
-
-  it("mergeWishlistProducts", async () => {
+  it("mergeWishlistProducts", () => {
+    vi.mocked(useUser).mockReturnValue({
+      isLoggedIn: computed(() => false),
+      isGuestSession: computed(() => true),
+      // @ts-ignore
+      canSyncWishlist: { value: false },
```
patzick, line 49: ts-ignore is not a solution, useUser doesn't have `canSyncWishlist` so shouldn't be in the mock
```suggestion
```

packages/composables/src/useWishlist.test.ts
```diff
@@ -1,37 +1,159 @@
...
+    });
+    vi.mocked(useSyncWishlist).mockReturnValue({
+      getWishlistProducts: () => undefined,
+      // @ts-ignore
```
patzick, line 53: same - no ts-ignore, we need to make sure types are correct in composables

patzick (changes requested):
Thanks for the contribution! 

Few suggestions for the API of composables + ts-ignore to be removed from tests :)

BrocksiNet:
@mkucmus and @patzick fixed the ts errors, updated tests, tested orders pagination, and fixed it.

You can review it again. I added @kstala for testing, also.

mdanilowicz (approved):

kstala:
@BrocksiNet can you take a look at this, I have 2 problems (both cases involve guest user) 
- as a guest user, adding a product to the wishlist doesn't update the number on the heart wishlist icon
https://github.com/shopware/frontends/assets/30267600/0548bf58-c9ff-4530-a59b-5bb52a191a9e

- pagination for wishlist doesn't work for guest users
Uploading Screen Recording 2024-07-08 at 09.29.05.mov…

packages/composables/src/useUser.ts
```diff
@@ -151,6 +155,10 @@ export function useUser(): UseUserReturn {
   const salutation: Ref<Schemas["Salutation"] | null> = ref(null);
   const user = computed(() => _user.value);
 
+  const canSyncWishlist = computed(
+    () => isLoggedIn.value && !isGuestSession.value,
+  );
```
patzick, line 160: that is not what this condition tells us. Also both `isLoggedIn` and `isGuestSession` are already exposed by the `useUser` let's revert changes in this file and check it in actual wishlist

packages/composables/src/useCustomerOrders.ts
```diff
@@ -39,12 +58,35 @@ export function useCustomerOrders(): UseCustomerOrdersReturn {
...
-  const changeCurrentPage = async (pageNumber: number | string) =>
-    await loadOrders({ page: +pageNumber });
+  const loadTotalOrdersCount = async (
+    parameters: Schemas["Criteria"] = {},
+  ): Promise<void> => {
+    const fetchedOrders = await apiClient.invoke("readOrder post /order", {
+      body: parameters,
+    });
+    totalOrderItemsCount.value = fetchedOrders.data.orders.total ?? 0;
+  };
```
patzick, line 68: can't we do the same with loadOrders method? It seems unnecessary to fetch it again only for the count

templates/vue-demo-store/pages/account/order.vue
```diff
@@ -33,6 +97,10 @@ await useAsyncData("getOrders", () => {
     ],
   });
 });
+
+await useAsyncData("getTotalOrders", async () => {
+  loadTotalOrdersCount();
+});
```
patzick, line 103: we're already loading orders, so total count should already be available

templates/vue-demo-store/pages/account/order.vue
```diff
@@ -20,8 +56,36 @@ useBreadcrumbs([
...
+const limit = ref(route.query.limit ? Number(route.query.limit) : defaultLimit);
+
+const changeLimit = async (limit: Event) => {
+  const select = limit.target as HTMLSelectElement;
+
+  await router.push({
+    query: {
+      ...route.query,
+      limit: select.value,
+      p: defaultPage,
```
patzick, line 68: shouldn't the page be the current page or the same param from router?

.changeset/tough-queens-swim.md
```diff
@@ -0,0 +1,6 @@
+---
+"vue-demo-store": minor
+"@shopware-pwa/composables-next": minor
```
patzick, line 3: composable changeset should contain exact changes to the API of composables, not just a description

patzick (changes requested):
Apart of @kstala comment, there are also some things in the code to improve before merge :)

templates/vue-demo-store/pages/account/order.vue
BrocksiNet: That is false because the total count differs when you set a limit.
But this could also be a bug in the admin API endpoint.

patzick: sounds like a bug in the API or criteria differs, the total should only change when the criteria changes, the limit cannot change the `total`

templates/vue-demo-store/pages/account/order.vue
BrocksiNet: nope, changing the limit will always bring you back to the default page

packages/composables/src/useUser.ts
BrocksiNet: moved again

kstala (approved):

packages/composables/src/useSyncWishlist.ts
```diff
@@ -28,9 +31,15 @@ export type UseSyncWishlistReturn = {
    * Wishlist items count
    */
   count: ComputedRef<number>;
+  /**
+   * Current page number
+   */
+  currentPage: Ref<number>;
```
patzick, line 37:
```suggestion
  currentPage: ComputedRef<number>;
```

packages/composables/src/useWishlist.ts
```diff
@@ -13,7 +15,10 @@ export type UseWishlistReturn = {
   /**
    * Get products list added to wishlist
    */
-  getWishlistProducts(): void;
+  getWishlistProducts(
+    page?: number,
+    query?: Schemas["Criteria"],
```
patzick, line 20: page is a part of the endpoint body params, so we should just have the `query`

packages/composables/src/useWishlist.ts
```diff
@@ -23,9 +28,21 @@ export type UseWishlistReturn = {
    */
   items: ComputedRef<string[]>;
   /**
-   * Wishlist items count
+   * Current page number
+   */
+  getCurrentPage(): ComputedRef<number>;
```
patzick, line 33: same here - should be only a computed property, not a function
```suggestion
  currentPage: ComputedRef<number>;
```

packages/composables/src/useWishlist.ts
```diff
@@ -23,9 +28,21 @@ export type UseWishlistReturn = {
...
+  /**
+   * total pages count
+   */
+  getTotalPagesCount(limit?: number): ComputedRef<number>;
```
patzick, line 37: you shouldn't return computed property from a function.

Btw you already have `ref` called `limit` - you can expose it for use and change and operate in computed properties on that value


```suggestion
  totalPagesCount: ComputedRef<number>;
```

templates/vue-demo-store/app.vue
```diff
@@ -111,6 +111,7 @@ if (languages.value?.elements.length && router.currentRoute.value.name) {
 
 onMounted(() => {
   refreshCart();
+  // this request is needed to show the wishlist heart on every category page without any limit / page
```
patzick, line 114: so why do we need a pagination here if all wishlist items are already loaded?

packages/composables/src/useCustomerOrders.ts
```diff
@@ -13,11 +13,22 @@ export type UseCustomerOrdersReturn = {
    *
    * In order to change a page with additional parameters please use `loadOrders` method.
    */
-  changeCurrentPage(pageNumber: number | string): Promise<void>;
+  changeCurrentPage(
+    pageNumber: number | string,
+    limit: number | string,
```
patzick, line 18: is the limit necessary? It would be a breaking change

templates/vue-demo-store/app.vue
BrocksiNet: This is to show if a product is already on the wishlist.
Or in other words, to show the :heart:

packages/composables/src/useCustomerOrders.ts
BrocksiNet: Somehow, we need to pass the limit to the request.
We could also rely on the limit that is set by `total pages`, but that will indirectly couple this.

BrocksiNet:
FYI: I removed the wishlist call from `onMounted` because it is not needed information for SSR. I now make sure that the wishlist call is only executed once per page, not twice on the wishlist page (like it was before). So overall, this will improve the SSR times a bit.

patzick (approved):

PR #903, opened 2024-04-25
feat(api-gen): blob type support

mdanilowicz:
### Description

Add Blob type support


BrocksiNet:
How to validate this via a test?

mkucmus (approved):

PR #913, opened 2024-05-07
feat: keyboard navigation

BrocksiNet:
### Description

It allows you to use ArrowDown, ArrowUp, and Escape to navigate the top navigation.

### Type of change

### Additional context


patzick (changes requested):
Works good, the only thing is styling issue while submenu object is focused 
image

BrocksiNet:
Closed. See https://github.com/shopware/frontends/pull/922.

PR #918, opened 2024-05-10
feat(nuxt3-module): secure Sw-Context-Token cookie

jhony1104:
### Description

The object of this PR is to set the secure flag for the `Sw-Context-Token` cookie if possible.

### Type of change

### Additional context

This came up as a minor issue in a pentest of an app using the nuxt3-module.


patzick (approved):
Thank you! :)

PR #922, opened 2024-05-13
feat: keyboard navigation and improved mobile sidebar

BrocksiNet:
### Description

- It allows you to use ArrowDown, ArrowUp, and Escape to navigate the top navigation.
- You can now scroll inside the mobile top navigation

### Type of change

### Additional context

### DIFF
--- a/.changeset/young-files-laugh.md
+++ b/.changeset/young-files-laugh.md
@@ -1,0 +1,5 @@
+---
+"@shopware-pwa/cms-base": patch
+---
+
+Moved the svg 3d media placeholder to the vue component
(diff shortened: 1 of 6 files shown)


BrocksiNet:
@patzick the focus style bug is fixed here also.

patzick:
Thanks @BrocksiNet 
one more thing I've noticed - a small inconsistency with the navigation

Reproduction:
- tab through the first menus to the "Cooking..."
- now two arrows UP allows you to go back to "Summer trends", but two arrows DOWN will bring you to the submenu instead back to the "Cooking..." menu

What would you say to use arrows UP/DOWN in the current menu (either main or submenu) and then enter child navigation only by the TAB?

BrocksiNet:
Closed, see https://github.com/shopware/frontends/pull/930. :joy:

PR #927, opened 2024-05-14
fix(examples): express checkout

jhony1104:
### Description

fixed express-checkout example.

It is still not working, since the connected Shopware Server does not offer a payment method with shortName `pay_pal_payment_handler`.

### Type of change


mkucmus (approved):
nice 🎖️ !

PR #928, opened 2024-05-15
fix(gen-api): additional properties

mdanilowicz:
### Description
From
```
    OrderRouteResponse: {
      paymentChangeable?: GenericRecord;
    };
```
To
```
    OrderRouteResponse: {
      paymentChangeable?: {
        [key: string]: boolean;
      };
    };
```


mkucmus (approved):
nice catch!

PR #930, opened 2024-05-16
feat: keyboard navigation and improved mobile sidebar

BrocksiNet:
### Description

- It allows you to use ArrowDown, ArrowUp, and Escape to navigate the top navigation.
- You can now scroll inside the mobile top navigation

### Type of change

### Additional context


BrocksiNet:
@patzick and @kstala:

The arrow navigation is changed, as Patryk suggested.
Also, when you jump out of the navigation or a new page is loaded, the navigation is now closed.

Over and out :pray:

PR #977, opened 2024-06-05
feat(demo-store): link back to product page from the minicart

### LINKED ISSUE #559: [FEATURE] Link back to product from Mini-Cart and Cart-Page
### Description

Currently, we do not have any link from the line item in the mini-cart or cart page back to the product page.

### Use Case

Improve the usability for the user.

### Proposed Solution

With the `referencedId` we already know the corresponding `productId` but we do not know the nice speaking URL for that product. So how do we solve this without additional requests? Is this possible? I could imagine that we use the `payload` of the line item during the add item to the cart action/event.

### Alternatives Considered

Consider moving the function into a reusable composable.

### Additional Context

The solution needed for the demo-store template.

khanSoliheen:
### Description

closed #559


templates/vue-demo-store/components/checkout/CheckoutCartItem.vue
```diff
@@ -5,13 +5,15 @@ const props = withDefaults(
   defineProps<{
     cartItem: Schemas["LineItem"];
     maxQty?: number;
+    // eslint-disable-next-line vue/require-default-prop
+    controller?: ReturnType<typeof useModal>;
```
patzick, line 9: This rule exists for a reason :) Disabling that you're silencing linter but causing runtime error. For example in checkout clicking it is causing error:
image

I'll propose a different solution in summary

templates/vue-demo-store/components/checkout/cart/ProductItem.vue
```diff
@@ -92,7 +99,11 @@ const removeCartItem = async () => {
       <div
         class="flex flex-col lg:flex-row justify-between text-base font-medium text-secondary-900"
       >
-        <h3 class="text-base" data-testid="cart-product-name">
+        <h3
+          class="text-base cursor-pointer"
+          data-testid="cart-product-name"
+          @click="navigateToItemDetailPage(cartItem.id)"
+        >
```
patzick, line 106: It should be wrapped with NuxtLink + additional @click event. Otherwise causes accessibility issue. Look inside LayoutSideMenu component for the inspiration

patzick (changes requested):
Thank you for the contribution! :)
We need to improve few things before we can merge that. Passing a `conroller` deep inside props makes it hard to manage. Instead let's:

1. create a new shared controller in `useModal.ts` file. Simliar to `useSideMenuModal` and we can call it `useMiniCartModal`
2. replace `sidebarController` in `LayoutHeader` with this and remove passing it through props but use `useMiniCartModal` directly where needed
3. then in `ProductItem` you can also directly use `useMiniCartModal` and close it on link click

What do you think of that approach?

templates/vue-demo-store/components/checkout/CheckoutCartItem.vue
khanSoliheen: It is not required now.

templates/vue-demo-store/components/checkout/cart/ProductItem.vue
khanSoliheen: I've implemented the functionality to close the miniCartModal when the NuxtLink click event occurs. Please review the changes and let me know if anything needs adjustment. Thank you!

patzick (approved):

Issue #941, opened 2024-05-24
[FEATURE] Standardise request handling in composables


freb97:
### Description

Currently, when using composables to update or retrieve data from the API, the usage is different with each composable.

For example, when using the `useCustomerPassword` composable and its functions `updatePassword` or `resetPassword`, errors are written to an object `errors` that is returned by the composable. This object is typed, so that you should be able to obtain the individual errors by accessing the object like this: `errors.resetPassword.forEach((error) => { // custom logic here })`. When using this, the errors object actually looks like this: `errors.resetPassword.errors.forEach((error) => { // custom logic here })`. The `updatePassword` and `resetPassword` also return a Promise of type boolean.

This is different from the usage of for example the `useUser` composable. If you want to update a user's personal information, you could call the `updatePersonalInfo` function, which does not return a Promise of type boolean but of type void. Also the function does not wrap the API-Call in a try-catch, so the error handling looks very different, you have to try-catch on your own and there is no errors object that gets exposed by the composable.

Also some functions do not have a consistent signature. The function `refreshUser` of the `useUser` composable has an argument called 'params', which is the criteria that would be used for the request, but the function signature in the `UseUserReturn` object does not contain any arguments. This way, typescript will not allow calling the function with any criteria except when explicitly ignoring it using `@ts-ignore`.

### Use Case

Standardising the way requests are wrapped in composable functions would help users by not constantly having to check for new ways to handle possible errors. It would also make it easier to use the composables overall, as it eliminates the need to check what the function does under the hood, if every similar function behaves in the same way.

### Proposed Solution

* Decide on a way that errors in requests are handled and use it across all composables consistently
* Check for missing or differing function signatures and fix them
* Fix the typing for the 'errors' object returned by some composables

patzick:
This was the last thing before the major update; thank you for spending the time to write this down!

Different approaches exist because some composables exist even in the pre-frontend phase (from shopware pwa). Cleaning this up should bring more consistency and, as you said, predictiveness - hence better DX.


### Acceptance criteria
- remove `error` fields from the composables (unless it makes sense; in that case, a comment on the method definition for clarification
- usual rule of thumb - errors are thrown by the methods directly, not caught inside the composables (otherwise commented like above)
- check error handling and provide some helper function for that if needed
- whenever possible - return the value from API, like in the example above - `updatePersonalInfo` instead of void should return whatever API is returning as a response


@freb97 would that sum it all?

freb97:
Hey @patzick,

[...]

PR #978, opened 2024-06-06
refactor!: change useCustomerPassword error handling

patzick:
### Description

Part of #941

Breaking the API of `useCustomerPassword` for consistency in error handling.

### Type of change

Breaking change


mkucmus (approved):
DIY error handling:
- trust on your types definition (@shopware/api-client 🚀) - we don't int to intercept them and shadow the final output
- catch the errors only if necessary (logic layer provided by us) 
- handle errors in UI part on your own

PR #987, opened 2024-06-10
feat(api-client): updated types overrides for 6.6.3.0

patzick:
### Description

Updated overrides to new schema from 6.6.3.0.

This reduces the number of minimal TODOs on API schema from over a 100 to **53**

Still work to do but we're getting there!


mkucmus (approved):
image

PR #991, opened 2024-06-10
refactor(composables): return data from the API in most of the methods

### LINKED ISSUE #941: [FEATURE] Standardise request handling in composables

patzick:
### Description

closes #941 

### Type of change

Most of the methods return now the data from API, even if storing data locally as well. Error handling now more consistent and predictable.

### DIFF
--- a/.changeset/big-plums-hunt.md
+++ b/.changeset/big-plums-hunt.md
@@ -1,0 +1,16 @@
+---
+"@shopware-pwa/composables-next": minor
+---
+
+Few changes in composables API to access data returned from the backend:
+
+- `useAddress` - `loadCustomerAddresses` returns addresses now
+- `useCart` - `removeItem` returns updated cart
+- `useCartItem` - `removeItem` returns updated cart, similar to `useCart`
+- `fetchCountries` - returns countries with the response
+- `useNewsletter` - `getNewsletterStatus` returns full response from the API
+- `useOrderDetails` - `loadOrderDetails` returns order details now, `cancel` returns order state, `changePaymentMethod` returns success response info
+- `changePaymentMethod` - `changePaymentMethod` returns success response info now
+- `useProductReviews` - `loadProductReviews` returns reviews response now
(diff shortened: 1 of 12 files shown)


patzick:
codspeed regression result is not correct, this PR is not changing any logic in this area

mkucmus (approved):

PR #998, opened 2024-06-11
fix(examples): ignore amazon pay due to out of context module

mkucmus:
### Description

the example cannot run on stackblitz because the module is out of context a stackblitz runtime (../src/module).

### Type of change


kstala:
I ran the job directly from this branch and the code example passed https://github.com/shopware/frontends/actions/runs/9512422763

kstala (approved):

patzick (approved):

PR #999, opened 2024-06-11
docs: new reference page

### LINKED ISSUE #621: [FEATURE] `@shopware/api-client` reference
### Description

documentation of public API of [@shopware/api-client](https://www.npmjs.com/package/@shopware/api-client) package, including all possible types.

### Use Case

do what `typer` package normally does.

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

mkucmus:
preview PR for new structure of packages reference

closes #621


patzick (changes requested):
Looks great!
A few minor improvements/questions:
- address `/packages/api-client-next.html` -> can we name it without `next`?
- you needed to rename the file `createApiClient` but missed the admin file, so the source is a dead link

Every header should be a link (and is but it's very hard to click it, you need to find dark pixels it on the side), let's fix it to easily link to specific composables:
image


Additionally - there are a lot of warnings in the console:
image

mkucmus:
> Looks great! A few minor improvements/questions:
> 
> * address `/packages/api-client-next.html` -> can we name it without `next`?
✅ 
> * you needed to rename the file `createApiClient` but missed the admin file, so the source is a dead link
✅ - because the url points to the `main` branch and it's not been merged yet
> 
> Every header should be a link (and is but it's very hard to click it, you need to find dark pixels it on the side), let's fix it to easily link to specific composables: image
> 
✅ done via css workaround

> Additionally - there are a lot of warnings in the console: image

☑️ ✅ I was able to fix only the one for "undefined" link. the missing chevron prop can be fixed only in the source base theme

kstala (approved):

PR #1071, opened 2024-06-28
feat: repeat order

### LINKED ISSUE #1068: [FEATURE] Repeat order
### Description

Add the possibility of repeating the order 

### Use Case

.

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

mdanilowicz:
### Description

Screenshot 2024-06-28 at 13 07 36

- Added `Repeat order` possibility
- Added `addProducts` function that allows add set of product to the cart

 closes #1068


packages/composables/src/useCart.ts
```diff
@@ -131,6 +137,22 @@ export function useCartFunction(): UseCartReturn {
     return addToCartResult;
   }
 
+  async function addProducts(
```
patzick, line 140: Could we maybe refactor `addProduct` to use this new function inside so we have consistency and less code?

.changeset/tidy-comics-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-pwa/composables-next": minor
+---
+
+Added `addProducts` function that allows add set of product to the cart
```
patzick, line 5: It's not clear in which composable this function is added. My suggestion
```suggestion
`useCart` - Added `addProducts` function that allows adding a set of products to the cart
```

packages/composables/vitest.config.ts
```diff
@@ -8,10 +8,10 @@ export default defineConfig({
       enabled: true,
       // "100": true, // TODO: our goal ;)
       thresholds: {
-        statements: 94,
+        statements: 95,
         branches: 80,
-        functions: 82,
-        lines: 92,
+        functions: 84,
+        lines: 95,
```
patzick, line 14: closer and closer ;)

packages/composables/src/useCart.test.ts
```diff
@@ -0,0 +1,66 @@
+import { describe, expect, it } from "vitest";
+import { useCart } from "./useCart";
+import { useSetup } from "./_test";
+import type { operations } from "#shopware";
+
+describe("useCart", () => {
+  it("adding products", async () => {
```
patzick, line 7: can we make it more per unit and test `addProduct` and `addProducts` separately?

patzick (changes requested):
good feature, few minor comments from me

packages/composables/src/useCart.test.ts
mdanilowicz: done

packages/composables/src/useCart.ts
mdanilowicz: This would be a breaking change. I changed `addProduct` to be deprecated.

patzick: that would not be a breaking change, as the API for `addProduct` would not change, only internally it would call `addProducts` inside

kstala (approved):

patzick (approved):

PR #1074, opened 2024-07-01
fix(composables): load shipping address association

mkucmus:
### Description

- vue-demo-store: Fixes missing order status and missing shipping address at the order success page
- api-gen: takes care of translated properties and their format in output .d.ts files
- all: removes optional chaining for translated object and its properties

### Type of change


packages/composables/src/useOrderDetails.ts
```diff
@@ -192,7 +193,9 @@ export function useOrderDetails(
   const shippingCosts = computed(() => _sharedOrder.value?.shippingTotal);
   const subtotal = computed(() => _sharedOrder.value?.price?.positionPrice);
   const total = computed(() => _sharedOrder.value?.price?.totalPrice);
-  const status = computed(() => _sharedOrder.value?.stateMachineState?.name);
+  const status = computed(
```
mdanilowicz, line 196: Translations are not part of the business logic. We should use "core" values in the composables. A better solution might be to return the whole status object, where the translated value can be read on the template. What do you think?

mdanilowicz (changes requested):

packages/composables/src/useOrderDetails.ts
mkucmus: cool, let's solve that here, because I don't see any problem of using built-in property in the composable as well as our composable function do rely on the context, including a language.

I'm not entirely convinced yet, because in this particular case, we have translated property value that comes directly from API according to session context language - which is related to - so it is a part of the core logic.
in the backend it's called `TranslatedField` class and the values come for the specific snippet built-it. 

For custom/manual translation in the frontend project I would use - the `statusTechnicalName` + vue-demo-store snippet translation because it's probably out of language context and can be used as a translation key.

kstala (approved):

mkucmus:
@kstala if you don't mind, please retest this PR, but now check also the translated strings coming from API, like names, descriptions, breadcrumbs, etc..

.changeset/afraid-keys-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-pwa/helpers-next": patch
+---
+
+Removed optional chaining for translated properties
```
patzick, line 5: what is exact change in API of helpers, needs to be described here

.changeset/early-trains-run.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-pwa/composables-next": patch
+---
+
+Load shipping address for the order details
```
patzick, line 5: what is exact change in API of composables, something like:
`useSomething` - in `loadOrders` function we're also fetching shipping address

.changeset/funny-onions-call.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware/api-gen": patch
+---
+
+Set translated proeprty as required in provided schemas
```
patzick, line 5:
```suggestion
Set `translated` property as required in provided schemas
```

.changeset/sour-starfishes-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware/api-client": patch
+---
+
+Update schemas' translated properties as required
```
patzick, line 5:
```suggestion
Update default schemas' `translated` properties as required
```

packages/api-client/api-types/storeApiTypes.d.ts
```diff
@@ -1020,14 +1014,13 @@ export type Schemas = {
...
-      previewMediaId?: string;
-      type?: string;
-      versionId?: string;
+    translated: {
+      cssClass: string;
+      entity: string;
+      name: string;
+      previewMediaId: string;
+      type: string;
+      versionId: string;
```
patzick, line 1023: shouldn't this also be in exceptions?

packages/api-gen/src/commands/generate.ts
```diff
@@ -124,11 +124,14 @@ export async function generate(args: {
             schemaObject.type === "object" &&
             !!schemaObject?.properties?.translated
           ) {
+            schemaObject.required?.push("translated");
```
patzick, line 127: this line probably should be only when we're adding `translated` property, so around line 160

patzick (changes requested):
great work, just few small things to polish it 💅

mkucmus:
since now on, we access for translated and not translated values for properties using `translated` object, which also contain a fallback value if there is no translation.

.changeset/early-trains-run.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-pwa/composables-next": patch
+---
+
+Load shipping address for the order details. Ivoking a `loadOrderDetails` method now will fetch also a `shippingOrderAddress` association.
```
patzick, line 5:
```suggestion
`useOrderDetails` - Load shipping address for the order details. Ivoking a `loadOrderDetails` method now will fetch also a `shippingOrderAddress` association.
```

.changeset/afraid-keys-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-pwa/helpers-next": patch
+---
+
+Removed optional chaining for translated properties. Expecting a different argument's type for `getLanguageName` and `getShippingMethodDeliveryTime` methods.
```
patzick, line 5:
```suggestion
Removed optional chaining for translated properties. Expecting a different argument type for the `getLanguageName` and `getShippingMethodDeliveryTime` methods.
```

patzick (approved):

PR #1077, opened 2024-07-02
Fix for SeoUrls in the useListing.ts

cfreak:
### Description

Fix for SeoUrls in the useListing.ts because the sw-include-seo-urls header is missing.

### Type of change


patzick:
Tests are failing as it is not fixing the schema as well, I prepared fix in #1078 addressing your issue

PR #1082, opened 2024-07-02
fix(api-gen): better validation rule for COMPONENTS_API_ALIAS flag and proper overrides loading

patzick:
### Description

- fixes problem with overrides loading
- improves validation message suggesting that schema component name might be incorrect
- added more tests for the rule
image


mkucmus (approved):
image

BrocksiNet (approved):

PR #1086, opened 2024-07-03
fix(docs): wrong links in pageRef

BrocksiNet:
### Description
We can avoid some 301 redirects when using the correct links in the sidebar.

Also, tools like [deadlink-checker](https://www.deadlinkchecker.com/website-dead-link-checker.asp) are complaining about this.

### Type of change

image


BrocksiNet:
@patzick, now I am done here.

That's the result now:
image

FYI @kstala

patzick (approved):

PR #1100, opened 2024-07-10
feat: use import.meta.client

BrocksiNet:
### Description

`process.client` is deprecated we should now use `import.meta.client`.

### Type of change


BrocksiNet:
In `packages/nuxt3-module/plugin.ts`, we also use `process.server` separated PR, or can this be done here? :thinking:

patzick (approved):

PR #1103, opened 2024-07-12
fix(api-gen): re-enabled GenericRecords for objects in api-type schema

trimethylpentan:
### Description

In pull request #1032 the transformation of generic openAPI objects was commented out. In result, all generic objects (e.g. associations) have been generated as `Record<string, never>`, making them unusable. This PR reverts that behaviour.

### Type of change

### TODOs

- [X] Changeset file provided [read more](https://github.com/shopware/frontends/blob/main/CONTRIBUTION.md#changelog-preparation)
- [ ] Documentation added/updated
- [ ] E2E-Tests added/updated
- [ ] Related Issue updated


patzick (changes requested):
Hey @trimethylpentan 
Thank you for your contribution. However, we're aiming in the direction of improving schema, not making it generic. Setting GenericRecord is like `any`.
Please look at the newest schema, associations are already typed: https://github.com/shopware/frontends/blob/main/packages/api-client/api-types/storeApiTypes.d.ts#L1276 and with every version we'll be adding better typings

trimethylpentan:
Hello @patzick,
thanks for your quick response. Unfortunately, this is currently not working for the admin api schema. Associations there are typed as `Record<string, never>`, as you can see here: https://raw.githubusercontent.com/shopware/frontends/main/packages/api-client/api-types/adminApiTypes.d.ts

This means associations are currently not usable without overwriting their types.

patzick (approved):
I just regenerated schemas to match code changes. We're good to go 🙌

PR #1107, opened 2024-07-15
refactor: upgrade openapi-typescript to 7.x

patzick:
### Description

Upgrade to openapi-typescript 7.x. Requires still updates or fixes on lib part. Leaving as a draft for now


packages/api-client/api-types/storeApiTypes.d.ts
```diff
@@ -302,41 +286,41 @@ export type Schemas = {
   };
   ArrayStruct: components["schemas"]["Struct"];
   Association: {
-    [key: string]: components["schemas"]["Association"];
+    [key: string]: components["schemas"]["Association"] | undefined;
```
patzick, line 289: This is the main change that still should be resolved. We don't want to `undefined` values here if the schema correctly suggest types which are here

patzick:
continued in #1405

PR #1111, opened 2024-07-16
fix(api-client): use correct components name

BrocksiNet:
### Description

This will solve the renaming problem of the component in the core.


BrocksiNet:
For release we should have also a increased version number :smile:

BrocksiNet:
See https://github.com/shopware/frontends/pull/1112

PR #1131, opened 2024-07-18
feat(composables): add 100% coverage

mdanilowicz:


packages/composables/src/useProductPrice.ts
```diff
@@ -71,37 +71,49 @@ export function useProductPrice(
...
-  const displayFrom: ComputedRef<boolean> = computed(
-    () =>
-      (product.value?.calculatedPrices?.length ?? 0) > 1 ||
-      !!(_displayParent.value && displayFromVariants.value),
-  );
+  const displayFrom: ComputedRef<boolean> = computed(() => {
+    // Change me mind:
+    //  `!!(_displayParent.value && displayFromVariants.value)` will never return true
+    // - _displayParent.value expects `product.value?.parentId === null`
+    // - displayFromVariants.value expects `!!product.value?.parentId`
```
patzick, line 84: yes, looks like a dead case, let's remove it

packages/composables/src/useProductPrice.ts
```diff
@@ -71,37 +71,49 @@ export function useProductPrice(
...
-        );
-        return (
-          lowest || (_cheapest.value as unknown as Schemas["CalculatedPrice"])
-        );
+        return product.value?.calculatedPrices?.reduce((previous, current) => {
+          return current.unitPrice < previous.unitPrice ? current : previous;
+        });
+
+        // Change my mind:
+        // There is no possibility to return _cheapest.value because of the getProductTierPrices that is iterating over calculatedPrices
```
patzick, line 113: If somehow in the `calculatedPrices` the `unitPrice` would be `0` then in theory it is possible to return `_cheapest`. If that is the case anywhere? I'm not sure. I'd leave this logic (without casting to unknown and CalculatedPrice) and create a follow-up issue for the discovery of that case

mkucmus: now, when the types reflect the actual shape of the data from API we can trust it, so removing redundant code can be done. I've checked the preview and compared the variants, from prices and so on. The prices are displayed correctly - in comparison to `main` branch which reflects the storefront behavior.

packages/composables/src/useProduct.ts
```diff
@@ -31,10 +30,7 @@ export function useProduct(
   configurator?: Ref<Schemas["PropertyGroup"][]> | Schemas["PropertyGroup"][],
 ): UseProductReturn {
   const _product = useContext("product", { context: product });
-  if (!_product.value) {
-    // TODO link docs with composables context usage
-    throw new ContextError("Product");
-  }
```
patzick, line 37: why is this removed?
If product is not provided with param or context it would cause runtime errors

packages/composables/src/useProductAssociations.spec.ts
```diff
@@ -88,4 +90,18 @@ describe("useProductAssociations", () => {
 
     expect(vm.productAssociations).toStrictEqual(mockedCrossSelling);
   });
+
+  it("init without product", () => {
```
patzick, line 94:
```suggestion
  it("init without product should throw an error", () => {
```

packages/composables/src/useProductAssociations.spec.ts
```diff
@@ -88,4 +90,18 @@ describe("useProductAssociations", () => {
...
+    expect(() =>
+      useSetup(() =>
+        useProductAssociations(
+          // Api case
+          ref(null) as unknown as ComputedRef<Schemas["Product"]>,
```
patzick, line 99:
```suggestion
          // @ts-expect-error if API returns null we want to be prepared for it
          ref(null),
```

packages/composables/src/useProductPrice.test.ts
```diff
@@ -63,4 +62,65 @@ describe("useProductPrice", () => {
...
+            },
+          ) as unknown as Schemas["Product"],
+        ),
+      ),
+    );
+    expect(vm.displayFrom).toBe(true);
+    expect(vm.isListPrice).toBe(false);
+  });
+
+  it("isListPrice - 3", () => {
```
patzick, line 91: 2345678765456
better documented please :)

packages/composables/vitest.config.ts
```diff
@@ -7,14 +7,20 @@ export default defineConfig({
...
-      // "100": true, // TODO: our goal ;)
       thresholds: {
-        statements: 97,
-        branches: 90,
-        functions: 88.4,
-        lines: 97,
+        statements: 100,
+        branches: 100,
+        functions: 100,
+        lines: 100,
```
patzick, line 14: we can replace it with that
```suggestion
        "100": true,
```

packages/composables/vitest.config.ts
```diff
@@ -7,14 +7,20 @@ export default defineConfig({
...
       },
-      exclude: ["**/devtools/**"],
+      exclude: [
+        "**/devtools/**",
+        "**/types/**",
+        "**/src/*.test.ts",
+        "**/src/*.spec.ts",
+        // After refactoring should be removed
```
patzick, line 21: link to GH issue so it's known when resolved when someone will see this TODO in future

packages/composables/src/_test/useSetup.ts
```diff
@@ -8,6 +8,7 @@ type Injections = {
   apiClient: {
     invoke: ReturnType<typeof vi.fn>;
   };
+  [key: string]: unknown;
```
patzick, line 11: why do we need this generic injections? Would be better to know exactly what we're dealing with

patzick (changes requested):
Great work 🙌  I reviewed the code and left some comments. Mainly to make test names more descriptive. People should not need to read the test code tested code to understand what the test is doing. This way, we also have proper documentation for the tested code.

While applying suggestions, please make sure that:
- test names are properly documenting features/behaviours of the code
- mocks are cleared between tests, so there is no impact between singular tests
- fill the PR description appropriately so when someone looks at this PR in the future description summarizes what's changed here

packages/composables/src/useNotifications.ts
```diff
@@ -93,7 +93,7 @@ export function useNotifications(): UseNotificationsReturn {
     const messageId = geterateId();
     _notifications.value.push({
       id: messageId,
-      type: options.type || "info",
+      type: options.type as NotificationType,
```
patzick, line 96: this casting is incorrect. `type` is optional so we had `info` as default type, now it would send `undefined as NotificationType`

patzick: okay, so then just typing needs to change so we won't need to do casting, so this internal method needs to require `options.type`

mdanilowicz: done

packages/composables/src/useOrderPayment.test.ts
```diff
@@ -32,5 +32,49 @@ describe("useOrderPayment", () => {
...
+      expect.objectContaining({
+        body: {
+          orderId: "018c6832f1a1731db276839b3f251fc2",
+          paymentMethodId: "test",
+        },
+      }),
+    );
+  });
+
+  it("empty order", async () => {
```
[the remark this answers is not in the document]
mdanilowicz: done

packages/composables/src/usePrice.test.ts
```diff
@@ -1,12 +1,50 @@
...
+const currentNavigator = global.navigator;
+
+vi.mock("./useSessionContext.ts");
+const sessionContext = ref();
+
+vi.mocked(useSessionContext).mockReturnValue({
+  sessionContext,
+} as unknown as ReturnType<typeof useSessionContext>);
 
 describe("usePrice", () => {
```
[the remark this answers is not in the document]
mdanilowicz: done

patzick (approved):

PR #1132, opened 2024-07-18
chore: cleanup dependencies

patzick:
### Description

cleanup of some unused dependencies and old scripts for doc types generation. This should decrease node_modules size


mkucmus (approved):
code cleaner IRL
image

PR #1155, opened 2024-07-24
feat: add `paymentDetails` within handle-payment operation

mkucmus:
### Description

solves ` Support Ticket (#SWAG-278302)`

### DIFF
--- a/.changeset/eleven-bottles-tickle.md
+++ b/.changeset/eleven-bottles-tickle.md
@@ -1,0 +1,19 @@
+---
+"@shopware/api-client": patch
+---
+
+Update `handlePaymentMethod` operation to have `paymentDetails` parameter available:
+
+```ts
+apiClient.invoke(
+  "handlePaymentMethod post /handle-payment",
+  {
+    body: {
+      orderId,
+      errorUrl,
+      finishUrl,
+      paymentDetails, // NEW PARAMETER AVAILABLE
+    },
+  },
+);
+```
--- a/.changeset/honest-phones-bathe.md
+++ b/.changeset/honest-phones-bathe.md
@@ -1,0 +1,5 @@
+---
+"vue-demo-store": patch
+---
+
+Update api-types for additional `paymentDetails` param within `handlePayment` operation.
--- a/.changeset/twelve-bottles-clap.md
+++ b/.changeset/twelve-bottles-clap.md
@@ -1,0 +1,18 @@
+---
+"@shopware-pwa/composables-next": patch
+---
+
+Update `useOrderPayment` composable's `handlePayment` method to accept also additional payment details as a last argument:
+
+```ts
+const { handlePayment } = useOrderPayment(/** args ommited */);
+
+await handlePayment(
+  "http://localhost:3000/success",
+  "http://localhost:3000/failure?payment-failed",
+  {
+    stateData // pass extra parameter fro api payment processor
+  },
+);
+
+```
--- a/examples/adyen-dropin-component/api-types/storeApiTypes.d.ts
+++ b/examples/adyen-dropin-component/api-types/storeApiTypes.d.ts
@@ -7799,6 +7799,9 @@ export type operations = {
       finishUrl?: string;
       /** Identifier of an order */
       orderId: string;
+      paymentDetails?: {
+        [key: string]: unknown;
+      };
     };
     response: {
       redirectUrl: string;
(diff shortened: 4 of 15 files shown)


mkucmus:
I've created this PR to show how the additional parameter can be added in the api client and the schema itself. 
However, the core API does not support any additional parameter in the payment handler (processor) - [see the code](https://github.com/shopware/shopware/blob/6204df836c4ebee18fc76192c072da2ab6318390/src/Core/Checkout/Payment/SalesChannel/HandlePaymentMethodRoute.php). That's why merging the PR is not recommended. 

To solve that problem in the end-project there are required steps:
- ensure that openapi schema of a payment extensions alters the default definition (added additional parameter to handle payment) 
- sync your schema using `api-gen` package
- there is also a possibility to ignore the backend part, and edit only the project's schema by changing `storeApiTypes.overrides.ts` locally. then regenerated types will contain the newly added params ready to use in your project.

golliholzland:
Hi @mkucmus,
Is this fix related to a feature shopware release ? or is it blocked because of a feature showpare release ?
could you please add the shopware version,you're using.

BrocksiNet:
@golliholzland, from my understanding, @mkucmus is saying that we can not merge the changes because they are not supported in the core payment handler. But depending on what kind of plugin or payment provider you are using, there could be this new parameter. To avoid TypeScript errors, you can extend the TypeDefinition, as shown in the changed files of this MR. Currently, he is on vacation, so his feedback will be delayed.

golliholzland:
@BrocksiNet, the code referenced by @mkucmus comes with shopware version: v6.7.0 (not released yet). that was the reason behind my question.

BrocksiNet:
Sure, I got it. But also in older versions, only parameters like `orderId`, `finishUrl`, and `errorUrl` were used.

mkucmus:
> Sure, I got it. But also in older versions, only parameters like `orderId`, `finishUrl`, and `errorUrl` were used.
> 
> So, I do not see the `paymentDetails` parameter in the core code. It is something added by plugins/apps?

Apparently it's only reason why it was present in the schema before - because it was added by some extension or a plugin. Currently the API client contains only the core-ish stuff so if you need such parameter you would need to override the schema manually or provide a plugin containing a proper openapi schema defined.

I created the PR that exaplains how to add this kind of things when your plugin does not support OpenAPI Schema and you have to take care of it by your own. Of course there is no need to do it like in this PR - I added `paymentDetails` globally, and it was added to the related project by a special command.  

To override a schema definition in your project check this: https://github.com/shopware/frontends/discussions/471 (`Overriding types` chapter)

veliu:
@BrocksiNet @mkucmus I'm a little bit confused, maybe you can help me out. 

Shopware documented how to transmit additional payment information for handle-payment endpoint [here](https://shopware.stoplight.io/docs/store-api/8218801e50fe5-handling-the-payment#transmit-additional-payment-details).

When you check the code base the parameter `paymentDetails` is not handled or validated at all, BUT [the whole request data bag](https://github.com/shopware/shopware/blob/6204df836c4ebee18fc76192c072da2ab6318390/src/Core/Checkout/Payment/SalesChannel/HandlePaymentMethodRoute.php#L49) is passed down to the handlers and could be accessed accordingly. 

Did I miss something?

mkucmus:
> @BrocksiNet @mkucmus I'm a little bit confused, maybe you can help me out.
> 
> Shopware documented how to transmit additional payment information for handle-payment endpoint [here](https://shopware.stoplight.io/docs/store-api/8218801e50fe5-handling-the-payment#transmit-additional-payment-details).
> 
> When you check the code base the parameter `paymentDetails` is not handled or validated at all, BUT [the whole request data bag](https://github.com/shopware/shopware/blob/6204df836c4ebee18fc76192c072da2ab6318390/src/Core/Checkout/Payment/SalesChannel/HandlePaymentMethodRoute.php#L49) is passed down to the handlers and could be accessed accordingly.
> 
> Did I miss something?

hey! thanks for pointing the actual core's code to see what's inside 👍🏼
the problem is not that the core does or doesn't handle the additional parameters. The actual issue is that **in order to keep the types safe, we don't allow to leave anything as `any`.** to trust the code, we need to ensure it reflects the actual state, without any doubts or guesses. If anything can be passed down in the request's payload, it needs to be documented in open api schema. thanks to this, the client will be aware of that without getting any static errors or warning. 

If you are sure that there are some additional parameters allowed by some plugin - please contain the information in openapi schema (like overriding point from this [doscussion](https://github.com/shopware/frontends/discussions/471)) and everything will be fine, and you will able to use it 🙌🏼

PR #1172, opened 2024-07-29
feat(examples): Using multiple sales channels

freb97:
### Description

Adds an example for how to use the existing composables and helpers with multiple sales channels.
The shopware plugin (`/plugins/shopware.ts`) can be used as an optional replacement for the nuxt3-module in the vue-demo-store.

### ToDo's

[&nbsp;&nbsp;&nbsp;&nbsp;] Update current example configuration in nuxt.config.ts with different sales channel access keys (optional).
[✅] Documentation added/updated

Screenshot 2024-07-29 at 15 14 23

### Additional context

The current configuration in the nuxt.config.ts uses the same sales channel three times. It would be possible to add multiple sales channels here if the current demo store is configurated accordingly.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -158,13 +158,13 @@ importers:
         version: link:../../packages/api-gen
       '@unocss/nuxt':
         specifier: 0.61.9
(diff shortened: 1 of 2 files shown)


BrocksiNet:
Why was this closed? :cry:

BrocksiNet:
New PR can be found [here](https://github.com/shopware/frontends/pull/1172).

PR #1184, opened 2024-08-01
feat(composables): Export default order associations

PheysX:
### Description

This PR exports the default order associations.

I tried to extend the useOrderDetails function and I need the defaults. I don't want to copy it.


patzick (commented):
we'd need to indicate which associations they are. It also shouldn't be considered as "public API" as many people then would like to change the associations and provide "fixes" for their usecases.

I'd propose a different option - create `useDefaultOrderAssociations` composable, which returns this object, this way it could be overridden or reused across projects. Would that suit your needs?

patzick:
closing in favour of #1194 
thank you for raising this issue 🙌

PR #1187, opened 2024-08-01
docs: add warning info for pages generated from md files

mkucmus:
### Description

Adds an information about source of the content on the page, links the source of it in case of any issues.

works only if `<!-- load: ... -->` snippet is being used.

image


BrocksiNet:
Can we make it more descent? Currently it looks like a urgent warning :joy:

patzick:
yeah +1 for "less disturbing", I'd also move this to the bottom as it's not the first information someone wants to read on the documentation page ;P

mkucmus:
how about that:
image

PR #1191, opened 2024-08-05
fix(helpers): correct srcset for images

mkucmus:
### Description

wrong `srcset` attribute causes some problems during loading an image. because the wrong format, like `400w` without a corresponding image, the image was being loaded through a nuxt router as a normal page (in an entrypoint `[...all].vue`).

### ToDo's

image
### Additional context

image


kstala (approved):

patzick (approved):
great, love detailed changeset 💙

PR #1194, opened 2024-08-05
feat(composables): useDefaultOrderAssociations for accessing associations

patzick:
### Description

another approach of #1184 to allow access and override default order associations


mkucmus (approved):
image

docs updated ✅ 
js docs added ✅

PR #1218, opened 2024-08-14
fix(vue-demo): modal for resetting password page

### LINKED ISSUE #1214: [BUG] Incorrect behavior on sign in modal

### Current Behavior

If we select the "Forgot you password" option from the sign im modal, we can see that we are redirected to the correct page, but the modal is still displayed and can only be closed by clicking outside. I think a better solution would be to close it after redirection.

https://github.com/user-attachments/assets/c12b2082-a9b2-46c8-904e-898cc03f5f54

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
- Node:
- pnpm:
```

### Anything else?

_No response_

mdanilowicz:
### Description

- Fixed closing modal for resetting password page

closes #1214


templates/vue-demo-store/components/account/AccountMenu.vue
```diff
@@ -39,12 +45,11 @@ async function invokeLogout() {
           >
             <div class="flex items-center justify-end">
               <div class="text-sm">
-                <NuxtLink
-                  :to="formatLink(`/account/recover`)"
-                  class="font-medium text-indigo-600 hover:text-indigo-500"
+                <span
```
patzick, line 48: this changes routes, so should stay link

if click event is really needed than it should be a button for accessibility

maybe some router hook here to close it on navigation?

patzick: but you also can attach additional `@click` for link and that would solve both problems

mdanilowicz: done

patzick (approved):

PR #1246, opened 2024-08-28
Feat/composables doc

mdanilowicz:


mkucmus (changes requested):
now it's way better organized structure 🙌🏼 

from the things missing:
- no direct links to the source of the composable to gh repository on main branch to the exact line - was useful and please do not get rid of this:
image

- I would increase the headers of categories +1 to be available in the context menu:
category-links

patzick (approved):
Looks good, @mkucmus please review as well

mkucmus (approved):
image

PR #1247, opened 2024-08-28
feat(examples): Using multiple sales channels

patzick:
### Description
pushed changes from #1227


patzick:
the audit check is not related with that PR, we can skip this here

Issue #641, opened 2024-02-20
[REFACTOR] Migrate provide / inject syntax to reusable composables


maartenvanhunsel:
### Description

Numerous composables we utilize are closely integrated with each other. Nevertheless, the current integration approach renders several key Nuxt/Vue features unusable, which is regrettable. Specifically, Plugins and Middlewares become inaccessible for Shopware-related composables due to the necessity of utilizing `useContext` with `inject()`, a feature restricted to the setup function.

As a developer, it is crucial to have the user context at the earliest stage possible. By fetching data from the `app.vue`, we inadvertently overlook numerous valuable functionalities that could enhance the user experience.

### Use Case

Authentication

### Proposed Solution

I recommend adopting a more composable approach such as `useAuth` from [@sidebase/nuxt-auth](https://sidebase.io/nuxt-auth/v0.6/application-side/session-access-and-management#useauth-composable). This solution is remarkably user-friendly, eliminating the requirement for provide and inject.

### Alternatives Considered

I have reviewed https://github.com/shopware/frontends/issues/26, and I firmly believe that this approach is not sustainable. It risks misrepresenting the core concepts that should guide our efforts.

### Additional Context

I am happy to assist and engage in discussions for a sustainable solution that allows us to leverage the best of both (Shopware / Nuxt) worlds.

maartenvanhunsel:
@elkmod see refactor demo as we discussed earlie.

- Stackblitz (https://stackblitz.com/~/github.com/maartenvanhunsel/shopware-nuxt-3-refactor)
- Github (https://github.com/maartenvanhunsel/shopware-nuxt-3-refactor)

ignasradzius:
I like your idea as I ran into limitation of using useUser composable in the middleware as well.

PR #1257, opened 2024-08-30
poc: session context from nuxt plugin

mkucmus:
## How it works
PoC Exposes a current session in computed property in a nuxt plugin level (client&server) `const { $getCurrentSession } = useNuxtApp()`

it's not a final state, only a proposition that may solve an issue like described [here](https://github.com/shopware/frontends/issues/641).

This little change move logic responsible for fetching a context to the lower level of a nuxt app (from app.vue to nuxt plugin). thanks to this, session context can be read in a nuxt middleware, for instance. 

## Why?
The current session is the data frequently needed from the very beginning state of a running application, when nuxt context is not set, vue lifecycle hooks are not available.

- allows us to use the current session according to the `sw-context-token` provided in the cookie (remains stateless - only cookie is used, no other storage is needed)
- it saves requests, in comparison to using the plain `apiClient` instance in the middleware
- does not require changes in the other composables API
- does not make a singleton of any reference - bases only on a plugin

## What does it contain?

- moved logic of fetching a context from `app.vue` to nuxt3-module's plugin
- in SSR it fetches the context only once, then pass it via ssrContext.payload (similar way of how it's managed when using `useAsyncData`)
- the session is being read on the client side (also in a plugin) if it had been set in SSR, otherwise it's being loaded using api client
- fixes unnecessary language patching

## TEST IT:
- run app and go to `/shopware` route to see the middleware working https://frontends-demo-git-poc-session-contex-26fff6-shopware-frontends.vercel.app/shopware
- log in 
- reload the page
- log out
- etc...

## Next steps
- Could be also a good solution for the another layer of composables - nuxt composables when `useNuxtApp` is available - then use this approach, when it's a pure vue, use composition api for vue context only.
- check if works in CSR mode only
- check possibly memory leaks


templates/vue-demo-store/middleware/auth.ts
```diff
@@ -0,0 +1,9 @@
+export default defineNuxtRouteMiddleware((context) => {
+  const { $getCurrentSession, ssrContext } = useNuxtApp();
+
+  console.warn(
+    ssrContext ? "(server)" : "(client)",
+    "customer email:",
+    $getCurrentSession.value.customer?.email,
```
BrocksiNet, line 7: Hmm, maybe we just show first name and surname and not email.

templates/vue-demo-store/pages/shopware.vue
```diff
@@ -26,6 +32,9 @@ useBreadcrumbs([
 
 <template>
   <LayoutBreadcrumbs />
+  <div>
+    <pre>customer: {{ customer }}</pre>
```
BrocksiNet, line 36: Hidden in our easter egg :joy:

Issue #1263, opened 2024-09-03
[BUG] [error.vue] Cannot add property data-v-inspector, object is not extensible for VUE-DEMO-STORE


kstala:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Vue-demo-store is currently not working on stackblitz https://stackblitz.com/github/shopware/frontends/tree/main/templates/vue-demo-store?file=README.md

` ERROR  [error.vue]: Cannot add property data-v-inspector, object is not extensible
`

### Environment

```markdown
- OS:
- Node:
- pnpm:
```

cfreak:
Some here... does anyone know a solution?

mkucmus:
> Some here... does anyone know a solution?

temporarily I recommend disable `componentInspector` option for devtools:

```ts
// nuxt.config.ts
export default defineNuxtConfig({
  devtools: {
    componentInspector: false, // <-- this one
  },
...
 ```

I found that the problem comes from `CmsElementText.vue` component of `cms-base` package, precisely there might be a bug within `renderHtml` helper in combination of extra attributes taken from devtools when `data-v-inspector` is being added.

PR #1276, opened 2024-09-10
fix(cms-base): render function for CmsElementText

mkucmus:
### Description

closes: #1263 

to not directly render the child vnodes 

working preview of the same solution: https://stackblitz.com/edit/github-94vefp?file=error.vue

### DIFF
--- a/packages/cms-base/components/public/cms/element/CmsElementText.vue
+++ b/packages/cms-base/components/public/cms/element/CmsElementText.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import type { CmsElementText } from "@shopware-pwa/composables-next";
 import { useCmsElementConfig, useUrlResolver } from "#imports";
-import { computed, getCurrentInstance, h } from "vue";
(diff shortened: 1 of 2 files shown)


packages/cms-base/components/public/cms/element/CmsElementText.vue
```diff
@@ -25,120 +25,124 @@ const style = computed<CSSProperties>(() => ({
 
 const hasVerticalAlignment = computed(() => !!style.value.alignItems);
 
-const CmsTextRender = () => {
-  const { resolveUrl } = useUrlResolver();
+const CmsTextRender = defineComponent({
```
mkucmus, line 28: I've could also left it as it was to not use `defineComponent` and return wrapped nodes in `h` function. 

```ts
return () => h("div", {}, renderHtml(rawHtml, config, h, context, resolveUrl));
```

kstala (approved):

patzick (approved):

PR #1028, opened 2024-06-19
fix: product search

mdanilowicz:
### Description

Removed duplicated search requests

closes #1012

PR #1281, opened 2024-09-13
fix: listing and search quick wins

BrocksiNet:
### Description

This PR fixes two things:
- Bring back the count on search page (without duplicated request)
- Let the price filter set the selected value when there is some (CMS-base package fix)

### Type of change

### Additional context
There is an [old PR](https://github.com/shopware/frontends/pull/1028/files) that broke the count on search.


packages/cms-base/components/listing-filters/SwFilterPrice.vue
```diff
@@ -43,8 +45,12 @@ const prices = reactive<{ min: number; max: number }>({
 });
 
 onMounted(() => {
-  prices.min = Math.floor(props.filter?.min || 0);
-  prices.max = Math.floor(props.filter?.max || 0);
+  prices.min = Math.floor(
+    props.selectedFilters.price.min || props.filter?.min || 0,
```
mdanilowicz, line 49: We should use `??` operator because 0 is a valid value

0 ?? 10 -> 0
0 || 10 -> 10

patzick (approved):

packages/cms-base/components/listing-filters/SwFilterPrice.vue
patzick: agree, changed that

PR #1311, opened 2024-09-24
fix: add missing ssr endpoint

mdanilowicz:

### DIFF
--- a/apps/docs/src/getting-started/templates/blank-template.md
+++ b/apps/docs/src/getting-started/templates/blank-template.md
@@ -25,16 +25,18 @@ The blank template is pre-configured to connect to a public Shopware backend, so
 
 In order to connect it to your own store backend, you need to edit the `nuxt.config.ts` file and add a configuration details:
 
-```ts{4-7}
+```ts
 /* ... */
 export default defineNuxtConfig({
-  /* ... */
-  shopware: {
-    endpoint: "https://your-business.shopware.store",
-    accessToken: "access-token-from-settings",
-  },
   // or
   runtimeConfig: {
+    // shopware: {
+    /**
+     * SSR Shopware Endpoint
+     * More here: https://frontends.shopware.com/getting-started/templates/custom-vue-project.html#shopware-endpoint-on-the-ssr-mode
+     */
+    //   endpoint: ""
+    // },
     public: {
       shopware: {
         endpoint: "https://your-business.shopware.store",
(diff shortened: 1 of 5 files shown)


apps/docs/src/getting-started/templates/blank-template.md
```diff
@@ -25,16 +25,18 @@ The blank template is pre-configured to connect to a public Shopware backend, so
...
-```ts{4-7}
+```ts
 /* ... */
 export default defineNuxtConfig({
-  /* ... */
-  shopware: {
-    endpoint: "https://your-business.shopware.store",
-    accessToken: "access-token-from-settings",
-  },
   // or
```
mkucmus, line 31: remove this comment as well it if you don't propose the way without runtimeConfig

apps/docs/src/getting-started/templates/demo-store-template.md
```diff
@@ -73,20 +73,22 @@ In order to connect it to your own store backend, you need to edit the `nuxt.con
 ```ts
 /* ... */
-  /* ... */
-  shopware: {
-    endpoint: "https://your-business.shopware.store",
-    accessToken: "access-token-from-settings",
-  }
   // or
```
mkucmus, line 76: no `or` anymore

packages/nuxt3-module/plugin.ts
```diff
@@ -26,12 +26,21 @@ declare module "vue" {
 
 export default defineNuxtPlugin((NuxtApp) => {
   const runtimeConfig = useRuntimeConfig();
-  const shopwareEndpoint =
+  const shopwareEndpointCSR =
     runtimeConfig.public?.shopware?.endpoint ??
     runtimeConfig.public?.shopware?.shopwareEndpoint;
+
+  const shopwareEndpointSSR =
+    runtimeConfig.shopware?.endpoint ?? shopwareEndpointCSR;
```
mkucmus, line 34: maybe we could add some warning in the build type in `src/index.ts` that in SSR the endpoint defined for CSR will be used as a fallback - or other way around.

mkucmus (changes requested):
two comments to remove and one thing to consider to add

packages/nuxt3-module/plugin.ts
mdanilowicz: I'm not convinced because ...
- This SSR endpoint should be used only for a very specific occasion, so we should avoid logging unnecessary information. The developer can easily debug it
- Plugins are triggered by the request, so the console will be spammed. 
- CRS requests never reach the SSR endpoint

mkucmus: I was only mentioning to log it in the build time in the module, not a plugin - so no spamming here. we should better inform the developer what endpoint/configuration is being used, because some of them will never reach our repository to find the specific line of code - that's just my opinion 🙌🏼

mdanilowicz: Ah got it, done

mkucmus (approved):
image

PR #1317, opened 2024-09-26
fix(templates): show/hide order details conditionally in vue-demo-store

khanSoliheen:


mkucmus (commented):
Hey, thanks @khanSoliheen for providing this PR, works like a charm 💪🏼 

Could you please revert the pnpm-lock.yaml file I accidentally deleted? 
And the second thing - to make this PR done there is no need to update `pnp-lock.yaml` because you've changed only the logic part in the template.

khanSoliheen:
> Hey, thanks @khanSoliheen for providing this PR, works like a charm 💪🏼
> 
> Could you please revert the pnpm-lock.yaml file I accidentally deleted? And the second thing - to make this PR done there is no need to update `pnp-lock.yaml` because you've changed only the logic part in the template.

It is done please review it.

mkucmus (approved):
🥇

PR #1325, opened 2024-10-02
feat(vue-demo): rules flow

### LINKED ISSUE #1320: [BUG] Checkout problems with the rules

- We have to refresh payments and shipping methods after each checkout action. This is related to the rules that can change payment or shipping
- If the selected shipping/payment is no longer available, we should choose the new one (? to confirm)
- We should inform customer that his payment/shipping method is not available anymore 

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
- Node:
- pnpm:
```

### Anything else?

_No response_

mdanilowicz:
### Description

closes #1320

### DIFF
--- a/.changeset/unlucky-birds-pay.md
+++ b/.changeset/unlucky-birds-pay.md
@@ -1,0 +1,6 @@
+---
+"vue-demo-store": minor
+---
+
+- Added `refreshShoppingAndPaymentMethods` method that is used for rules scenarios that are related to the shipping or payment
+- Added check of shipping method selection on the checkout process
--- a/packages/composables/src/useCartErrorParamsResolver/useCartErrorParamsResolver.ts
+++ b/packages/composables/src/useCartErrorParamsResolver/useCartErrorParamsResolver.ts
@@ -2,7 +2,7 @@ import { useCart } from "#imports";
 
 export type UseCartErrorParamsResolver = {
   resolveCartError: (errorObject: ErrorObjectParam) => {
-    params: { [key: string]: string | number | null } | null;
+    params: { [key: string]: string | number | null | undefined } | null;
     messageKey: string;
   };
 };
@@ -48,6 +48,10 @@ export function useCartErrorParamsResolver(): UseCartErrorParamsResolver {
       return cartItems.value.find((item) => item.id === id);
     };
 
+    const buildShippingMethodBlocked = (): string => {
+      return errorObject.message?.replace("shipping-method-blocked-", "") || "";
+    };
+
     switch (errorObject.messageKey) {
       case "product-stock-reached":
         params = buildProductStockReached();
@@ -55,6 +59,11 @@ export function useCartErrorParamsResolver(): UseCartErrorParamsResolver {
           messageKey = "product-stock-reached-empty";
           params = null;
         }
+        break;
+      case "shipping-method-blocked":
+        params = {
+          name: buildShippingMethodBlocked(),
+        };
         break;
       default:
         params = { ...errorObject };
--- a/packages/helpers/src/cms/getCmsTranslate.ts
+++ b/packages/helpers/src/cms/getCmsTranslate.ts
@@ -7,7 +7,7 @@
  */
 export function getCmsTranslate(
   key: string,
-  params?: { [key: string]: string | number | null } | null,
+  params?: { [key: string]: string | number | null | undefined } | null,
 ) {
   if (!params) return key;
 
--- a/templates/vue-demo-store/pages/checkout/index.vue
+++ b/templates/vue-demo-store/pages/checkout/index.vue
@@ -74,6 +74,7 @@ const selectedShippingMethod = computed({
   async set(shippingMethodId: string) {
     isLoading[shippingMethodId] = true;
     await setShippingMethod({ id: shippingMethodId });
+    await refreshShoppingAndPaymentMethods({ skipShipping: true });
     isLoading[shippingMethodId] = false;
   },
 });
@@ -84,6 +85,7 @@ const selectedPaymentMethod = computed({
   async set(paymentMethodId: string) {
     isLoading[paymentMethodId] = true;
     await setPaymentMethod({ id: paymentMethodId });
+    await refreshShoppingAndPaymentMethods({ skipPayment: true });
     isLoading[paymentMethodId] = false;
   },
 });
@@ -95,6 +97,7 @@ const selectedShippingAddress = computed({
   async set(shippingAddressId: string) {
     isLoading[`shipping-${shippingAddressId}`] = true;
     await setActiveShippingAddress({ id: shippingAddressId });
+    await refreshShoppingAndPaymentMethods();
     if (shippingAddressId === selectedBillingAddress.value)
       customShipping.value = false;
     isLoading[`shipping-${shippingAddressId}`] = false;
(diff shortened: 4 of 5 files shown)


templates/vue-demo-store/pages/checkout/index.vue
```diff
@@ -204,12 +209,20 @@ const placeOrder = async () => {
...
 
   isLoading["placeOrder"] = true;
-  const order = await createOrder();
-  isLoading["placeOrder"] = false;
-  await push("/checkout/success/" + order.id);
-  refreshCart();
+
+  try {
+    const order = await createOrder();
+    console.log("order", order);
```
BrocksiNet, line 218: remove the log

templates/vue-demo-store/pages/checkout/index.vue
```diff
@@ -268,6 +281,42 @@ async function invokeLogout() {
...
+      : null,
+  ];
+
+  if (!isVirtualCart.value) isLoading["shippingMethods"] = true;
+  isLoading["paymentMethods"] = true;
+  return Promise.any(promises).finally(() => {
+    isLoading["shippingMethods"] = false;
+    isLoading["paymentMethods"] = false;
+  });
+};
```
BrocksiNet, line 302: Just for easier readability. But not tested (!). So this is just a suggestion to think about readability.

```suggestion
const REFRESH_OPTIONS = {
    FORCE_RELOAD: { forceReload: true },
    SKIP_SHIPPING: 'skipShipping',
    SKIP_PAYMENT: 'skipPayment'
};

const refreshShoppingAndPaymentMethods = (params = {}) => {
    let { skipShipping, skipPayment } = params;

    let loadShipping = !isVirtualCart.value || !skipShipping;
    let loadPayment = !skipPayment;

    if (loadShipping) {
        isLoading["shippingMethods"] = true;
    }

    if (loadPayment) {
        isLoading["paymentMethods"] = true;
    }

    const promises = [
        loadPayment ? getPaymentMethods(REFRESH_OPTIONS.FORCE_RELOAD) : null,
        loadShipping ? getShippingMethods(REFRESH_OPTIONS.FORCE_RELOAD) : null,
    ];

    return Promise.any(promises).finally(() => {
        if (loadShipping) {
            isLoading["shippingMethods"] = false;
        }

        if (loadPayment) {
            isLoading["paymentMethods"] = false;
        }
    });
};
```

.changeset/unlucky-birds-pay.md
```diff
@@ -0,0 +1,6 @@
+---
+"vue-demo-store": minor
+---
+
+- Added `refreshShoppingAndPaymentMethods` method that is used for rules scenarios that are related to the shipping or payment
```
patzick, line 5: shouldn't be `shipping` not `shopping`?
```suggestion
- Added `refreshShippingAndPaymentMethods` method that is used for rules scenarios that are related to the shipping or payment
```
also in the code

packages/composables/src/useCartErrorParamsResolver/useCartErrorParamsResolver.ts
```diff
@@ -2,7 +2,7 @@ import { useCart } from "#imports";
 
 export type UseCartErrorParamsResolver = {
   resolveCartError: (errorObject: ErrorObjectParam) => {
-    params: { [key: string]: string | number | null } | null;
+    params: { [key: string]: string | number | null | undefined } | null;
```
patzick, line 5: why not `?` then?
```suggestion
    params: { [key: string]?: string | number | null } | null;
```

packages/helpers/src/cms/getCmsTranslate.ts
```diff
@@ -7,7 +7,7 @@
  */
 export function getCmsTranslate(
   key: string,
-  params?: { [key: string]: string | number | null } | null,
+  params?: { [key: string]: string | number | null | undefined } | null,
```
patzick, line 10: same with the undefined definition
```suggestion
  params?: { [key: string]?: string | number | null } | null,
```

templates/vue-demo-store/pages/checkout/index.vue
```diff
@@ -74,6 +74,7 @@ const selectedShippingMethod = computed({
   async set(shippingMethodId: string) {
     isLoading[shippingMethodId] = true;
     await setShippingMethod({ id: shippingMethodId });
+    await refreshShoppingAndPaymentMethods({ skipShipping: true });
```
patzick, line 77: skipping makes the logic much more complicated, wouldn't be simpler to have two separate methods and just calling what you need?

templates/vue-demo-store/pages/checkout/index.vue
```diff
@@ -204,12 +209,20 @@ const placeOrder = async () => {
...
   isLoading["placeOrder"] = true;
-  const order = await createOrder();
-  isLoading["placeOrder"] = false;
-  await push("/checkout/success/" + order.id);
-  refreshCart();
+
+  try {
+    const order = await createOrder();
+    console.log("order", order);
+    isLoading["placeOrder"] = false;
```
patzick, line 219: this should go to the `finally` block as in case of error loading will be always `off`

templates/vue-demo-store/pages/checkout/index.vue
```diff
@@ -204,12 +209,20 @@ const placeOrder = async () => {
...
+    await push("/checkout/success/" + order.id);
+    refreshCart();
+  } catch (error) {
+    if (error instanceof ApiClientError)
+      console.log("error", error.details.errors);
```
patzick, line 224: maybe pushing some error message to the client to notify what went wrong?

templates/vue-demo-store/pages/checkout/index.vue
patzick: I'd still go for two separate methods for readability, maintenance and DX

patzick (changes requested):
Few notes from me, mostly I'd propose splitting it to two methods, as already there are cases when we need to use skipping

templates/vue-demo-store/pages/checkout/index.vue
mdanilowicz: done

.changeset/unlucky-birds-pay.md
mdanilowicz: ups, done, thank you

packages/composables/src/useCartErrorParamsResolver/useCartErrorParamsResolver.ts
mdanilowicz: Are you sure that this is a valid type? `?` - throws type syntax error

templates/vue-demo-store/pages/checkout/index.vue
mdanilowicz: I have chosen the one-method approach because all business logic is included in this function, like virtual product support and loaders. But I also don't think one function per method is wrong, so sure, why not, done.

kstala (approved):

packages/composables/src/useCartErrorParamsResolver/useCartErrorParamsResolver.ts
patzick: yeah, you might be right in that case, so let's leave it like this :)

patzick (approved):
👌

Issue #1326, opened 2024-10-02
[BUG] api-get overrides cause ENOENT open '.../_DELETE_' error


fabianeichinger:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The command `npx @shopware/api-gen generate --apiType=store` may fail with a confusing error message due to a bad interaction between the current default schema overrides and the example store schema json.

```
Error while generating types. Checkout the OpenAPI Schema and try again.
 Error: ENOENT: no such file or directory, open '{cwd}/_DELETE_'
    at Object.openSync (node:fs:596:3)
    at Object.readFileSync (node:fs:464:35)
    at load (file://{cwd}/node_modules/openapi-typescript/dist/load.js:73:33)
    at file://{cwd}/node_modules/openapi-typescript/dist/load.js:152:26
    at walk (file://{cwd}/node_modules/openapi-typescript/dist/utils.js:27:5)
    at walk (file://{cwd}/node_modules/openapi-typescript/dist/utils.js:29:9)
    at walk (file://{cwd}/node_modules/openapi-typescript/dist/utils.js:29:9)
    at walk (file://{cwd}/node_modules/openapi-typescript/dist/utils.js:29:9)
    at walk (file://{cwd}/node_modules/openapi-typescript/dist/utils.js:29:9)
    at walk (file://{cwd}/node_modules/openapi-typescript/dist/utils.js:29:9) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '{cwd}/_DELETE_'
}
```

### Expected Behavior

A schema should still be generated or the error should be more descriptive.

### Steps To Reproduce

1. Create a folder `api-types`
1. Download https://github.com/shopware/frontends/blob/ebb10eba629b3ec2c5a4a50fa12ef0b134601d6f/packages/api-client/api-types/storeApiSchema.json and https://github.com/shopware/frontends/blob/ebb10eba629b3ec2c5a4a50fa12ef0b134601d6f/packages/api-client/api-types/storeApiSchema.overrides.json to `api-types`.
1. Delete [line 6541](https://github.com/shopware/frontends/blob/ebb10eba629b3ec2c5a4a50fa12ef0b134601d6f/packages/api-client/api-types/storeApiSchema.json#L6541) from `storeApiSchema.json`.
1. Run `npx @shopware/api-gen@0.0.0-canary-20241002082236 generate --apiType=store`

### Environment

- OS: macOS (native and docker)
[...]

PR #1330, opened 2024-10-04
fix(api-gen): ignore internal _DELETE_ value for $ref in schema overrides input

mkucmus:
### Description

closes: #1326 

### ToDo's


BrocksiNet (approved):
The change is fixing the error. I am just curious if we hide some error/problem in schema or if this is an error from `json5` :thinking:

mkucmus:
> The change is fixing the error.
> 
> I am just curious if we hide some error/problem in schema or if this is an error from `json5` 🤔

It's how `json5` copes with $ref's - it tries to load it from some other file if it's not available in the same file. and `_DELETE_` is used for our internal transformer.

packages/api-gen/src/jsonOverrideUtils.ts
```diff
@@ -71,7 +71,13 @@ export async function loadJsonOverrides({
       const jsonOverridesFile = await readFileSync(pathToResolve, {
         encoding: "utf-8",
       });
-      const content = json5.parse(jsonOverridesFile);
+      const content = json5.parse(jsonOverridesFile, (key, value) => {
```
patzick, line 74: can we move this to outside helper file and add tests?
So then we have documentation in tests what kind of results we're expecting in case of any errors

patzick (changes requested):
I like it, let's just extract it as separate function and add some tests for documentation purposes :)

packages/api-gen/src/jsonOverrideUtils.ts
mkucmus: yeah, good idea - I'm on it rn! 
should I also apply the same callback for any `json5.parse` invokation? Now I can see I missed some of the places, but to be honest, the local `*overrides.json` is the only place where this special `_DELETE_` mapping can be a case, or not? 🤔

patzick (approved):

PR #1338, opened 2024-10-08
fix: setCurrentFilters function

### LINKED ISSUE #907: [BUG]  setCurrentFilters-method of useListing-composable not working

## Problem
Currently when using the setCurrentFilters method of useListing composable with any filter, the search does not inject the filter into the API request body properly.

For example when calling the method as follows, a request to the client is made but filters are not applied properly:

**Example**
```ts
// const {filterResults} = useListing({ /* omitted */ });

function filterResults(filter) {
  // ... transformers, custom stuff...etc.
  serCurrentFilters({
    code: 'manufacturer',
    value: '018d35f5b5757076adea38044bb96937' // 👈 manufacturer ID from my Shopware
  });
  // ... maybe more stuff
}
```

**Request body to the Shopware API**
```JSON
{
  "limit":10,
  "search":"",
  "p":"1",
  "navigationId":"018db20234207be8948e3a4b46501435",
  "manufacturer":"",
  "price": {"min":0,"max":0},
  "rating": null,
  "shipping-free":false,
  "properties":"",
  "code":"manufacturer",
  "value":"018d35f5b5757076adea38044bb96937"
}
```

## Cause
I've tracked this problem to the code of setCurrentFilters-method:
```ts
const setCurrentFilters = (filter: { code: string; value: any }) => {
    const appliedFilters: RequestParameters<"searchPage"> = Object.assign(
      {},
      getCurrentFilters.value,
      filter, // 🚨 Here's the problem
      {
        query: getCurrentFilters.value?.search,
        manufacturer: getCurrentFilters.value?.manufacturer?.join("|"),
        properties: getCurrentFilters.value?.properties?.join("|"),
      },
    );
    if (_storeAppliedListing.value) {
      _storeAppliedListing.value.currentFilters = {
        ...appliedFilters,
        manufacturer: appliedFilters.manufacturer?.split("|"),
        properties: appliedFilters.properties?.split("|"),
      };
    }
    return search(appliedFilters);
  };
```
This method does not apply the filters as Shopware API reads them, as it simply adds "code" and "value" property into the `const appliedFilters` object. 

What it should do, is to either overwrite default properties of the `const appliedFilters`.

### Expected Behavior

setCurrentFilters should apply the key and value to appliedFilters -object in a way that Shopware API wants them. 

Consider calling setCurrentFilters like this:
```ts
function filterResults(filter) {
  serCurrentFilters({
    code: 'manufacturer',
    value: '018d35f5b5757076adea38044bb96937' // 👈 manufacturer ID from my Shopware
  });
}
```

The request should be like this:
```JSON
{
  "limit":10,
  "search":"",
  "p":"1",
  "navigationId":"018db20234207be8948e3a4b46501435",
  "manufacturer":"018d35f5b5757076adea38044bb96937",  // 👈 where the filter value should go
  "price": {"min":0,"max":0},
  "rating": null,
  "shipping-free":false,
  "properties":"",
}
```

Not like this:

```JSON
{
  "limit":10,
  "search":"",
  "p":"1",
  "navigationId":"018db20234207be8948e3a4b46501435",
  "manufacturer":"",
  "price": {"min":0,"max":0},
  "rating": null,
  "shipping-free":false,
  "properties":"",
  "code":"manufacturer",  // 👈 Not like this
  "value":"018d35f5b5757076adea38044bb96937" // 👈 Not like this
}
```

### Steps To Reproduce

1. Set up Nuxt project with the Shopware Nuxt modules
2. Create listing using `useListing()` -composable
3. Create event that calls `setCurrentFilters()` -method with argument like this for example:
```ts
setCurrentFilters({
  code: 'manufacturer',
  value: 'any-manufacturer-id`
})
```

Shopware version: 6.5.7.3 
Shopware-composables version: 0.0.0-canary-20240426081217 (see: [Github, composables, useListing]( https://github.com/shopware/frontends/blob/main/packages/composables/src/useListing.ts#L479))

### Environment

```markdown
- OS: MacOS Sonoma, 14.4.1 (23E224)
- Node: v18.18.0
- yarn: v1.22.19
```

### Anything else?

_No response_

BrocksiNet:
### Description

closes #907

### Type of change

### Additional context

I see this as a hotfix until we improve useListing composable.

### DIFF
--- a/packages/composables/src/useListing/useListing.ts
+++ b/packages/composables/src/useListing/useListing.ts
@@ -36,6 +36,17 @@ function merge<T extends { [key in keyof T]: unknown }>(
 
 export type ListingType = "productSearchListing" | "categoryListing";
 
+export type ShortcutFilterParam = {
+  code:
+    | "manufacturer"
+    | "min-price"
+    | "max-price"
+    | "rating"
+    | "shipping-free"
+    | "properties";
+  value: string | number | boolean;
+};
+
 export type UseListingReturn = {
   /**
    * Listing that is currently set
@@ -146,7 +157,7 @@ export type UseListingReturn = {
    * @param filters
    * @returns
    */
-  setCurrentFilters(filters: { code: string; value: unknown }): Promise<void>;
+  setCurrentFilters(filters: ShortcutFilterParam[]): Promise<void>;
   /**
    * Indicates if the listing is being fetched
    */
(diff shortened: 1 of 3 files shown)


packages/composables/src/useListing/useListing.ts
```diff
@@ -36,6 +36,17 @@ function merge<T extends { [key in keyof T]: unknown }>(
...
 export type ListingType = "productSearchListing" | "categoryListing";
 
+export type ShortcutFilterParam = {
+  code:
+    | "manufacturer"
+    | "min-price"
+    | "max-price"
+    | "rating"
+    | "shipping-free"
+    | "properties";
```
patzick, line 46: what about custom filters?
This probably should be a `keyof ProductListingCriteria` from the Schema

patzick (changes requested):
Very nice improvement! I'd just change manually typed filters to the ones from the schema, so it's more flexible

packages/composables/src/useListing/useListing.ts
BrocksiNet: Would currently not work with setCurrentFilters but was also never working :thinking:

patzick (approved):

PR #1360, opened 2024-10-16
docs(examples): api-client tutorial made with StackBlitz TutorialKit

mkucmus:
### Description

preview: https://mkucmus.github.io/api-client-tutorial/

please note that webcontainers don't run without extra cors headers set up - [more info](https://webcontainers.io/guides/troubleshooting#webcontainers-not-loading-and-postmessage-error)

running example of the current stage: https://stackblitz.com/github/shopware/frontends/tree/feat/api-client-tutorial/examples/api-client-tutorial

TODO:
- [x] add more steps
- [ ] cover admin api instance
- [x] adjust styling
- [x] move to examples
- [x] fix it to be able to run via pnpm as well


BrocksiNet:
I like this kind of tutorial :blue_heart: great work!

I have only one question :grin: 
Can we embed this into our docs without redirecting to StackBlitz?

That would be awesome. :rocket:

mkucmus:
> I like this kind of tutorial 💙 great work!
> 
> I have only one question 😁 Can we embed this into our docs without redirecting to StackBlitz?
> 
> That would be awesome. 🚀

thanks! this is possible, but we would need to hack a vitepress router - and this is a little bigger thing to do. 

the idea is to place a link to this tutorial in both README's: api-gen and api-client packages.

patzick (approved):

PR #1369, opened 2024-10-17
feat(vue-demo): add dynamic breadcrumbs

### LINKED ISSUE #139: [BUG] Category breadcrumbs path is missing

mdanilowicz:
### Description

This pull request adds a new request to fetch category breadcrumbs. The request is included on the category and product pages, while all other static pages build breadcrumbs with "local" data.

Added `getCategoryBreadcrumbs` method for fetching backend breadcrumbs in `useBreadcrumbs` composable
Added `buildDynamicBreadcrumbs` method for building breadcrumbs structure
Added `pushBreadcrumb` method to push single breadcrumb at the top of the breadcrumbs list
Added `associations` option to the `search` method in `useProductSearch` composable

closes #139


BrocksiNet:
@kstala could you test this please :raised_hands:

templates/vue-demo-store/components/FrontendNavigationPage.vue
```diff
@@ -30,11 +29,8 @@ if (!categoryResponse.value) {
   throw error.value;
 }
 
-const breadcrumbs = getCategoryBreadcrumbs(categoryResponse.value, {
-  startIndex: 2,
-});
-
-useBreadcrumbs(breadcrumbs);
+const { buildDynamicBreadcrumbs } = useBreadcrumbs();
+buildDynamicBreadcrumbs(props.navigationId);
```
patzick, line 33: it's happening asynchronously, so I have two questions:
1. how are you assuring it's loaded on Server Side Render?
2. isn't there a blink on client side? What about showing previous breadcrumbs and replace them when new ones are loaded? Does it make sense?

packages/composables/src/useBreadcrumbs/useBreadcrumbs.ts
```diff
@@ -48,8 +66,38 @@ export function useBreadcrumbs(
     _breadcrumbs.value = [];
   };
 
+  const getCategoryBreadcrumbs = async (categoryId: string) => {
+    const response = await apiClient.invoke(
+      "readBreadcrumb get /breadcrumb/{id}",
```
patzick, line 71: this endpoint is not only to load category breadcrumbs, I'm not sure about exposing it here, what is the designed flow here looks like?

packages/composables/src/useBreadcrumbs/useBreadcrumbs.ts
```diff
@@ -48,8 +66,38 @@ export function useBreadcrumbs(
...
+      {
+        pathParams: {
+          id: categoryId,
+        },
+      },
+    );
+    return response.data;
+  };
+
+  const buildDynamicBreadcrumbs = async (categoryId: string) => {
```
patzick, line 81: what about product breadcrumbs?

packages/composables/src/useBreadcrumbs/useBreadcrumbs.ts
```diff
@@ -48,8 +66,38 @@ export function useBreadcrumbs(
...
+      _breadcrumbs.value = breadcrumbs.breadcrumbs.map((breadcrumb) => {
+        // Adjust path to be compatible with the router
+        return {
+          ...breadcrumb,
+          path: `/${breadcrumb.path}`,
+        };
+      });
+    } catch (error) {
+      console.error("Error while fetching breadcrumbs", error);
+    }
```
patzick, line 94: we should not handle errors in composable, it should be thrown to invoking component to handle in the project - analytics/notifications etc. however they need to handle it

.changeset/funny-hotels-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-pwa/composables-next": minor
+---
+
+Added `getCategoryBreadcrumbs` method for fetching backend breadcrumbs in `useBreadcrumbs` composable
```
patzick, line 5: the `buildDynamicBreadcrumbs` is more important functionality, should be mentioned and explained here more

packages/helpers/src/cms/getCmsBreadcrumbs.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Build the breadcrumbs for the CMS page
+ *
+ * @param page
+ * @returns
+ */
+export function getCmsBreadcrumbs<
```
patzick, line 7: not sure about this helper - it's not creating breadcrumbs, just displaying entity translated property, when would that be useful for the users?

patzick (commented):
Good feature incoming! We need to clarify a few things here first :)

patzick:
@mdanilowicz there are many unanswered comments in my previous CR. Please address them and explain the concept and the flow in the PR description, no need to rewrite it before agreeing on the API of these packages 🙌

packages/composables/src/useBreadcrumbs/useBreadcrumbs.ts
patzick: the endpoint is for both and takes more parameters
image


I just find `getCategoryBreadcrumbs` misleading as I would search for other types of breadcrumbs

patzick: My point is - I would not keep this endpoint in the composable, it does not interfere with the logic and everyone can just use apiClient method for that and pass the data through `buildDynamicBreadcrumbs`

patzick (changes requested):
I just have one more hesitation for useBreadcrumbs and I think the rest is in very good shape!

packages/composables/src/useBreadcrumbs/useBreadcrumbs.ts
mdanilowicz: done

BrocksiNet (approved):
There are some conflicts. But good work. :muscle:

.changeset/funny-hotels-watch.md
```diff
@@ -0,0 +1,8 @@
...
```
patzick, line 5:
```suggestion
```

patzick (approved):

PR #1370, opened 2024-10-18
feat(composable): add useMedia composable

mdanilowicz:
### Description

Add useMedia composable

### DIFF
--- a/packages/composables/src/useMedia/useMedia.ts
+++ b/packages/composables/src/useMedia/useMedia.ts
@@ -1,0 +1,35 @@
+import { useShopwareContext } from "#imports";
+import type { operations } from "#shopware";
+
+interface UseMedia {
+  /**
+   * Fetches media by ids
+   * @param {string[]} ids
+   */
+  fetchMedia(
+    ids: string[],
+  ): Promise<operations["readMedia post /media"]["response"]>;
+}
+
+/**
+ *
+ * Composable for media handling
+ *
+ * @returns {UseMedia} media composable
+ */
+export function useMedia(): UseMedia {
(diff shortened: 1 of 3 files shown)


pnpm-lock.yaml
BrocksiNet, line 1: pnpm-lock.yaml changes needed? :thinking:

packages/composables/src/useMedia/useMedia.ts
```diff
@@ -0,0 +1,35 @@
...
+  ): Promise<operations["readMedia post /media"]["response"]>;
+}
+
+/**
+ *
+ * Composable for media handling
+ *
+ * @returns {UseMedia} media composable
+ */
+export function useMedia(): UseMedia {
```
mkucmus, line 20: I would like to see what kind of media this composable is made for. Maybe we could add a parameter in type `Media` for composable function, and then we could extract ids from argument and replace/merge it with `ids` provided in `fetchMedia` function which could then become optional. 

the second option to tell something more about the purpose is providing an example in functions' description and changeset file.

packages/composables/src/useMedia/useMedia.ts
```diff
@@ -0,0 +1,35 @@
+import { useShopwareContext } from "#imports";
+import type { operations } from "#shopware";
+
+interface UseMedia {
```
mkucmus, line 4: it's our first `interface` within composables, maybe we could keep it consistent and change to the type 😉

mkucmus (commented):
the composable we all needed! 🥇 

I left some small notes - not crucial but who knows :)

PR #1371, opened 2024-10-21
feat(api-client): new onDefaultHeaderChanged hook to track default headers modifications

patzick:
### Description
This PR brings the new hook `onDefaultHeaderChanged` which (in both store and api client) is called on the default header change - invoked either by the user or by the client itself (in reaction for API responses)

to test it:
```typescript
apiClient.hook("onDefaultHeaderChanged", (key, value) => {
  // here we can detect that the default header has changed, either by the user or by the headers incoming from the API
});
```


mkucmus (approved):

mkucmus:
Looks exactly what is needed. Although there is a _significant_ performance breakdown noticed - it still performs well: 129 µs (1 µs is 1/1 000 000 s) 👍🏼

PR #1396, opened 2024-10-28
fix(vue-demo): close minicart after route update

### LINKED ISSUE #1307: [BUG]  Open empty mini cart on thankyou page

### Current Behavior

Screen Recording 2024-09-23 at 16.58.58

### Expected Behavior

the open minicart will not be displayed

### Steps To Reproduce

add anything to your cart and place your order

### Environment

```markdown
- OS:
- Node:
- pnpm:
```

### Anything else?

_No response_

mdanilowicz:
### Description

 closes #1307


patzick (changes requested):
Not every layout is using the minicart, would be better to place it in the root of where it's used, so that would be the `LayoutHeader`

patzick (approved):

PR #1404, opened 2024-10-30
feat: page scrolling

mdanilowicz:
### Description

Add smooth scrolling on the routing change (excluding product variants)


templates/vue-demo-store/app/router.options.ts
```diff
@@ -0,0 +1,29 @@
+import type { RouterOptions } from "@nuxt/schema";
+
+export default <RouterOptions>{
+  scrollBehavior(to, from, savedPosition) {
+    /**
+     * Do not scroll to top when switching between product variants
+     *
+     * Assumptions:
+     * - product with variants contains minimum 2 subdirectories
```
mkucmus, line 9: we should add some more info (docs or here or both?) that the format of URL is not fixed - can be configured and then, the condition won't work. 
image

mkucmus (changes requested):
smooth routing is what we love 💙 

please add one info about the URL format - If I understand it correctly

kstala (approved):

mkucmus (approved):

.changeset/strange-pets-relax.md
```diff
@@ -0,0 +1,6 @@
+---
+"vue-demo-store": patch
+---
+
+- Add smooth scrolling on the routing change (excluding product variants)
+- Add smooth scrolling for listing pagination
```
patzick, line 6: that's another changeset entry in cms-base package

.changeset/strange-pets-relax.md
```diff
@@ -0,0 +1,6 @@
+---
+"vue-demo-store": patch
+---
+
+- Add smooth scrolling on the routing change (excluding product variants)
```
patzick, line 5:
```suggestion
Add smooth scrolling on the routing change (excluding product variants)
```

using `-` causes ugly changeset entry, like here:
image

patzick (changes requested):

.changeset/strange-pets-relax.md
patzick: yes, pure string without `-` is correct :)

patzick (approved):

.changeset/strange-pets-relax.md
mdanilowicz: Ahh :D @patzick made me realise that it was a diff `+` ... ups 🙃

PR #1405, opened 2024-10-31
feat(api-gen): updated openapi-typescript to v7

patzick:
### Description

replacement for outdated #1107

- upgraded openapi-typescript to v7
- updates schema definitions
- improved generation and problems detection


packages/api-gen/src/generateFile.ts
```diff
@@ -238,12 +242,12 @@ export async function prepareFileContent({
...
-                routePath,
-                // "method",
-                // method,
-              );
+              // console.error(
+              //   "No requests for method",
+              //   routePath,
+              //   // "method",
+              //   // method,
+              // );
```
BrocksiNet, line 250: Is that intended?

patzick: nope, good point! Fix on the way

PR #1414, opened 2024-11-04
fix: e2e missing package and linter

patzick:
### Description
- added biome linter and fixed linting errors
- added missing faker package
- exported `findEnv` to helper for reusability
- using `sdk` from the package


kstala:
@patzick Unfortunately in this solution, tests for stackblitz (template and exmaples) will not work

PR #1425, opened 2024-11-07
fix: override header value when invoke

quando1910:
### Description
- Got the unexpected behavior when try to override the default headers with adminApiClient. because of passing wrong params position into defu function

Screen Shot 2024-11-07 at 4 17 02 PM

### Reproduce
- Set defaut headers: `adminApiClient.defaultHeaders.apply({ "sw-context-token": "xxxxx" });`

- override when invoke:
```
 const response = await adminApiClient.invoke("path", {
   headers: {
      "sw-context-token": "yyyyy"
   }
});
```

- Request was called with sw-context-token = xxxxx. The expectation should be called with sw-context-token = yyyyy

### Type of change
Bug fix


patzick (changes requested):
NIce catch!
Please add test case for that in `createAdminApiClient.test.ts` and a patch changelog entry (instruction: https://github.com/shopware/frontends/blob/main/CONTRIBUTION.md#changelog-preparation )

.changeset/large-countries-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware/api-client": patch
+---
+
+- fix: override header value when invoke adminApiClient ans storeAPiClient
```
patzick, line 5:
```suggestion
- fix: override header value when invoke adminApiClient and storeAPiClient
```

patzick (approved):
Thanks @quando1910 !

PR #1434, opened 2024-11-07
fix: set correct access token when it expired

quando1910:

### DIFF
--- a/.changeset/fair-cows-repeat.md
+++ b/.changeset/fair-cows-repeat.md
@@ -1,0 +1,5 @@
+---
+"@shopware/api-client": patch
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/fair-cows-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware/api-client": patch
+---
+
+- fix: set the correct access token in headers if it has expired
```
patzick, line 5:
```suggestion
replaces authentication header instead of appending, when session has expired and is being refreshed
```

patzick (approved):
Very nice, thanks @quando1910 🙌

PR #1441, opened 2024-11-11
feat: add basic axe core asseccibility check

kstala:
### Description
Adding basic accessibility tests using Ax Core for the home page, category and product page.
Adding a separate workflow for accessibility check.
Adding additional tag (accessibility)


patzick (commented):
Looks good, just one thing to clarify

patzick (approved):

PR #1442, opened 2024-11-12
feat: add API headers from config

### LINKED ISSUE #964: [FEATURE] Add apiClient default headers to nuxt config
### Description

We should be able to set up with nuxt config the default headers, which are used by `apiClient` instance.

### Use Case

someone using nuxt-module and wanting to add global headers in a convenient place

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

mdanilowicz:
### Description

 closes #964

### DIFF
--- a/.changeset/stale-files-learn.md
+++ b/.changeset/stale-files-learn.md
@@ -1,0 +1,5 @@
+---
+"@shopware/api-client": minor
+---
+
+Add config default headers
--- a/.changeset/stale-files-learn.md
+++ b/.changeset/stale-files-learn.md
@@ -1,5 +1,5 @@
 ---
-"@shopware/api-client": minor
+"@shopware-pwa/nuxt3-module": minor
 ---
 
-Add config default headers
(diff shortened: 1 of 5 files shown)


.changeset/stale-files-learn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware/api-client": minor
```
patzick, line 2: that's wrong package for changeset, it should be nuxt module package

.changeset/stale-files-learn.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Add config default headers
```
patzick, line 5: please write a bit more, so people reading the changelog can adapt to changes

patzick (changes requested):
Very much needed feature! Just some docs to clarify

.changeset/stale-files-learn.md
patzick: very nice!

patzick (approved):
all good 👌

PR #1447, opened 2024-11-12
fix(api-gen): error messages and merging patches

patzick:
### Description

- improved error messages
- improved merging overrides
- sorting arrays, so always generated in the same order
- regenerated schema files

### DIFF
--- a/packages/api-gen/src/patchJsonSchema.ts
+++ b/packages/api-gen/src/patchJsonSchema.ts
@@ -20,7 +20,7 @@ export const extendedDefu = createDefu((obj, key, value) => {
   if (Array.isArray(obj[key]) && Array.isArray(value)) {
     // concat arrays but skip duplicates
     // @ts-expect-error - we know that obj[key] is an array as we're inside this if statement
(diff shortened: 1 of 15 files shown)


mdanilowicz (approved):

packages/api-gen/src/patchJsonSchema.ts
```diff
@@ -20,7 +20,7 @@ export const extendedDefu = createDefu((obj, key, value) => {
   if (Array.isArray(obj[key]) && Array.isArray(value)) {
     // concat arrays but skip duplicates
     // @ts-expect-error - we know that obj[key] is an array as we're inside this if statement
-    obj[key] = [...new Set([...obj[key], ...value])];
+    obj[key] = [...new Set([...obj[key], ...value])].sort();
```
mdanilowicz, line 23: Why is it important to sort it out?

patzick: for overriding - depending on which source came first it was looking random, also snapshots often changed when I experimented with tests, this way we keep consistency

PR #1469, opened 2024-11-20
feat: add onRequest hook to createAPIClient

JimTacobs:
### Description

Allow for usage of the `onRequest` interceptor that's provided by ofetch. Added onRequest to the ApiClientHooks and applied it in a similar way that was done for other hooks.

Added four testcases to make sure it works in the way you'd expect when using this hook.

Added a bit of info to the readme to make sure people know they can use it. 

### Type of change

### ToDo's

None


patzick:
I wouldn't encourage people to add custom headers via the hook, rather use global headers or `headers` option in request. I see the usecase for analytics purpose as an example. When would you feel you need to modify headers with this hook?

PR #1476, opened 2024-11-24
chore: fix api-gen typo

freb97:
### Description

Fixes small typo in loadSchema.
Don't think anyone really ran into problems with this, since the commands expects the base url anyway.

### Type of change

Typo fixed

### ToDo's

None


patzick (approved):
Good point, thank you! :)

PR #1489, opened 2024-11-27
fix(vue-demo): double opt in registration

### LINKED ISSUE #1462: [BUG] Customer Registration with Double Opt In is broken

### Current Behavior

With "Double opt-in on sign-up" enabled in Shopware, you're redirected to the home page after submission of the customer registration form. The app behaves like it's logged in, e.g. in vue-demo-store will show the new customer name in the header. Clicking a link to navigate to `/account` will trigger a warning notification "Login is required to access this page. You are redirected to the homepage."

The page stays broken until a full page reload, where you'll be in a logged-out state again.

### Expected Behavior

The app should notify the customer to check their emails after the registration form submission. It should not behave like it's logged in.

### Steps To Reproduce

1. Enable "Double opt-in on sign-up" in a current Shopware version.
2. Register from a vue-demo-store frontend.
3. Observe behavior mentioned above

### Environment

```markdown
- Shopware: 6.6.8.2 (clean dockware image)
- vue-demo-store as of ea9249f
```

### Anything else?

I'm pretty sure that this used to work. It might have been broken by a Shopware update. I assume it's because the `active` flag on customers used be `false` until double opt in was confirmed. Now it's `true` immediately instead and you have to check DOI state differently.

mdanilowicz:
### Description

This pull request includes changes to improve user registration and session handling.

Improvements to user registration and session handling:

* [`packages/composables/src/useUser/useUser.ts`](diffhunk://#diff-f6055843be7b7f31df7f2cd21efca6578d79ef68a42f71d128de3688761caabaR183-R188): Updated the `useUser` function to set user data when the user is active and not in double opt-in registration, and to refresh the session context accordingly.
* [`templates/vue-demo-store/components/account/AccountRegisterForm.vue`](diffhunk://#diff-5a57ed38ab2370278d457e6e57d5f6a1105e389c329940af29d94c60b86ddd5eL112-R118): Modified the `invokeSubmit` function to handle double opt-in registration flow correctly and navigate active users to the home page.

Added a confirmation page after the double opt-in option. The confirmation link is included in the email after registration.

 closes #1462


.changeset/funny-chicken-play.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-pwa/composables-next": minor
+---
+
+Set user details when fully active
```
patzick, line 5: I'm not sure what should I understand (as a user) from this changelog entry:
- in which composable?
- what does it mean?
- what impact does it have on my code?

is this really a feature or bugfix?

.changeset/light-games-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+"vue-demo-store": minor
+---
+
+Fix registration form for the double opt in registration flow
```
patzick, line 5: sounds like a patch?

packages/composables/src/useUser/useUser.ts
```diff
@@ -180,8 +180,12 @@ export function useUser(): UseUserReturn {
         storefrontUrl: getStorefrontUrl(),
       },
     });
-    _user.value = data;
-    if (_user.value?.active) await refreshSessionContext();
+    // Set data when uses is active and not double opt in
```
patzick, line 183:
```suggestion
    // Set data when user is active and double opt in feature is not activated in the shopware backend
```

templates/vue-demo-store/components/account/AccountRegisterForm.vue
```diff
@@ -109,14 +109,13 @@ const invokeSubmit = async () => {
...
-      if (response && response.active) router.push("/");
-      else if (response && !response.active) {
+      if (response && response.doubleOptInRegistration) {
         Object.assign(state, JSON.parse(JSON.stringify(initialState)));
         showDoubleOptInBox.value = true;
         await nextTick();
         doubleOptInBox.value?.scrollIntoView();
         $v.value.$reset();
-      }
+      } else if (response && response.active) router.push("/");
```
patzick, line 118: looks like there is no info for user with that flow, they are neither lonned in or have info to check the email, right?

.changeset/funny-chicken-play.md
mdanilowicz: Because we do not have a version in relation to the backend, this is a feature. We are implementing a new flow.
Changed

templates/vue-demo-store/components/account/AccountRegisterForm.vue
patzick: cool, but one more thing - after the registration the form should not be visible anymore, just the notification
image

packages/composables/src/useUser/useUser.ts
```diff
@@ -180,8 +180,12 @@ export function useUser(): UseUserReturn {
         storefrontUrl: getStorefrontUrl(),
       },
     });
-    _user.value = data;
-    if (_user.value?.active) await refreshSessionContext();
+    // Update the user data in the context if the user is active and not using double opt-in registration set in the Shopware Admin
+    if (data.active && !data.doubleOptInRegistration) {
+      _user.value = data;
+      await refreshSessionContext();
```
patzick, line 186: shouldn't we still refresh the context? User should not be in it

templates/vue-demo-store/components/account/AccountRegisterForm.vue
mdanilowicz: This is the same flow we have on the storefront. We decided to keep them as similar as possible

packages/composables/src/useUser/useUser.ts
mdanilowicz: hmm, I don't see the reason to refresh the context if the user is not "let in". Do you see any specific reason? Maybe I'm missing something.

patzick: just to "make sure" - the action is pretty relevant from the system and user perspective. If there is some another "flag" or business logic allowing to use the account for some time etc.
this way we're sure that the view is in sync with the backend state

templates/vue-demo-store/components/account/AccountRegisterForm.vue
mdanilowicz: done

packages/composables/src/useUser/useUser.ts
mdanilowicz: done

templates/vue-demo-store/components/account/AccountRegisterForm.vue
mdanilowicz: I created a task to format this page https://github.com/shopware/frontends/issues/1496

BrocksiNet (approved):
It works :tada:, but it redirects the customer to the default Storefront when he accepts the double opt-in link. Do we already have a task to fix that?

Also, we have some files with conflicts.

patzick (approved):
Looks good, let's link the issue @BrocksiNet mentioned and we can merge it!

mdanilowicz:
@BrocksiNet you mean the link in the email? This is a storefront domain configuration, we cannot change it on the cloud because of the domain verification

patzick:
@kstala can you check e2e here?

mdanilowicz:
@patzick double-opt-in was enabled

Issue #1403, opened 2024-10-29
[FEATURE] node.js update


mkucmus:
### Description

update and test the monorepo with recent LTS which is now 22.11

### Use Case

https://nodejs.org/en/blog/release/v22.11.0

PR #1490, opened 2024-11-27
feat: Update to node 22

patzick:
### Description

as vercel added support for node 22 we can update our repo as well

closes: #1403


mkucmus (approved):
what a significant improvement!

Issue #1435, opened 2024-11-07
[BUG] useOrderDetails.paymentChangeable is always false


fabianeichinger:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The computed `paymentChangeable` from `useOrderDetails`  is always false, even if the method may be changed.

### Expected Behavior

The value of `paymentChangeable` from `useOrderDetails` should be true if the payment method on the order can be changed.

### Steps To Reproduce

```
const { loadOrderDetails, paymentChangeable } = useOrderDetails(
  props.orderId,
);
await loadOrderDetails();
paymentChangeable.value // => always false
```

### Anything else?

`checkPromotion` is not sent as part of the request body one the readOrder request. Pretty sure this is a regression caused by the following change, as `checkPromotion` is in `mergedAssociations` but not `mergedAssociations.associations`.
https://github.com/shopware/frontends/commit/19540220d87788eed08991d35aaaead2e18564e5#diff-1aae20791a13b17b807d9677e2a8ba59b0bb6d3aa8716bc0a80d5ed09af61d41L205-R218

mkucmus:
hey! in order to see the paymentChangeable attribute filled with the order ids in the `loadOrderDetails` response, please pass the `checkPromotions: true` flag, If you are using the vue-demo-store template, the proper place is somewhere [here](https://github.com/shopware/frontends/blob/main/templates/vue-demo-store/pages/account/order.vue#L86). 

We are going to revert the change as soon as possible.

PR #1492, opened 2024-11-28
fix(composables): add `checkPromotions` flag to the `loadOrders` params of `useCustomerOrders` composable

mkucmus:
### Description
closes: #1435

### DIFF
--- a/.changeset/unlucky-dogs-help.md
+++ b/.changeset/unlucky-dogs-help.md
@@ -1,0 +1,5 @@
+---
+"@shopware-pwa/composables-next": patch
+---
+
+Add checkPromotions flag for customer's /order request signature
--- a/.changeset/unlucky-dogs-help.md
+++ b/.changeset/unlucky-dogs-help.md
@@ -2,4 +2,4 @@
 "@shopware-pwa/composables-next": patch
 ---
 
-Add checkPromotions flag for customer's /order request signature
+`useCustomerOrders` - added checkPromotions flag for loading orders
--- a/packages/composables/src/useCustomerOrders/useCustomerOrders.ts
+++ b/packages/composables/src/useCustomerOrders/useCustomerOrders.ts
@@ -1,7 +1,7 @@
 import { ref, computed } from "vue";
 import type { Ref, ComputedRef } from "vue";
 import { useShopwareContext } from "#imports";
-import type { Schemas } from "#shopware";
+import type { Schemas, operations } from "#shopware";
 
 export type UseCustomerOrdersReturn = {
   /**
@@ -18,7 +18,7 @@ export type UseCustomerOrdersReturn = {
    * Fetches the orders list and assigns the result to the `orders` property
    */
   loadOrders(
-    parameters?: operations['readOrder post /order']['body'],
+    parameters?: operations["readOrder post /order"]["body"],
   ): Promise<void>;
   /**
    * Current page number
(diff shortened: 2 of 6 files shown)


packages/composables/src/useCustomerOrders/useCustomerOrders.ts
```diff
@@ -17,7 +17,9 @@ export type UseCustomerOrdersReturn = {
   /**
    * Fetches the orders list and assigns the result to the `orders` property
    */
-  loadOrders(parameters?: Schemas["Criteria"]): Promise<void>;
+  loadOrders(
+    parameters?: Schemas["Criteria"] & { checkPromotions?: boolean },
```
patzick, line 21: it's already in the readOrder schema, what about
```suggestion
    parameters?: operations['readOrder post /order']['body'],
```

packages/composables/src/useCustomerOrders/useCustomerOrders.ts
```diff
@@ -50,7 +52,7 @@ export function useCustomerOrders(): UseCustomerOrdersReturn {
   const currentParams = ref<Schemas["Criteria"]>({});
 
   const loadOrders = async (
-    parameters: Schemas["Criteria"] = {},
+    parameters: Schemas["Criteria"] & { checkPromotions?: true } = {},
```
patzick, line 55:
```suggestion
    parameters: operations['readOrder post /order']['body'] = {},
```

.changeset/unlucky-dogs-help.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware-pwa/composables-next": patch
+---
+
+Add checkPromotions flag for customer's /order request signature
```
patzick, line 5: this flag was already working, so would be more like
```suggestion
`useCustomerOrders` - added checkPromotions flag for loading orders
```

patzick (changes requested):
I'd avoid having manual types and reuse schema types

mkucmus:
> I'd avoid having manual types and reuse schema types

totally right, I forgot that most of the types are already on hand :)

templates/vue-demo-store/pages/account/order.vue
```diff
@@ -85,6 +85,7 @@ await useAsyncData("getOrders", () => {
   return loadOrders({
     limit: limit.value,
     page: route.query.p ? Number(route.query.p) : defaultPage,
+    checkPromotions: true, // determines whether "paymentChangeable" list is available in the response
```
mdanilowicz, line 88: We should use here `orderAssociations`, because currently we have 2 ways of loading it

https://github.com/shopware/frontends/blob/main/packages/composables/src/useOrderDetails/useOrderDetails.ts#L140
https://github.com/shopware/frontends/blob/main/packages/composables/src/useDefaultOrderAssociations/useDefaultOrderAssociations.ts#L42

mdanilowicz (changes requested):

templates/vue-demo-store/pages/account/order.vue
mkucmus: "yes and no" 😄 
 maybe we could create a different helper for params in general because the `checkPromotion` flag is not an association, but the root level param.

mdanilowicz: But now, you cannot change this value by the injection. This property can affect the response time. So it is good to have a flexibility in it

mkucmus: the layout in the template is fully customizable and the useOrderDetails is for logged-in users only, not that bad imo so far. 
I'm not sayin that keeping `checkPromotion` or other params in adjustable composable is bad, but for now the composable is designed for associations. I will create a ticket for that case, wdyt?

mkucmus: https://github.com/shopware/frontends/issues/1493

patzick (approved):

PR #1498, opened 2024-11-29
feat: move cms-base package lonting to biome

patzick:
### Description

switched `cms-base` package linting to biome for better accuracy and performance
test: `time pnpm run lint` inside package

before
```
pnpm run lint  6.63s user 0.69s system 154% cpu 4.725 total
```

after
```
pnpm run lint  0.35s user 0.15s system 85% cpu 0.590 total
```

apart of the speed it was more accurate in finding problems and making the code more consistent


mkucmus (approved):
the times saved are impressive 🕐 ⏩ 🚤 

do we need to adjust our IDEs config in order to work with biome correctly?

patzick:
@mkucmus I'll add biome extension to list in VSCode config, but it's not mandatory, more like a helper

PR #1506, opened 2024-12-03
feat: move vue-demo-store to biome

patzick:
switched `vue-demo-store` package linting to biome for better accuracy and performance
test: `time pnpm run lint` inside package

before
```
pnpm run lint  22.16s user 2.47s system 151% cpu 16.283 total
```

after
```
pnpm run lint  0.43s user 0.07s system 118% cpu 0.416 total
```
apart of the speed it was more accurate in finding problems and making the code more consistent


patzick:
sidenote: needed to move biome back to 1.8.3 and turn off IDE plugin until https://github.com/biomejs/biome/issues/4684 is resolved. As that causes problem in `shopware.d.ts` files

mkucmus (approved):
🚀 `pnpm run lint  0.43s user 0.07s system 118% cpu 0.416 total`  why not 0.1s ? 😄

Issue #1270, opened 2024-09-05
[FEATURE] optimize `loadNavigationElements` method of `useNavigation`


mkucmus:
### Description

allow to pass a full search criteria according to the endpoints specification.
place to optimize: https://github.com/shopware/frontends/blob/main/packages/composables/src/useNavigation.ts#L18

### Use Case

allow users to optimize the response size or shape of the response in general.

### Additional Context

community discussion: https://shopwarecommunity.slack.com/archives/C050L6NCMGQ/p1725520199690769

mdanilowicz:
done in https://github.com/shopware/frontends/pull/1540

PR #1540, opened 2024-12-18
feat: allow to pass search criteria to loadNavigationElements of useNavigation

kasztof:
### Description

This closes the issue https://github.com/shopware/frontends/issues/1270 by allowing to pass full search criteria to the `loadNavigationElements` method of `useNavigation` composable.


patzick (changes requested):
Thank you for your contribution!
Opening params to match endpoint params is a good approach, let's just make it by extending params, not adding additional one :)

patzick (approved):
Thanks! 🙌 

That potentially could also be unified, I guess it was also about the default associations, but that's another case

PR #1574, opened 2025-01-10
fix: min-height display behaviour

### LINKED ISSUE #1573: [BUG] Min Height is being added to all displayModes

### Current Behavior

The current behaviour is, that the min-height css style of a CmsElementImage is being added every time.

### Expected Behavior

Only add the min-height, if the displayMode is set to "cover".

### Steps To Reproduce

1. Create a new layout and add a cms image element.
2. Add a picture
3. Load the site in the frontend app.

### Environment

```markdown
- OS: Windows 11 Version 10.0.26100 Build 26100
- Node: 22.10.0
- pnpm: 9.12.3
```

### Anything else?

_No response_

divide29:
### Description

closes shopware/shopware#6265 - fix the min-height display behaviour. Add the css property only if the images displayMode is set to cover.

### Type of change

### ToDo's

Nothing? :)

### DIFF
--- a/.changeset/smooth-llamas-lay.md
+++ b/.changeset/smooth-llamas-lay.md
@@ -1,0 +1,5 @@
+---
+"@shopware-pwa/composables-next": patch
+---
+
+fix: min-height display behaviour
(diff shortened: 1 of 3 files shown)


patzick:
I posted the question in the issue, let's figure it out there first :)

patzick:
Thank you for this contibution. Adding this to the lib would hit us back in future, better to fix it at the source - the API :)

Linking the workaround as this can be easily done in every frontends project using your code: https://github.com/shopware/shopware/issues/6265#issuecomment-2593187802

Let me know in case of any further problems! 🙌

Issue #1406, opened 2024-10-31
Deprecate old packages


patzick:
# **📚 Documentation Improvements**

## Description

Cleanup of the packages, get rid of the old `shopware-pwa` npm namespace

Example of a deprecated package: https://www.npmjs.com/package/@shopware-pwa/api-client
- info in readme
- points to the alternative solution
- deprecated on NPM

### Migrate packages
- [x] migrate CMS package https://www.npmjs.com/package/@shopware-pwa/cms-base to `@shopware/cms-base` and deprecate old package (#1599)
- [x] migrate nuxt3-module https://www.npmjs.com/package/@shopware-pwa/nuxt3-module to `@shopware/nuxt-module` (#1599)
- [x] migrate composables-next package https://www.npmjs.com/package/@shopware-pwa/composables-next to `@shopware/composables` (#1594, #1596 )
- [x] migrate helpers-next package https://www.npmjs.com/package/@shopware-pwa/helpers-next to `@shopware/helpers` (#1599)
- [x] remove api-client-old package from the repo (#1409)

### From the old PWA project
- [x] depcerate PWA client https://www.npmjs.com/package/@shopware-pwa/shopware-6-client 
- [x] depcerate PWA helpers https://www.npmjs.com/package/@shopware-pwa/helpers
- [x] deprecate PWA composables https://www.npmjs.com/package/@shopware-pwa/composables
- [x] deprecate PWA default theme https://www.npmjs.com/package/@shopware-pwa/default-theme
- [x] depcerate PWA CLI https://www.npmjs.com/package/@shopware-pwa/cli
- [x] deprecate PWA commons package with interfaces https://www.npmjs.com/package/@shopware-pwa/commons
- [x] deprecate PWA nuxt module https://www.npmjs.com/package/@shopware-pwa/nuxt-module
- [x] deprecate PWA theme-base package https://www.npmjs.com/package/@shopware-pwa/theme-base
- [x] deprecate PWA types package https://www.npmjs.com/package/@shopware-pwa/types - point to the new API client (#1413)

BrocksiNet:
Close to being done 👍

PR #1594, opened 2025-01-16
chore: deprecate composables-next package for @shopware/composables

patzick:
### Description

related #1406 

This PR marks package `@shopware-pwa/composables-next` as deprecated, further updates will be published under `@shopware/composables`


mkucmus (approved):
good bye my old friend

PR #1617, opened 2025-01-21
feat(vue-demo): UX for changing payment method - order details

mdanilowicz:
### Description

Added loader to the changing payment method on the order details page. 
Screenshot 2025-01-21 at 14 54 30

Added better UX for the change payment method modal
Screenshot 2025-01-21 at 14 54 28


patzick (changes requested):
The one thing I saw is that you also have layout shoft when payment methods are loading, maybe we could also add a loader there so content wouldn't jump so much?

image
image

mdanilowicz:
> The one thing I saw is that you also have layout shoft when payment methods are loading, maybe we could also add a loader there so content wouldn't jump so much?
> 
> image image

Done 

Screenshot 2025-01-27 at 13 13 02

patzick (approved):

PR #1641, opened 2025-01-24
feat(docs): add combined docs for paypal

cyl3x:
### Description

The current documentation on integrating PayPal only covers Express Checkout. Therefore, we would like to adapt the existing documentation to cover both a standard PayPal checkout and an express checkout.

### Type of change

Tweak the existing documentation, as most of it is generic for a PayPal checkout anyway, and add some dedicated sections for the express case.

### ToDo's

Currently, the different sections for standard and express checkout are duplicated. Perhaps there is a better way to structure this?


apps/docs/src/resources/integrations/payments/paypal.md
```diff
@@ -157,6 +190,42 @@ In the example above we do a couple of things:
...
 
+```ts
+  ...
+  // part of window.paypal.Buttons({}) params
+  onApprove: async (data: OnApproveData, actions: OnApproveActions) => {
+    // createOrder from useCheckout composable
+    orderCreated.value = await createOrder({
+      paypalOrderId: data.orderID,
+      // optional - can be acdc, applepay, googlepay or venmo
+      product: 'sbp',
```
mstegmeyer, line 201: `spb`, not `sbp` ;) 
Also, please document which product for which payment method, since it may be important, because different paymentSources get filled.

apps/docs/src/resources/integrations/payments/paypal.md
```diff
@@ -157,6 +190,42 @@ In the example above we do a couple of things:
...
+    });
+    refreshCart()
+    // apiInstance from useShopwareContext composable
+    const handlePaymentResponse = await apiInstance.invoke.post(
+      "/store-api/handle-payment",
+      {
+        orderId: orderCreated.value.id,
+        successUrl: `${window.location.origin}/order/success?order=${orderCreated.value.id}&success=true`,
+        errorUrl: `${window.location.origin}/order/success?order=${orderCreated.value.id}&success=false`,
```
mstegmeyer, line 210: doesn't that need the `paypalOrderId` to capture the created order? Same below in the existing example, I believe this was always an issue that this was not documented.

mstegmeyer (changes requested):
If you want to go the extra mile, you could also document other payment methods, such as ACDC, Google Pay, Apple Pay, PUI 🤔 or at least list the ones, for which the buttons are required and this tutorial works.

mkucmus (approved):
hey! great stuff 💙 

I fixed a dead link in docs and added link in one readme. 
thank you for this PR 🙌🏼

mstegmeyer (approved):

PR #1657, opened 2025-01-30
fix: add missing sw-language-id header

### LINKED ISSUE #1651: [BUG] Language context is broken for the multi language tabs
### Current Behavior

1. Open https://frontends-demo.vercel.app/
2. Open https://frontends-demo.vercel.app/pl-PL/ in the new tab
3. Go back to the first tab, and open any category page

Problem: 
- Content from the backend is in Polish

### Expected Behavior

Content from the backend should be in English

### Steps To Reproduce

_No response_

### Anything else?

_No response_

mdanilowicz:
### Description

This pull request includes several changes to add the missing `sw-language-id` header property for API calls across multiple components and modules. The most important changes are grouped by theme as follows:

### API Header Updates:

* Added `sw-language-id` header property for API calls in `useCart`, `useCategorySearch`, `useCheckout`, `useCustomerOrders`, `useInternationalization`, and `useLandingSearch` composables. [[1]](diffhunk://#diff-62200ee55ed613e59b385b7ec684bf93fc968a1b9e225cd5874c6f24ec7e73f3L118-R123) [[2]](diffhunk://#diff-59f00f31c4631dff545a876871f302a647d0f73c6f93e3d08b70a2e1f3ad0fcdR52) [[3]](diffhunk://#diff-2d2cc8b786b5e6cd149ee55e12ff8ae548f0c8d95fa3a5dff28e36b574c87067R113-R115) [[4]](diffhunk://#diff-688fbb7babbf724c12fccbb46483ef62455139404770fc465728998017688618R64-R66) [[5]](diffhunk://#diff-c4940faeb2dea6555863308a552c86e7bcc6ee300ccffc67c0d84c439f8e066fL97-R101) [[6]](diffhunk://#diff-a5b7fe35bf3598def1dc75b4ac047456de52dd386c03418f9a947f1ddcd537e3R50-R52)

### Import and Context Updates:

* Updated imports to include `useSessionContext` in various files to access `languageIdChain`. [[1]](diffhunk://#diff-62200ee55ed613e59b385b7ec684bf93fc968a1b9e225cd5874c6f24ec7e73f3L4-R4) [[2]](diffhunk://#diff-59f00f31c4631dff545a876871f302a647d0f73c6f93e3d08b70a2e1f3ad0fcdL1-R1) [[3]](diffhunk://#diff-2d2cc8b786b5e6cd149ee55e12ff8ae548f0c8d95fa3a5dff28e36b574c87067R83) [[4]](diffhunk://#diff-688fbb7babbf724c12fccbb46483ef62455139404770fc465728998017688618L3-R3) [[5]](diffhunk://#diff-c4940faeb2dea6555863308a552c86e7bcc6ee300ccffc67c0d84c439f8e066fL3-R3) [[6]](diffhunk://#diff-a5b7fe35bf3598def1dc75b4ac047456de52dd386c03418f9a947f1ddcd537e3L1-R1)

### Test Updates:

* Updated tests to include the `sw-language-id` header property for API call expectations in `useCart`, `useCategorySearch`, `useInternationalization`, and `useListing` components. [[1]](diffhunk://#diff-b2cd481f495a6152f73d8b6733e28479c474dad27287a5055c4965a9bc08f419R89-R93) [[2]](diffhunk://#diff-2c16d82d947b5558a3086a92322a8518dd15cef83767459c28f6f83d0baa2616R22) [[3]](diffhunk://#diff-749139724d0fbfee450a00b3413ba12108f5a95f26f1a9f416c72d61ea5153f7R32-R36) [[4]](diffhunk://#diff-b187eeee38d4dd2695b9a302c77c354e9d540371674878446c36114a4b7e9da2L24-R29)

### Component-Specific Changes:

* Added `sw-language-id` header property in `SwContactForm.vue` and `App.vue` components. [[1]](diffhunk://#diff-7c04909181931fa9b991085a41a2977fb37b7092dfa8bfc67e8a9f5d11f700c2R153-R155) [[2]](diffhunk://#diff-12260e5806d489206db5e938f4e1027c4ac4362d4f7f6e3a8957cf3017e9a88cR112-R114)

closes #1651


mdanilowicz:
> why does API ignore the current session language set by `PATCH /context` and we need to pass it explicitly?

It's not ignoring. The second tab request overrides this session language by a `PATCH /context` request, and when you get back to the first tab, you have language from the second tab. That is why we have to use this header

mkucmus (changes requested):
looks like a lot of work done down here 🔥  ⛑️ 🏅 

TODO:
- fix breadcrumbs for navigation because it still mixes the languages (when I refresh the page on the different language in a second tab - breadcrumbs at the first tab is in a wrong language)

TO CONSIDER:
- as `languageIdChain` is broadly used now, it still tells nothing and the name might be confusing for developers, maybe we can alias it to something more clear and fancy?
- maybe we can hook-in the apiClient instance and set `sw-language-id` globally for entire app?

mdanilowicz:
> * fix breadcrumbs for navigation because it still mixes the languages (when I refresh the page on the different language in a second tab - breadcrumbs at the first tab is in a wrong language)

I'm not able to reproduce the problem with breadcrumbs. Are you sure that this is not a problem with the missing translations? Please add steps to reproduce

> * as `languageIdChain` is broadly used now, it still tells nothing and the name might be confusing for developers, maybe we can alias it to something more clear and fancy?

sure, why not

> * maybe we can hook-in the apiClient instance and set `sw-language-id` globally for entire app?

Not all endpoints consume `sw-language-id`. So, adding it globally can be misleading

mkucmus:
> > * fix breadcrumbs for navigation because it still mixes the languages (when I refresh the page on the different language in a second tab - breadcrumbs at the first tab is in a wrong language)
> 
> I'm not able to reproduce the problem with breadcrumbs. Are you sure that this is not a problem with the missing translations? Please add steps to reproduce
> 
- go to https://frontends-demo-git-fix-gh-1651-shopware-frontends.vercel.app/pl-PL/Wakacyjne-Trendy/
- the breadcrumbs is in a correct language (coming from API I believe)
- in the second tab go to, let's say https://frontends-demo-git-fix-gh-1651-shopware-frontends.vercel.app/en-GB/Summer-BBQ/
- in the first tab click around the shop to some translated page, like Products and then back to Wakacyjne-Trendy.
- the creadcrumbs are in english now.

> > * as `languageIdChain` is broadly used now, it still tells nothing and the name might be confusing for developers, maybe we can alias it to something more clear and fancy?
> 
> sure, why not
💪🏼 
> 
> > * maybe we can hook-in the apiClient instance and set `sw-language-id` globally for entire app?
> 
> Not all endpoints consume `sw-language-id`. So, adding it globally can be misleading

you are adding it globally but behind each composable function's logic. 
If API doesn't consume the header, it will be simply ignored, right? so it would be just a code lines saver. 

If the context has its language set, the client could be aware of it in a global state to avoid that some requests won't be in a right language - within the whole composables package. @patzick is that improvement or shadowing the whole thing with translations and current context? or it's just not possible though 😆

mdanilowicz:
Ok, I see, fixed and added a task for backend https://github.com/shopware/shopware/issues/6630

@mkucmus, but please remember that header is not the only way to set the language. We also use context to do it. I'm afraid that we can overcomplicate it, but I'm not totally against global header :D

mkucmus (approved):

mkucmus:
> Ok, I see, fixed and added a task for backend [shopware/shopware#6630](https://github.com/shopware/shopware/issues/6630)
> 
> @mkucmus, but please remember that header is not the only way to set the language. We also use context to do it. I'm afraid that we can overcomplicate it, but I'm not totally against global header :D

yup, it's already overcomplicated because we have dynamic content, reactive, bound to the Id and we synchronize the session between browser tabs - that leads to the problem we encounter. 

Let's do the rework of that in the future, starting from the backend to ensure that the global sw-language-id header can be applied for every endpoint.

BrocksiNet:
> Let's do the rework of that in the future, starting from the backend to ensure that the global sw-language-id header can be applied for every endpoint.

This should be collected and driven by the API working group. :100:

Issue #1652, opened 2025-01-28
handle listing query params for `isr` rendering


mkucmus:
### Describe the feature

utilize configuration described here https://github.com/nitrojs/nitro/pull/2780 to make vercel (or other providers serving `isr`) aware of dynamic routes that can also have a dynamic query params. 

currently the whole entrypoint for a specific route is cached regardless the query params provided in URL.

### Description

--

### Additional information

- [x] Would you be willing to help implement this feature?

### Checks

- [x] Check existing [discussions](https://github.com/shopware/frontends/discussions) and [issues](https://github.com/shopware/frontends/issues).

mkucmus:
the issue was reported here: https://github.com/nitrojs/nitro/issues/1880#issuecomment-2639226324

mkucmus:
fixed `nitropack` lib will be available in a next release of nuxt (probably upcoming `4.2.3`), so we could close this issue then. currently the fixed release of nitropack [`2.13.0` ](https://github.com/nitrojs/nitro/releases/tag/v2.13.0) is in a [main](https://github.com/nuxt/nuxt/blob/main/package.json#L126) branch

mkucmus:
nuxt@4.4.6 ships nitropack@2.13.4 which contains the fix. All templates (vue-demo-store, vue-starter-template, vue-starter-template-extended, vue-blank) are already on this version. closing.

PR #1674, opened 2025-02-06
feat(vue-demo-store): disable isr for catch all route

mkucmus:
### Description

due to issue #1652 we have to disable the isr for default vercel rendering mode.

### DIFF
--- a/templates/vue-demo-store/nuxt.config.ts
+++ b/templates/vue-demo-store/nuxt.config.ts
@@ -82,7 +82,7 @@ export default defineNuxtConfig({
       ssr: false,
     },
     "/**": {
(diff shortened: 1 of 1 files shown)


mdanilowicz (approved):

patzick:
It's fixing only listing pages with filtering, but removes a very good caching option overall

what about reloading it on CSR when query params are present?

for comparision:
- https://frontends-demo-git-feat-disable-isr-f-7f52d6-shopware-frontends.vercel.app/Specials-Sale/
- https://frontends-demo.vercel.app/Specials-Sale/

page is loading over 10x slower without ISR

PR #1698, opened 2025-02-14
feat: programmatic API for api-gen commands

freb97:
### Description

When using the api-gen package, the only possible way currently is to use it via cli. Would it be okay to externalize the `generate`, `loadSchema` and `validateJson` commands to allow users to integrate it into their code?

The point of this would be to allow generating the types at build time with a nuxt hook or similar, but maybe there are other cases where the methods would be useful.

### Type of change

### ToDo's

### Additional context

None


patzick:
that is a very good idea! I'll change this PR status to draft until it's finished but I like it 🙌

freb97:
@patzick I updated the pull request and tested the build with npm link locally. I think the import statements look a lot cleaner now and also the functions are typed. Also added some (very rudimentary) documentation

patzick (approved):
Great stuff 🙌 

Let us know how you're gonna use it in your project!

PR #1789, opened 2025-03-17
feat: add multi domain example

mdanilowicz:
### Description

Added example of how to configure multi-domain approach


package.json
```diff
@@ -108,9 +108,6 @@
         "CVE-2024-31207",
         "CVE-2024-43788"
       ]
-    },
-    "patchedDependencies": {
-      "@vueuse/core@11.1.0": "patches/@vueuse__core@11.1.0.patch"
```
patzick, line 113: patch files should also be removed

mdanilowicz: fixed

patzick (approved):

PR #1801, opened 2025-03-26
fix: undefined content type in headers multipart/form-data

joostaasman:
### Description

After setting the content type in the api client headers like this:
```
headers: {
    "Content-Type": "multipart/form-data", // <-- set this one
},
```
the actual content type is set to undefined. This fix ensures that the content type is removed in the API client and correctly set by the browser.


patzick (approved):
Thanks, I adjusted tests a bit, so it's clear now what headers are sent

PR #1812, opened 2025-04-01
fix: date formatting

patzick:
### Description
Fix date formatting based on proper locales


packages/cms-base-layer/components/SwProductReviews.vue
```diff
@@ -12,12 +16,14 @@ const props = defineProps<{
...
   product: {
     noReviews: string;
+    reviewNotAccepted: string;
   };
 };
 
 let translations: Translations = {
   product: {
     noReviews: "No reviews yet.",
+    reviewNotAccepted: "Your review has not been approved yet",
```
mdanilowicz, line 26: Translations should also be added to the translation files

templates/vue-demo-store/pages/checkout/success/[id]/index.vue
```diff
@@ -63,10 +63,20 @@ const toggleView = () => {
...
-  new Date(date).toLocaleDateString(
-    (typeof navigator !== "undefined" && navigator.language) || "en-US",
-  );
+const { sessionContext } = useSessionContext();
+const navLang =
+  sessionContext.value?.languageInfo.localeCode ||
+  navigator?.language ||
+  "en-US";
+
+const formatDate = (date: Date | number) =>
```
Comment (mdanilowicz) [inline templates/vue-demo-store/pages/checkout/success/[id]/index.vue:72]: Why not as a helper function?

packages/cms-base-layer/components/SwProductReviews.vue
```diff
@@ -42,15 +48,21 @@ const reviewsList = computed<Schemas["ProductReview"][]>(
...
   day: "numeric",
   hour: "numeric",
   minute: "numeric",
-  hour12: true,
 };
 
-const formatDate = (date: string) =>
-  new Date(date).toLocaleDateString("en-us", format);
+const { sessionContext } = useSessionContext();
+const navLang =
```
mdanilowicz, line 58: localeCode from the session is not the best source. I saw that in many examples, `en` code was used for `en-GB` and `en-US`

mdanilowicz (changes requested):

packages/cms-base-layer/components/SwProductReviews.vue
patzick: this is moved from override component from template, not sure why it was created there, but cms-base should handle it. 
Translations are already there in files

packages/cms-base-layer/components/SwProductReviews.vue
patzick: do you mean session? This is preventing from hardcoded locales, session now contains new `languageInfo. localeCode` which exactly indicates the locale. What is your proposition here?

templates/vue-demo-store/pages/checkout/success/[id]/index.vue
Comment (patzick) [reply templates/vue-demo-store/pages/checkout/success/[id]/index.vue:72]: as format is the biggest part of it, you can see exactly what will be displayed (here with hours and minutes for example). I'm okay with helper function for that but that was not the case for this fix

mkucmus (approved):

mdanilowicz (approved):

PR #1814, opened 2025-04-02
fix(api-gen): merge json overrides

### LINKED ISSUE #1746: [BUG] @shopware/api-gen not behaving as expected when applying multiple patches.
### Current Behavior

The [@shopware/api-gen](https://www.npmjs.com/package/@shopware/api-gen) package doesn't work when two patches affect the same schema.

### Expected Behavior

I found the patch linked by the npm package here:
https://raw.githubusercontent.com/shopware/frontends/main/packages/api-client/api-types/storeApiSchema.overrides.json

But I also want to customize the `customFields` property to our needs. You can see a minimal example in this repository
(`storeApiTypes.overrides.json`). Since both apply changes to the `Product` schema, only the first one (in this case the
shopware default patch) is applied.

Simply checkout this repository and run

```bash
pnpm api:generate
```

The result is the applied patch including the one for `calculatedCheapestPrice`, but not the patch for `customFields`.
Swap the order of both patches in `api-gen.config.json` and the result will change. Now the patch for `customFields` is
applied but not the patch for `calculatedCheapestPrice`.

---

I really don't understand the underlying mechanic as I expected it to apply patches by simply deepmerging all patches
after each other into the `storeApiSchema.json` and then generating the `storeApiTypes.d.ts` based on the openAPI specs.

The documentation of `@shopware/api-gen` states the following:

> Partial overrides are applied directly to the JSON schema, so the syntax needs to be correct

Why do I have to use a different schema than the `storeApiSchema.json`?
The structure there is:

```json
{
  "components": {
    "schemas": {
      "Product": {}
    }
  }
}
```

whereas the `storeApiTypes.overrides.json` requires the following structure:

```json
{
  "components": {
    "Products": {}
  }
}
```

If I use the same structure as originally defined, it doesn't work at all. 

### Steps To Reproduce

1. Checkout my reproduction repository: https://github.com/max-zinn/api-gen-patch-bug
2. read the README.md, it should give enough explanation on the issue.

### Anything else?

_No response_

patzick:
### Description
closes #1746 

related to #1755 which still can be merged after slight adjustmets

Fixing problem with multiple overrides, when one of them is using object as types. In further work we should encourage only arrays

### DIFF
--- a/.changeset/khaki-ideas-clean.md
+++ b/.changeset/khaki-ideas-clean.md
@@ -1,0 +1,5 @@
+---
+"@shopware/api-gen": patch
+---
+
+Merging JSON overrides allows to have multiple pverrides as objects and arrays - combining them into arrays.
--- a/examples/b2b-employee-management/api-types/storeApiTypes.d.ts
+++ b/examples/b2b-employee-management/api-types/storeApiTypes.d.ts
@@ -11118,7 +11118,6 @@ export type operations = {
   "readRoles get /role": {
     contentType?: "application/json";
     accept?: "application/json";
-    body?: components["schemas"]["Criteria"];
     response: {
       elements?: components["schemas"]["B2bComponentsRole"][];
     } & components["schemas"]["EntitySearchResult"];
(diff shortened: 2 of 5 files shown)


mkucmus (approved):
image

PR #1808, opened 2025-03-28
feat(docs): add tag for dynamically generated content

mkucmus:
### Description

closes: #1610 

from now on, the pages containing dynamically generated content will have:

```html
<div data-placeholder="dynamic-markdown"></div>
```

the `<!-- dynamic-markdown -->` is stripped out during minification process

PR #1817, opened 2025-04-03
docs: remove dynamic markdown placeholder from non-transformed code

bojanrajh:
### Description

This PR removes markdown placeholder from non-transformed pages. Other plugins work fine, see https://github.com/shopware/frontends/pull/1808 .


mkucmus (approved):
@bojanrajh are you sure this is the only place to remove it? the `div` with a placeholder is being used in other places as well.

PR #1833, opened 2025-04-13
feat: init starter template

mdanilowicz:
### Description

Init vue starter template


templates/vue-starter-template/uno.config.ts
```diff
@@ -0,0 +1,112 @@
...
+  defineConfig,
+  presetAttributify,
+  presetIcons,
+  presetTypography,
+  presetWind3,
+} from "unocss";
+
+export default defineConfig({
+  theme: {
+    colors: {
```
mkucmus, line 11: I think we are ready to take the config from https://github.com/shopware/frontends/blob/nuxt-layers/packages/layers/ui-light-theme/uno.config.ts#L9 as it's been resolved for hex values and grouped for other types than colors

PR #1843, opened 2025-04-22
feat: init CMS routing for starter template

mdanilowicz:
### Description

Add CMS component resolving for starter template


templates/vue-starter-template/app/app.vue
```diff
@@ -1,3 +1,14 @@
+<script setup lang="ts">
+import type { Schemas } from "#shopware";
+
+const { apiClient } = useShopwareContext();
+const sessionContextData = ref<Schemas["SalesChannelContext"]>();
+const contextResponse = await apiClient.invoke("readContext get /context");
```
patzick, line 6: do we want to stop rendering for context fetching?

1. it should probably be CSR-only action
2. would be nice to make it more asynchronous

if 2. does not make sense then maybe it should be in asyncData then?

templates/vue-starter-template/app/components/FrontendAccountCustomerGroupRegistrationPage.vue
```diff
@@ -0,0 +1,47 @@
...
+  },
+);
+</script>
+
+<template>
+  <div class="container mx-auto bg-white flex flex-col">
+    <h1
+      class="mb-4 text-2xl font-extrabold leading-none tracking-tight text-secondary-900 md:text-3xl lg:text-5xl dark:text-white text-center"
+    >
+      {{ registrationResponse?.translated.registrationTitle }}
```
patzick, line 27: we should not have optional chaining here - either we do have the response or not, if not - error page

templates/vue-starter-template/app/components/FrontendDetailPage.vue
```diff
@@ -0,0 +1,87 @@
...
+        pathParams: {
+          id: props.navigationId,
+        },
+      }),
+    ]);
+
+    for (const response of responses) {
+      if (response.status === "rejected") {
+        console.error("[FrontendDetailPage.vue]", response.reason.message);
+        errorDetails.value = response.reason.message;
```
patzick, line 33: we display a single error, maybe concat them?

templates/vue-starter-template/app/components/FrontendDetailPage.vue
```diff
@@ -0,0 +1,87 @@
...
+      }
+    }
+
+    return {
+      productResponse:
+        responses[0].status === "fulfilled" ? responses[0].value : null,
```
patzick, line 39: why do we need `null`? I'd go with the undefined so
```suggestion
      productResponse: responses[0].value,
```

templates/vue-starter-template/app/components/FrontendDetailPage.vue
```diff
@@ -0,0 +1,87 @@
...
+      breadcrumbs:
+        responses[1].status === "fulfilled" ? responses[1].value : null,
+    };
+  },
+);
+const productResponse = ref(data.value?.productResponse);
```
patzick, line 45: `data.value` is never null here, we do not need optional chaining for it


additionally we do not replace this value, I would avoid rewriting that to separate ref, just use it everywhere by the reference, or even just put it into the const

```suggestion
const productResponse = data.value.productResponse;
```

templates/vue-starter-template/app/components/FrontendLandingPage.vue
```diff
@@ -0,0 +1,42 @@
...
+
+if (!landingResponse?.value) {
+  console.error("[FrontendLandingPage.vue]", error.value?.message);
+  throw createError({
+    statusCode: 500,
+    message: error.value?.message,
+  });
+}
+
+const landingPage = landingResponse as Ref<Schemas["LandingPage"]>;
```
patzick, line 35: can we type it while fetching, so we don't need to cast it here?

templates/vue-starter-template/app/components/FrontendLandingPage.vue
```diff
@@ -0,0 +1,42 @@
...
+useCmsHead(landingPage, { mainShopTitle: "Shopware Frontends Demo Store" });
+</script>
+
+<template>
+  <LayoutBreadcrumbs />
+  <CmsPage v-if="landingResponse?.cmsPage" :content="landingResponse.cmsPage" />
```
patzick, line 41: what should happen if the cmsPage is not here?

templates/vue-starter-template/app/components/FrontendNavigationPage.vue
```diff
@@ -0,0 +1,74 @@
...
+        })
+        .catch(() => {
+          console.error("Error while fetching breadcrumbs");
+        }),
+    ]);
+
+    for (const response of responses) {
+      if (response.status === "rejected") {
+        console.error("[FrontendNavigationPage.vue]", response.reason.message);
+        errorDetails.value = response.reason.message;
```
patzick, line 41: same with displaying only a single error

also with returning values etc

templates/vue-starter-template/app/composables/useCmsHead.ts
```diff
@@ -0,0 +1,82 @@
...
+  getCategoryImageUrl,
+  getSmallestThumbnailUrl,
+  isLandingPage,
+  isProduct,
+} from "@shopware/helpers";
+import type { Schemas } from "#shopware";
+
+export type UseCmsHeadReturn = ReturnType<typeof useCmsHead>;
+
+type CmsPageEntity =
```
patzick, line 11: why not use Schemas["CmsPage"] ?

templates/vue-starter-template/app/composables/useCmsHead.ts
```diff
@@ -0,0 +1,82 @@
...
+  name: string;
+  content: string;
+};
+export function useCmsHead(
+  entity: Ref<CmsPageEntity> | ComputedRef<CmsPageEntity>,
+  options?: {
+    mainShopTitle?: string;
+  },
+): void {
+  const unrefEntity = unref(entity) as CmsPageEntity;
```
patzick, line 26: unref should return the same type without casting

templates/vue-starter-template/nuxt.config.ts
```diff
@@ -1,13 +1,14 @@
 // https://nuxt.com/docs/api/configuration/nuxt-config
 export default defineNuxtConfig({
-  extends: ["@shopware/composables/nuxt-layer"],
+  ssr: false,
```
patzick, line 3: why?

templates/vue-starter-template/app/components/FrontendNavigationPage.vue
```diff
@@ -0,0 +1,74 @@
...
+      }),
+      apiClient
+        .invoke("readBreadcrumb get /breadcrumb/{id}", {
+          pathParams: {
+            id: props.navigationId,
+          },
+        })
+        .catch(() => {
+          console.error("Error while fetching breadcrumbs");
+        }),
```
patzick, line 35: let's not do this in allSettled, let's catch errors after the resolve

patzick (changes requested):
Overall very good, I left few comments

- possibly avoid casting to make sure we already have good typing provided
- let's make api more reliable so we don't need to use optional chaining in template (cms is fine, it consider more versions)

templates/vue-starter-template/app/app.vue
mdanilowicz: I don't know the context of the change here, https://github.com/shopware/frontends/pull/309/files#diff-4359a385df4e4b0363f4e54f8cc7240a106ecfcf1be47ec09b91f7537990805fL20, but it was `asycData` before, and changed by you.  

Moving request to CSR will break the app because the session context will  not be set on the SSR

patzick: yeah, the reason was to not share session token from server-cached render, that's why my first suggestion was to make it CSR-only

we should break down what exactly will break and maybe fix it? Would be nice to have server render session-independent. Additionally not to wait with CSR render/hydration with fetching session, but rather async load it

mdanilowicz: I created separated task for this https://github.com/shopware/frontends/issues/1844

templates/vue-starter-template/app/components/FrontendAccountCustomerGroupRegistrationPage.vue
mdanilowicz: If you check the code below, it is not like that. no response =/= error page. I created a task for that. Component is moved from demo template

https://github.com/shopware/frontends/issues/1845

templates/vue-starter-template/nuxt.config.ts
mdanilowicz: for test, removed

templates/vue-starter-template/app/composables/useCmsHead.ts
mdanilowicz: https://github.com/shopware/frontends/pull/987/files#diff-810b2e943f5c2779277b977880c39cd7d94df321439f7645839a466e7047c8fa

Types do not overlap

templates/vue-starter-template/app/components/FrontendDetailPage.vue
mdanilowicz: This will lead to a type error. https://stackoverflow.com/questions/63783735/type-error-on-response-of-promise-allsettled <- context

templates/vue-starter-template/app/components/FrontendDetailPage.vue
mdanilowicz: done

templates/vue-starter-template/app/components/FrontendDetailPage.vue
mdanilowicz: done

templates/vue-starter-template/app/components/FrontendNavigationPage.vue
mdanilowicz: done

templates/vue-starter-template/app/components/FrontendNavigationPage.vue
mdanilowicz: done

templates/vue-starter-template/app/components/FrontendLandingPage.vue
mdanilowicz: https://github.com/shopware/frontends/issues/1846

templates/vue-starter-template/app/components/FrontendLandingPage.vue
mdanilowicz: https://github.com/shopware/frontends/issues/1846

templates/vue-starter-template/app/components/FrontendNavigationPage.vue
patzick: I still see the "catch" here

templates/vue-starter-template/app/composables/useCmsHead.ts
patzick: so what should we do to fix it? What is needed on the API declarations to make it correct?

mdanilowicz: We cannot replace `CmsPageEntity` with the `CmsPage` type because it is not a union for `Category`, 
  `LandingPage` and `Product` types

patzick (approved):

PR #1859, opened 2025-05-05
fix: account missing attributes for icons

vijith2hatslogic:
### Description

The class attributes are missing for account page navigation items in the left side

### Type of change

Adding class attribute for making the icons visible in front-end


patzick (approved):
in most cases attributify should cover it: https://unocss.dev/presets/attributify

but no harm in changing it 🙌  thanks

PR #1862, opened 2025-05-06
fix: remove category related code from CmsSectionSidebar

aheartforspinach:
### Description
This pr fixes the CmsSectionSidebar.vue when it's used on a landing page by removing useCategory and related code (otherwise we receive an error in the useCategory composable because there is no category available on a landing page)

### Type of change

### ToDo's

- [x] Changeset file provided


patzick (approved):
Good point, thank you 🙌

Issue #1870, opened 2025-05-16
[BUG] Reset password hash does not work


digitalwert-devs:
### Current Behavior

If you use the password reset function under /account/recover, you will receive a reset link that can be set in the admin dashboard under “Settings” - “Users & Registration”. 
Password recovery URL looks like this:
/account/recover/password?hash=%%RECOVERHASH%%

The link throws a 404 error in the frontend after using it. If you change it to the following format:
/account/recover?hash=%%%RECOVERHASH%% you will be redirected to the e-mail input of the recovery page again and again

Translated with DeepL.com (free version)

### Expected Behavior

After clicking on the link, you should be redirected to a page where you can create a new password.

### Steps To Reproduce

1. visit URL/account/recover
2. enter e-mail address
3. use link in email

mkucmus:
Actually there is a ready-to-use page route for that purpose: /reset-password page and that should be pointed in the settings: 

![Image](https://github.com/user-attachments/assets/533927b8-b9de-45a9-9110-fd6d3884439d)

however,

TODO:
- [ ] there is no success message
- [ ] no docs about it

the task will track the enhancement process of the entire flow.

PR #1905, opened 2025-06-24
feat(vue-demo-store): set the default route for recovery password and…

mkucmus:
closes: #1870


mkucmus:
hey @kstala. please check whether the flow for password recovery really works when the route path is set to default (reset to the default value is important).

kstala (approved):

Issue #1876, opened 2025-05-27
enable to use cms elements with default values for styling


mkucmus:
experiment to show that we can simply migrate an end project to use elements styled in a new way (based on new template theme) without breaking changes.

acceptance criteria:
- use the same cms element in both: old and new template
- element must contain default styling for custom tailwind css tokens 
- new template must display the element properly
- readme or migration guide is updated

PR #1907, opened 2025-06-27
feat: enable `nuxtLayers` functionality in uno config

mkucmus:
### Description

closes: #1876 

### Type of change

-- New feature (non-breaking change which adds functionality)


packages/cms-base-layer/uno.config.ts
```diff
@@ -0,0 +1,40 @@
...
+      Noto_Serif: "Noto Serif",
+    },
+  },
+  safelist: ["states-success", "states-error", "states-info", "states-warning"],
+  presets: [
+    presetWind3(),
+    presetIcons({
+      collections: {
+        carbon: () =>
+          import("@iconify-json/carbon/icons.json").then((i) => i.default),
```
patzick, line 39: we should have this as dependency in package

PR #1914, opened 2025-07-08
feat: setting payment and shipping on the checkout

mdanilowicz:

### DIFF
--- a/templates/vue-starter-template/app/components/form/BaseInput.vue
+++ b/templates/vue-starter-template/app/components/form/BaseInput.vue
@@ -1,7 +1,16 @@
 <script lang="ts" setup>
+type InputTypeAttribute =
+  | "email"
+  | "number"
+  | "password"
+  | "tel"
+  | "text"
+  | "time"
+  | "url";
+
 const { placeholder, type = "text" } = defineProps<{
   placeholder?: string;
-  type?: "text" | "password";
+  type?: InputTypeAttribute;
   invalid?: boolean;
 }>();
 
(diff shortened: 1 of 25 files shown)


templates/vue-starter-template/app/components/form/BaseInput.vue
```diff
@@ -1,6 +1,8 @@
 <script lang="ts" setup>
-defineProps<{
+const { placeholder, type = "text" } = defineProps<{
   placeholder?: string;
+  type?: "text" | "password";
```
patzick: can't we be more open to all types of inputs?
maybe something like HTMLInputElement['type'] or something similar for autocompletion?

templates/vue-starter-template/package.json
```diff
@@ -29,7 +29,8 @@
     "@shopware/helpers": "canary",
     "@shopware/nuxt-module": "canary",
     "nuxt": "3.17.5",
-    "vue": "3.5.16"
+    "vue": "3.5.16",
+    "@vuelidate/core": "2.0.3"
```
patzick: this should go higher to match alphabetical order

mdanilowicz: done

templates/vue-starter-template/app/components/form/BaseInput.vue
mdanilowicz: We cannot use "global type" because this component should not handle all types, like the `color`, `calendar`. I added additional types.

patzick (approved):

PR #1985, opened 2025-08-21
feat(cms-base-layer): product card component

mkucmus:
### Description

closes: #1782 

### Type of change

### DIFF
--- a/packages/cms-base-layer/app/components/SwProductCard.vue
+++ b/packages/cms-base-layer/app/components/SwProductCard.vue
@@ -99,7 +99,7 @@ const toggleWishlistProduct = async () => {
       );
     } else {
       await removeFromWishlist();
-      pushError(
+      pushSuccess(
         `${props.product?.translated.name} ${translations.product.removedFromTheWishlist}`,
       );
     }


packages/cms-base-layer/app/components/public/cms/element/CmsElementProductListing.vue
```diff
@@ -154,59 +154,26 @@ compareRouteQueryWithInitialListing();
...
-        >
-          <SwProductCard
-            v-for="product in getElements"
-            :key="product.id"
-            :product="product"
-            :is-product-listing="isProductListing"
-            class="p-4 border rounded-lg shadow-md hover:shadow-lg transition-shadow duration-200 ease-in-out w-full lg:w-3/7 2xl:w-7/24 mr-0 sm:mr-8 mb-8"
-          />
+        <div v-if="!loading" ref="productListElement"
+          class="flex justify-center flex-wrap p-4 md:p-6 lg:p-8 productListElement">
```
mdanilowicz: Why do we need `productListElement` class?

packages/cms-base-layer/app/components/SwProductCard.vue
```diff
@@ -153,134 +153,127 @@ const srcPath = computed(() => {
...
+const formattedPrice = computed(() => {
+  const price = product.value?.calculatedPrice?.unitPrice || 0;
+  return `${price.toFixed(2).replace(".", ",")} €`;
+});
+
+const formattedListPrice = computed(() => {
+  if (!isOnSale.value) return null;
+  const listPrice = product.value?.calculatedPrice?.listPrice?.price || 0;
+  return `${listPrice.toFixed(2).replace(".", ",")} €`;
+});
```
mdanilowicz: Maybe we can move it to separate helpers/utils?

packages/cms-base-layer/app/components/SwProductCard.vue
```diff
@@ -153,134 +153,127 @@ const srcPath = computed(() => {
...
+    <div class="flex flex-col items-start justify-start gap-4 self-stretch p-2">
+      <!-- Manufacturer and product name -->
+      <div class="flex flex-col items-start justify-start gap-2 self-stretch">
+        <div class="flex flex-col items-start justify-start gap-1 self-stretch">
+          <div v-if="productManufacturer"
+            class="text-surface-on-surface justify-start self-stretch font-['Inter'] text-sm font-bold leading-tight">
+            {{ productManufacturer }}
+          </div>
+          <RouterLink :to="buildUrlPrefix(getProductRoute(product), getUrlPrefix())"
+            class="text-surface-on-surface justify-start self-stretch font-['Noto_Serif'] text-2xl font-normal leading-9 h-[4.5rem] overflow-hidden line-clamp-2 display-webkit-box"
```
mdanilowicz: `h-[4.5rem]` it should work with `h-[4.5]` rem is a default unit

packages/cms-base-layer/app/components/SwProductCard.vue
```diff
@@ -153,134 +153,127 @@ const srcPath = computed(() => {
...
-          :disabled="!product.available"
-          @click="addToCartProxy"
-        >
-          {{ translations.product.addToCart }}
-          <div v-if="isInCart" class="flex ml-2">
-            <div class="w-5 h-5 i-carbon-shopping-bag text-gray-600" />
-            {{ count }}
+      </div>
+      <!-- Price section -->
+      <div :data-sale="isOnSale ? 'yes' : 'no'" class="inline-flex items-center justify-start gap-2">
```
mdanilowicz: Do we need translation here?

packages/cms-base-layer/app/components/SwProductCard.vue
```diff
@@ -153,134 +153,127 @@ const srcPath = computed(() => {
...
           </div>
-        </RouterLink>
+        </div>
+        <div v-else class="text-surface-on-surface justify-start font-['Inter'] text-base font-bold leading-normal">
+          {{ formattedPrice }}
+        </div>
       </div>
+
+      <!-- Color options -->
+      <!-- <div v-if="colorOptions.length > 0" class="inline-flex w-28 flex-wrap content-center items-center justify-start gap-2">
```
mdanilowicz: commented = no needed :D ?

packages/cms-base-layer/app/components/SwProductCard.vue
```diff
@@ -153,134 +153,127 @@ const srcPath = computed(() => {
...
+          ></div>
+        </div>
+      </div> -->
+      <SwButton variant="primary" v-if="!fromPrice" size="medium" :disabled="!product?.available" block
+        data-testid="add-to-cart-button" @click="addToCartProxy">
+        {{ translations.product.addToCart }}
+      </SwButton>
+      <!-- Details button for products with fromPrice -->
+      <RouterLink v-else :to="buildUrlPrefix(getProductRoute(product), getUrlPrefix())" class="self-stretch">
+        <SwButton block>
```
mdanilowicz: API for the template and CMS components should be similar. Or same component?

mdanilowicz (changes requested):

packages/cms-base-layer/app/components/SwProductCard.vue
mkucmus: 👍🏼 should be as similar as possible, but the template extends from cms so it's easy to get the details on how it should look like or what props and events it handles. the naming is still to consider, I mean the prefix if it's really neaded, but would be good to avoid naming collision as Button or BaseButton can be available in other ui libs...

packages/cms-base-layer/app/components/form/IconButton.vue
mdanilowicz, line 1: This should not be moved. You should be able to use the template without CMS.

packages/cms-base-layer/app/components/public/cms/element/CmsElementProductListing.vue
mdanilowicz: still current question :D

templates/vue-starter-template/app/components/layout/header/CartIcon.vue
```diff
@@ -10,7 +10,7 @@ defineProps<{
       name="shopware:shopping-cart"
       class="color-brand-primary"
     />
-    <LayoutMainCounter
```
mdanilowicz: The header elements should not be connected to the CMS. After disabling/removing CMS template will break

templates/vue-starter-template/app/components/layout/header/WishlistIcon.vue
mdanilowicz, line 1: It should stay in the template (wishlist page). There should be the ability to use it without CMS, and maybe CMS should try to load it from the template. If does't exist maybe fallback to own copy?

templates/vue-starter-template/app/components/layout/Header.vue
```diff
@@ -37,10 +37,10 @@ function handleMyAccountClick() {
             @click="toggleMobileSearch"
             class="hidden max-sm:block"
           />
-          <FormIconButton type="ghost" @click="handleMyAccountClick"
```
mdanilowicz: Should stay, same arguments as above

packages/cms-base-layer/app/components/SwProductCard.vue
```diff
@@ -148,139 +148,81 @@ function roundUp(num: number) {
...
+
+// Computed properties for display data
+const productName = computed(() => getProductName({ product: product.value }));
+const productManufacturer = computed(
+  () => product.value?.manufacturer?.translated?.name || "",
+);
+const isOnSale = computed(
+  () => (product.value?.calculatedPrice?.listPrice?.percentage ?? 0) > 0,
+);
+const isTopseller = computed(() => product.value?.markAsTopseller);
```
mdanilowicz: It should be moved to utils/helpers

templates/vue-starter-template/app/components/layout/Header.vue
mkucmus: I would propose the same idea as for uno config, if the component exist in a template - leave it in a template, but exact the same fallback component must be put in the cms. the CMS also need to live without a template

templates/vue-starter-template/app/components/layout/header/CartIcon.vue
mkucmus: I will do a copy of that, as main counter is used in cms as well according to https://github.com/shopware/frontends/pull/1985#discussion_r2321470815

templates/vue-starter-template/app/components/layout/header/WishlistIcon.vue
mkucmus: I will leave a copy of that

templates/vue-starter-template/app/components/layout/header/WishlistIcon.vue
```diff
@@ -1,13 +1,19 @@
 <script setup lang="ts">
 defineProps<{
-  counter: number;
+  filled?: boolean;
```
mdanilowicz: Please add default value for filled props

templates/vue-starter-template/app/components/layout/header/WishlistIcon.vue
```diff
@@ -1,13 +1,19 @@
...
+  counter?: number;
```
mdanilowicz: counter should be a required value

templates/vue-starter-template/app/components/layout/header/WishlistIcon.vue
```diff
@@ -1,13 +1,19 @@
...
-    <Icon size="1.2rem" name="shopware:heart" class="color-brand-primary" />
+    <!-- use when filled icon is ready -->
+    <!-- <Icon size="1rem" name="shopware:heart" v-if="type !== 'filled'" class="w-6 h-5 block hover:cursor-pointer" :class="[
+      styles[type]
+    ]" /> -->
+    <div class="i-carbon-favorite w-6 h-5 hover:cursor-pointer" v-if="!filled"></div>
+    <div class="i-carbon-favorite-filled w-6 h-5 hover:cursor-pointer" v-else></div>
     <LayoutMainCounter
-      v-if="counter > 0"
+      v-if="(counter ?? 0) > 0"
```
mdanilowicz: This should be handled by the parent component.  The less logic, the better. The API of this component should be easy and logical

templates/vue-starter-template/app/components/layout/header/WishlistIcon.vue
```diff
@@ -1,13 +1,19 @@
...
 </script>
 <template>
   <div class="relative">
-    <Icon size="1.2rem" name="shopware:heart" class="color-brand-primary" />
+    <!-- use when filled icon is ready -->
+    <!-- <Icon size="1rem" name="shopware:heart" v-if="type !== 'filled'" class="w-6 h-5 block hover:cursor-pointer" :class="[
+      styles[type]
+    ]" /> -->
+    <div class="i-carbon-favorite w-6 h-5 hover:cursor-pointer" v-if="!filled"></div>
+    <div class="i-carbon-favorite-filled w-6 h-5 hover:cursor-pointer" v-else></div>
```
mdanilowicz: Why do we need a filled icon for the header icon?

packages/helpers/src/product/isProductTopSeller.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Checks if a product is marked as a top seller
+ */
+export function isProductTopSeller(
+  product: { markAsTopseller?: boolean } | undefined,
+): boolean {
+  return !!product?.markAsTopseller;
+}
```
mdanilowicz:
```suggestion
/**
 * Checks if a product is marked as a top seller
 */
export function isProductTopSeller(
  product: { markAsTopseller?: boolean },
): boolean {
  return !!product.markAsTopseller;
}
```

Why is there a possibility of passing an empty product object? Can we avoid that scenario by checking the value before using the helper?

packages/helpers/src/product/isProductOnSale.ts
```diff
@@ -0,0 +1,12 @@
...
+ */
+export function isProductOnSale(product: {
+  calculatedPrice?: {
+    listPrice?: {
+      percentage?: number;
+    } | null;
+  };
+}): boolean {
+  return (product?.calculatedPrice?.listPrice?.percentage ?? 0) > 0;
+}
```
mdanilowicz: Same here, can we avoid having empty product object?

packages/helpers/src/product/getProductManufacturerName.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+/**
+ * Gets the translated name of the product manufacturer
+ */
+export function getProductManufacturerName(
+  product: ProductWithManufacturer | null | undefined,
+): string {
+  return product?.manufacturer?.translated?.name || "";
+}
```
mdanilowicz: Here is also problem with the empty product

packages/cms-base-layer/app/components/ui/WishlistIcon.vue
```diff
@@ -0,0 +1,21 @@
...
+    ]" /> -->
+    <div class="i-carbon-favorite w-6 h-5 hover:cursor-pointer" v-if="!filled"></div>
+    <div class="i-carbon-favorite-filled w-6 h-5 hover:cursor-pointer" v-else></div>
+    <SwMainCounter
+      v-if="(counter ?? 0) > 0"
+      class="absolute top-[-50%] left-[50%]"
+      :count="counter"
+    />
+  </div>
+</template>
```
mdanilowicz: I'm not sure that this is a good idea to have 2 copies of the same component. CMS should use a template component. Also, this is a header wishlist component, not a product. You don't need a counter on the product icon

packages/cms-base-layer/app/components/ui/IconButton.vue
```diff
@@ -0,0 +1,32 @@
+<script lang="ts" setup>
```
mdanilowicz, line 1: This component should be loaded from the template

packages/cms-base-layer/app/components/ui/MainCounter.vue
```diff
@@ -0,0 +1,20 @@
+<script lang="ts" setup>
```
mdanilowicz: This component should be loaded from the template

packages/cms-base-layer/app/components/SwProductCard.vue
```diff
@@ -5,9 +5,12 @@ import { useCmsTranslations } from "@shopware/composables";
 import {
```
mdanilowicz, line 5: This component should be refactored, split into smaller components. This will add more flexibility with the overriding

packages/cms-base-layer/app/components/ui/IconButton.vue
mkucmus: this will be used internally with an option to be replaced in a template

PR #1999, opened 2025-09-09
fix(vue-demo-store): newsletter race condition

mdanilowicz:
### Description

This pull request addresses a race condition when subscribing to the newsletter from the "My Account" page in the `vue-demo-store` project. The main improvement ensures the newsletter subscription process is atomic and prevents multiple simultaneous requests, resulting in a more reliable user experience.

Newsletter subscription reliability:

* Added a `newsletterDisabled` reactive variable to disable the newsletter checkbox during the subscription process, preventing duplicate submissions. (`templates/vue-demo-store/app/pages/account/index.vue`, [[1]](diffhunk://#diff-640f7989a59cd79f66d804083e1646583c646ac0f202df2d33eef4f501e9d5e9L37-R41) [[2]](diffhunk://#diff-640f7989a59cd79f66d804083e1646583c646ac0f202df2d33eef4f501e9d5e9R112)
* Refactored the `updateNewsletterStatus` function to use `await` for `getNewsletterStatus()` and ensure the checkbox is re-enabled only after the subscription status is updated. (`templates/vue-demo-store/app/pages/account/index.vue`, [templates/vue-demo-store/app/pages/account/index.vueL53-L62](diffhunk://#diff-640f7989a59cd79f66d804083e1646583c646ac0f202df2d33eef4f501e9d5e9L53-L62))

### DIFF
--- a/templates/vue-demo-store/app/pages/account/index.vue
+++ b/templates/vue-demo-store/app/pages/account/index.vue
@@ -34,8 +34,11 @@ useBreadcrumbs([
   },
 ]);
 
-const updateNewsletterStatus = async () => {
+const newsletterDisabled = ref(false);
+
+async function updateNewsletterStatus() {
   try {
+    newsletterDisabled.value = true;
     if (!newsletter.value) {
       await newsletterSubscribe({
         email: user.value?.email || "",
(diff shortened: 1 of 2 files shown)


templates/vue-demo-store/app/pages/account/index.vue
```diff
@@ -50,16 +53,15 @@ const updateNewsletterStatus = async () => {
     console.log("error", error);
     pushError(t("messages.error"));
   } finally {
-    getNewsletterStatus().then(() => {
-      newsletter.value = isNewsletterSubscriber.value;
-    });
+    await getNewsletterStatus();
```
patzick, line 64: shouldn't that be in another method? We can also have error with this endpoint and then the field stays disabled with no indication why

templates/vue-demo-store/app/pages/account/index.vue
```diff
@@ -107,6 +109,7 @@ onBeforeMount(async () => {
       <div class="flex">
         <input
           id="newsletter-checkbox"
+          :disabled="newsletterDisabled"
```
patzick, line 140: maybe some loading state instead of disabling field? So it's visible that something is going on

mkucmus: let's do the loading state implementation for forms or in general in a separate issue, I think it could be a nice component, shared over the entire project 🚀 I think we already have some UX for that or will be available shortly

templates/vue-demo-store/app/pages/account/index.vue
mdanilowicz: done

templates/vue-demo-store/app/pages/account/index.vue
mdanilowicz: sure, I added spinner in header section

patzick (approved):

PR #2003, opened 2025-09-12
feat: add loading state to dropdown components and new SalutationSelect component

### LINKED ISSUE #2002: Layers / Components / SalutationId

mdanilowicz:
### Description

This pull request introduces improved support for loading states in dropdown form components, adds a new `SalutationSelect` component with async data loading and caching, and enhances error message styling for input fields. Additionally, translations for the loading state have been added across multiple languages.

**Form dropdown loading state improvements:**

* Added a `loading` prop to both `BaseDropdown.vue` and `DropdownField.vue`, disabling the dropdown and showing a localized loading option when active. [[1]](diffhunk://#diff-6c9cebfb88d9a485dff119076abcd3439e75af0b449f8735903795baaadc0ba1R10) [[2]](diffhunk://#diff-6c9cebfb88d9a485dff119076abcd3439e75af0b449f8735903795baaadc0ba1L23-R28) [[3]](diffhunk://#diff-b63201a8006d3729fda2f17d22b12b07a8212a87eb087b30ba9dba062423c7b1R11) [[4]](diffhunk://#diff-b63201a8006d3729fda2f17d22b12b07a8212a87eb087b30ba9dba062423c7b1R35)
* Added `"loading"` translation strings to German, English, and Polish form locales. [[1]](diffhunk://#diff-3747d6798dd896a1b7732191464892c1aa6a127aabe9b0ad64f9c04aa2ad6531L40-R41) [[2]](diffhunk://#diff-d304c11288c2e01e7dbfc74fa77aa847ad6baee15580efc98f9c858795ecb0e4L40-R41) [[3]](diffhunk://#diff-4e9ff9245073449fb77a1037b0406ac177150a8e135656063b93ea55de036e4eL40-R41)

**New async dropdown component:**

* Introduced `SalutationSelect.vue`, which fetches salutation options asynchronously, caches results for 24 hours, and displays a loading indicator and error message as needed.

**Error message styling:**

* Updated the error message styling in `InputField.vue` to use the `text-states-error` class for consistency.

closes #2002


templates/vue-starter-template/app/components/form/SalutationSelect.vue
```diff
@@ -0,0 +1,53 @@
...
+const {
+  data: salutationData,
+  status,
+  error,
+} = await useAsyncData(
+  "salutationData",
+  () => apiClient.invoke("readSalutation post /salutation"),
+  {
+    transform: (apiData) => {
+      const options = apiData.data.elements.map((element) => ({
```
mkucmus, line 23: please prepare a helper/utils, even for the template that will transform API elements to be used as select option with one shared return interface "label" and "value". I think we could use it many times in the future 🤞🏼 like currencies, languages, states, etc.

templates/vue-starter-template/app/components/form/SalutationSelect.vue
```diff
@@ -0,0 +1,53 @@
...
+        label: element.displayName,
+        value: element.id,
+      }));
+      return { options, cachedDate: new Date() };
+    },
+    getCachedData: (key) => {
+      const data = nuxtApp.payload.data[key] || nuxtApp.static.data[key];
+
+      // 24 hours cache expiration
+      if (!data || Date.now() > new Date(data.cachedDate).getTime() + 86400000)
```
mkucmus: maybe we could move  `86400000` value to the public runtimeConfig? like to provide `defaultCacheLifetime` or smth similar we could control once. this caching is a good idea and we could use it more often in options for setting changing rarely

mkucmus (changes requested):
nicely done,
take a look at two comments. runtime config is optional

templates/vue-starter-template/app/components/form/SalutationSelect.vue
mdanilowicz: This is a one-liner. I'm not sure that adding a helper for that is a good idea. The main reason is that we can have different element fields for labels (not always `displayName`)

templates/vue-starter-template/app/components/form/SalutationSelect.vue
mdanilowicz: done

PR #2019, opened 2025-09-18
feat(account): add address management page

mdanilowicz:


templates/vue-starter-template/app/pages/account/address/index.vue
```diff
@@ -0,0 +1,185 @@
...
+  deletingAddresses.value.add(addressId);
+
+  try {
+    await deleteCustomerAddress(addressId);
+    await loadCustomerAddresses();
+    await refreshUser();
+  } catch (error) {
+    if (error instanceof ApiClientError) {
+      const errors = resolveApiErrors(error.details.errors);
+      for (const error of errors) {
```
mkucmus, line 38: rename inner `error` from the loop into `errorMessage` to avoid variable shadowing

templates/vue-starter-template/app/utils/validation/rules/addressFormRules.ts
```diff
@@ -0,0 +1,44 @@
...
+      required,
+    },
+    countryId: {
+      required,
+    },
+    countryStateId: {
+      required: requiredIf(() => {
+        return !!getStatesForCountry(
+          computed(() => state.value.countryId).value,
+        )?.length;
```
mkucmus: we are already in reactive context and wrapping and unpacking immediately is not needed I guess: 

```suggestion
        return !!getStatesForCountry(state.value.countryId)?.length;
```

templates/vue-starter-template/app/pages/account/address/new.vue
```diff
@@ -0,0 +1,62 @@
...
+        :subtitle="$t('account.address.new.subHeader')"
+      />
+
+      <div class="mb-10">
+        <AccountSectionHeader
+          class="mb-4"
+          :title="$t('account.address.new.personalDataSectionHeader')"
+        />
+
+        <AccountAddressForm :address="state" @handleSubmit="handleSubmit" />
```
mkucmus, line 55: take a look at the typecheck output:
`Type '{ firstName: string; ... }' is missing the following properties from type 'CustomerAddress': customerId, id`

templates/vue-starter-template/app/components/account/address/Form.vue
```diff
@@ -0,0 +1,143 @@
...
+    </p>
+
+    <div class="flex gap-4 mt-6">
+      <FormBaseButton type="submit" variant="primary">
+        {{ $t("account.address.saveButton") }}
+      </FormBaseButton>
+      <FormBaseButton
+        type="button"
+        variant="secondary"
+        @click="router.push('/account/address')"
```
mkucmus: should we use `formatLink` helper for i18n as well?

mkucmus (changes requested):
lovely PR, there are some small things to do 🤞🏼

PR #2033, opened 2025-09-26
chore: update ts config after migration to Nuxt4

stepan-twnty:
### Description
In the docs mentioned `.nuxt/tsconfig.json - Legacy configuration for backward compatibility`

### Additional context
With the old config, I had a problem with types in WebStorm


patzick (changes requested):
Good direction!
1. Unfortunately there's a typo - missing comma atfer the "include"
2. after that IDE tells that there's something wrong with the settings, any idea why? (screen)
3. would be great to also add this to vue-starter-template when it starts working with no errors

Image

mkucmus (commented):
currently we've got a weird situation that project references are the recommended future approach, but type checking currently requires the legacy format.

the community encountered some issues due to new references format in nuxt 4:

- https://github.com/nuxt/cli/issues/944
- https://github.com/nuxt/nuxt/issues/32714
- https://github.com/dargmuesli/vio/actions/runs/16426704531/job/46418719111?pr=316#step:7:726

`nuxt typecheck` will fail with the new config. I think we need to wait until the issue is fixed and then we can migrate entirely.

patzick:
hello, any updates with this?

PR #2041, opened 2025-09-30
docs: AGENTS.md guide for AI agents

mkucmus:
### Description

motivation: https://agents.md/

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,10 +1,29 @@
-# AI Agent Guide for Shopware Composable Frontends
+# AI Agent Guide for Shopware Frontends
 
-This document provides guidance for AI assistants working with the Shopware Composable Frontends monorepo.
+This document provides guidance for AI assistants working with the Shopware Frontends monorepo.
+
+## TL;DR
+
+**What**: Vue.js framework for Shopware 6 eCommerce storefronts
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,577 @@
+# AI Agent Guide for Shopware Frontends
```
mdanilowicz, line 1:
```
4. Large monorepo? Use nested AGENTS.md files for subprojects
Place another AGENTS.md inside each package. Agents automatically read the nearest file in the directory tree, so the closest one takes precedence and every subproject can ship tailored instructions. For example, at time of writing the main OpenAI repo has 88 AGENTS.md files.
```
Looks like in our case, we should have separated AGENTS.md for each subproject

mdanilowicz (changes requested):

AGENTS.md
mkucmus: so far we are not reaching more than 4 templates, and few packages - that means we are far away from being big monorepo with let's say 88 AGENTS.md files - also, the concerns from https://github.com/openai/agents.md/issues/53 are still not clear and we don't really know how it will behave.

let's start from one general one - thanks to the unified toolkit in env we have we can trust one general agents.md file and then we can extend it if needed.

patzick (approved):

PR #2084, opened 2025-10-20
test: nightly release of nitro to enable isr on vercel

mkucmus:
closes: #1652


mkucmus:
v2 was deprecated so we need to wait for v3 which is already in alpha stage. there is still an option with nightly version which has the fix - but not recommended for production as it's just a `main` branch released.

PR #2100, opened 2025-10-31
feat(cms-base-layer): image elements improvements

mkucmus:
### Description

- replaces  components
- provides config presets for common use cases
- adds provider for @nuxt/image module

### ToDo's

to consider: https://github.com/shopware/frontends/issues/2108

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -13984,6 +13984,9 @@ packages:
     resolution: {integrity: sha512-RwNA9Z/7PrK06rYLIzFMlaF+l73iwpzsqRIFgbMLbTcLD6cOao82TaWefPXQvB2fOC4AjuYSEndS7N/mTCbkdQ==}
     engines: {node: '>= 0.8'}
 
+  std-env@3.10.0:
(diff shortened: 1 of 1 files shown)


packages/cms-base-layer/app/components/public/cms/element/CmsElementImage.vue
```diff
@@ -27,7 +26,7 @@ const {
 } = useCmsElementImage(props.content);
 
 const DEFAULT_THUMBNAIL_SIZE = 10;
-const imageElement = useTemplateRef("imageElement");
+const imageElement = ref<HTMLImageElement>();
```
mdanilowicz: Why did you change from templateRef to ref?

mkucmus: typecheck was failing due to incompatibility between useTemplateRef and the old @nuxt/image module. Now it's like expected -  once I've upgraded to v 2.0 of the module 💪🏼

patzick (approved):
🖼️

PR #2127, opened 2025-11-19
fix(api-types): update LineItem cover type from ProductMedia to Media across examples and packages

mdanilowicz:
… across examples and packages

This pull request introduces an improvement to how product images are handled in the cart component, ensuring more consistent and simplified access to image data. The main change is in how the `cover` property is referenced and used for cart items, both in the API schema and the Vue component.

**API Schema Update:**

* Added a `cover` property to the `LineItem` schema, referencing the `Media` schema to standardize how media is attached to cart items.

**Component Logic Simplification:**

* Updated the product image rendering logic in `ProductItem.vue` to directly use `cartItem.cover` instead of accessing a nested `media` property, aligning with the updated API schema and reducing complexity.

### DIFF
--- a/packages/helpers/src/product/getMainImageUrl.ts
+++ b/packages/helpers/src/product/getMainImageUrl.ts
@@ -10,7 +10,7 @@
 export function getMainImageUrl<
   T extends
     | {
-        cover?: {
+        cover: {
           media?: {
             url: string;
           };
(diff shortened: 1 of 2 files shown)


mdanilowicz:
> Very nice!
> 
> One question though - isn't the `cover` supposed to be a required property? What are the cases that it's missing, are you able to reproduce that case?

Cover shouldn’t be required because the coupon code has the property set to null
Screenshot 2025-11-20 at 20 55 26

patzick (changes requested):
so should be
`cover: components["schemas"]["Media"] | null`
as the result

PR #2150, opened 2025-11-27
fix: audit with node-forge

patzick:
### Description

- audit fixes
- bump node to allow 24.x


mkucmus (approved):
todo: check docs deployment separately

PR #2162, opened 2025-12-08
feat(cms-base): use image placeholder and replace obsolete colors

mkucmus:
#### packages/cms-base-layer

Replace hardcoded colors with theme tokens, add image placeholder composable, improve URL encoding for special characters in image paths, enhance CMS block layouts, and use useTemplateRef for better type safety.

#### packages/helpers
 
Export encodeUrlPath helper function to properly handle special characters (spaces, commas) in image URLs


packages/cms-base-layer/package.json
```diff
@@ -15,13 +15,22 @@
...
+    "helpers",
+    "index.cjs",
+    "index.d.ts",
+    "app.config.ts",
+    "nuxt.config.ts",
+    "uno.config.ts"
+  ],
   "scripts": {
-    "build": "unbuild",
+    "build": "nuxt prepare && unbuild",
```
patzick, line 29: isn't `nuxt prepare` invoked after `pnpm install`?

mkucmus: the problem appeared when CI runs `typecheck` - then tsconfig expects the types are pregenerated. I added it and issue is gone. maybe it can be solved in other way around 🤔 

such problems will be solved entirely if we change the dev tooling of a package to be a nuxt based -> #2163

PR #2169, opened 2025-12-10
fix(docs): register missing pageRef component

mkucmus:
### Description

adds missing PageRef component, responsible for displaying internal links with description.

### Type of change

### DIFF
--- a/.changeset/fresh-feet-enter.md
+++ b/.changeset/fresh-feet-enter.md
@@ -1,0 +1,5 @@
+---
+"@shopware/cms-base-layer": patch
+---
+
+Added a new image placeholder.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,7 +8,7 @@ This document provides guidance for AI assistants working with the Shopware Fron
 **Structure**: pnpm monorepo with Turbo
 **Tech**: Vue 3, Nuxt 4, TypeScript, Vitest, Biome
 **Key Packages**: api-client, composables, helpers, cms-base-layer, nuxt-module
-**Templates**: vue-demo-store (full featured), vue-starter-template (production), vue-blank (minimal)
+**Templates**: vue-demo-store (full featured), vue-starter-template (production), vue-starter-template-extended (layer example), vue-blank (minimal)
 
 **Quick Start**:
 ```bash
@@ -40,11 +40,12 @@ frontends/
 │   ├── api-gen/              # Type generation tool
 │   └── eslint-config-shopware/
 ├── templates/         # Starter templates
-│   ├── vue-demo-store/       # Full demo with UnoCSS, i18n, CMS
-│   ├── vue-starter-template/ # Production-ready starter
-│   ├── vue-blank/            # Minimal Nuxt setup
-│   ├── vue-vite-blank/       # Minimal Vite setup
-│   └── astro/                # Astro integration example
+│   ├── vue-demo-store/                # Full demo with UnoCSS, i18n, CMS
+│   ├── vue-starter-template/          # Production-ready starter
+│   ├── vue-starter-template-extended/ # Layer extension example
+│   ├── vue-blank/                     # Minimal Nuxt setup
+│   ├── vue-vite-blank/                # Minimal Vite setup
+│   └── astro/                         # Astro integration example
 ├── apps/              # Applications
 │   ├── docs/                 # Documentation site
 │   └── e2e-tests/            # E2E test suite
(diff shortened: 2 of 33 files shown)


apps/docs/.vitepress/theme/index.ts
```diff
@@ -12,6 +13,7 @@ export default Object.assign(
   {
     enhanceApp({ app }: { app: App }) {
       // app.component("AI", AI);
+      app.component("PageRef", PageRef);
```
mkucmus: @bojanrajh can we register this component also locally? I guess you overwrite it anyway in developer portal right?

mkucmus: also, wdyt about having a redirection from frontends.shopware.com -> https://developer.shopware.com/frontends/ ? 

we have only one note: missing overview page, replaced by welcome page with a really nice graphics ;)

patzick (approved):

apps/docs/.vitepress/theme/index.ts
patzick: I'd keep it for now, separate domain is nice when you want to browse only frontends specific stuff, otherwise global search is fine but it's not for everyone and could be overwhelming for someone who just wants to have pretty shop ;)

apps/docs/.vitepress/theme/index.ts
```diff
@@ -12,6 +13,7 @@ export default Object.assign(
...
       // app.provide('some-injection-key-if-needed', VALUE)
     },
   },
```
bojanrajh: Actually, your theme overrides the whole `enhanceApp()` from the [base theme](https://github.com/shopware/developer-documentation-vitepress/blob/main/src/shopware/theme.ts#L107). This should fix it.

```suggestion
export default SWAGTheme()
```

You can then enhance the app by passing the `enhanceApp` as a config:

```ts
export default SWAGTheme({
  enhanceApp({ app }: { app: App }) {
    // app.component("PageRef", PageRef);
  }
})
```

mkucmus: thanks for mentioning it! unfortunately, the last change that makes PageRef links relative (by adding a dot ".") breaks our internal docs navigation. 

it takes wrong current dir as a context to relative link. so now the PageRef creates a link https://github.com/shopware/frontends/commit/57d179d44bb9fa72cd2e783453c19074e690bdb2#diff-1097c28b29e09dfb83f3fd2058a9c06e54c2ee81c47ebf312517019a281f9f0fR69 but relatively from "templates" dir, not the base docs path. that leads to 404... @bojanrajh does it work in developer portal context or we need to provide some fix for that?

image

apps/docs/package.json
```diff
@@ -7,6 +7,7 @@
   "scripts": {
     "dev": "vitepress",
     "build": "vitepress build",
+    "build:devhub": "VITEPRESS_BASE=/frontends/ vitepress build",
```
mkucmus, line 10: @bojanrajh I added the usage of `withBase()` function of vitepress to enable you and us to control the prefix in URL. since now on we can continue to provide an absolute path for PageRefs globally.

that fixes the problem in both environments - dev portal and our standalone docs.

cc: @patzick

apps/docs/.vitepress/theme/index.ts
mkucmus: after [that change ](https://github.com/shopware/frontends/pull/2169/changes/7897665c45509463bec40677116e91d90b732741)the problem should be gone.

apps/docs/package.json
bojanrajh: For sure? How does `developer-portal` know when to apply the `/frontends` base? The `PageRef` component from this repo is never registered.

mkucmus: hmmm, there is no other way to reconcile two environments I guess... so my suggestion would be to provide awareness of `baseUrl` the docs are deployed under.

bojanrajh: @mkucmus All those links are now broken, see last PageRef on this page:
 - https://developer.shopware.com/frontends/best-practices/images.html
 - https://frontends.shopware.com/best-practices/images.html

Frontends repo is mounted to `/frontends/` in DevHub, and to `/` in standalone mode. Links need to be relative because the base is not known upfront. Please, revert, those links were relative intentionally.

https://github.com/shopware/frontends/pull/2168/changes
https://github.com/shopware/frontends/pull/1937/changes

Issue #1764, opened 2025-03-14
[FEATURE] Initially load filters on product listings


AlexBa:
### Describe the feature

At the moment, there exist a "loading placeholder" for the initial loading of the product listing filters: https://frontends-demo.vercel.app/CMS/

However, because of SEO purposes, it would be better to show the full HTML markup initially. Additionally, the UI/UX is not very nice, because for the end user the filters are "just flickering" and changing too fast.

I understand that the client state is saved there and hydration mismatches are a topic, but the separation must be more granular. The available filters could be easily exposed in the initial HTML markup via SSR. 

Only the selected/active state must be determined on the client side. However, URLs query params for the active/selected filters also stored the URL. Thus, there is a possibility for SSR too.

### Description

At the moment, there exist a "loading placeholder" for the initial loading of the product listing filters: https://frontends-demo.vercel.app/CMS/

However, because of SEO purposes, it would be better to show the full HTML markup initially. Additionally, the UI/UX is not very nice, because for the end user the filters are "just flickering" and changing too fast.

I understand that the client state is saved there and hydration mismatches are a topic, but the separation must be more granular. The available filters could be easily exposed in the initial HTML markup via SSR. 

Only the selected/active state must be determined on the client side. However, URLs query params for the active/selected filters also stored the URL. Thus, there is a possibility for SSR too.

### Acceptance Criteria

- Product listing filters show be available in the initial markup (Server side rendered)
- Active/selected state can be determined on client side or with the help of query param via SSR.

### Additional information

- [ ] Would you be willing to help implement this feature?

### Checks

- [x] Check existing [discussions](https://github.com/shopware/frontends/discussions) and [issues](https://github.com/shopware/frontends/issues).

PR #2176, opened 2025-12-15
feat(cms-base-layer): make product-listing components SSR friendly

mkucmus:
### Description

closes: #1764

### Type of change

### DIFF
--- a/packages/helpers/src/cms/getProductListingFromCmsPage.ts
+++ b/packages/helpers/src/cms/getProductListingFromCmsPage.ts
@@ -1,8 +1,23 @@
+type CmsPageStructure = {
+  sections: Array<{
+    blocks?: Array<{
+      slots?: Array<{
+        type?: string;
+        data?: unknown;
+        [key: string]: unknown;
+      }>;
+      [key: string]: unknown;
+    }>;
+    [key: string]: unknown;
+  }>;
+  [key: string]: unknown;
+};
+
 /**
  * Extracts the product listing data from a CMS page structure.
  * Useful for SSR to get listing data early before components render.
  *
- * @param cmsPage - The CMS page object
+ * @param cmsPage - The CMS page object (Schemas["CmsPage"])
  * @returns The product listing result or null if not found
  *
  * @example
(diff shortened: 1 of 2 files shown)


packages/helpers/src/cms/getProductListingFromCmsPage.ts
```diff
@@ -0,0 +1,50 @@
...
+ * @returns The product listing result or null if not found
+ *
+ * @example
+ * ```ts
+ * const listing = getProductListingFromCmsPage(cmsPage);
+ * createCategoryListingContext(listing);
+ * ```
+ */
+export function getProductListingFromCmsPage<T = unknown>(
+  cmsPage: unknown,
```
patzick: we should write minimal expected type generic here, not rely on unknown and assume typing in function

packages/cms-base-layer/app/components/public/cms/CmsPage.vue
```diff
@@ -13,10 +14,24 @@ const props = defineProps<{
...
+
+// Function to initialize or update listing context
+function updateListingContext(content: Schemas["CmsPage"]) {
+  if (routeName.value === "frontend.navigation.page") {
+    const initialListing =
+      getProductListingFromCmsPage<Schemas["ProductListingResult"]>(content);
+
+    if (initialListing) {
+      createCategoryListingContext(initialListing);
+    }
```
patzick, line 26: why not invoke existing setInitialListing here?

patzick (changes requested):

packages/cms-base-layer/app/components/public/cms/CmsPage.vue
mkucmus: using `setInitialListing` approach has one inconvenience - we would need to init an instance of useListing* composable even if CmsPage is not in type of product listing. 

this approach is cleaner because we inject product listing (by creating a context) and the responsibility of owning a composable instance is moved to any child component with an easy access to this data by using useCategoryListing - so more flexibility in case someone didn't want to use our entire implementation.

packages/helpers/src/cms/getProductListingFromCmsPage.ts
mkucmus: I tried to make it more predictable

PR #2202, opened 2025-12-29
chore: GHSA-8452-54wp-rmv6

mkucmus:
### Description

fixes GHSA-8452-54wp-rmv6

core issue reported: https://github.com/shopware/developer-documentation-vitepress/issues/592

### DIFF
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -26,6 +26,7 @@
     "@types/node": "22.13.14",
     "@shopware-docs/cli": "^1.3.0-alpha.16",
     "@shopware-docs/storybook": "^1.3.0-alpha.16",
+    "storybook": "7.6.21",
     "@shopware-docs/typer": "^1.3.0-alpha.16",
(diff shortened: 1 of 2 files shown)


apps/docs/package.json
```diff
@@ -26,6 +26,7 @@
     "@types/node": "22.13.14",
     "@shopware-docs/cli": "^1.3.0-alpha.16",
     "@shopware-docs/storybook": "^1.3.0-alpha.16",
+    "storybook": "7.6.21",
```
mdanilowicz, line 29: Why not `pnpm.overrides`? It looks like we are adding a dependency, but we are not using it directly

mdanilowicz (changes requested):

apps/docs/package.json
mkucmus: it won't work for peer dep - already checked this one and created a ticket in the correspoding repository

PR #2226, opened 2026-01-15
feat(cms-base-layer): use cms config to control gallery navigation

mkucmus:
### Description

use config defined in cms to show gallery navigation


packages/cms-base-layer/app/components/public/cms/element/CmsElementImageGallery.vue
```diff
@@ -13,6 +13,16 @@ const props = defineProps<{
   content: CmsElementImageGallery;
 }>();
 
+const { getConfigValue } = useCmsElementConfig(props.content);
+
+const minHeight = computed(() => getConfigValue("minHeight") || "500px");
```
mdanilowicz: Can we move this default value to a variable? 
Do we need a default value here if it's declared in `SwProductGallery`?

packages/cms-base-layer/app/components/public/cms/element/CmsElementImageGallery.vue
```diff
@@ -139,7 +155,12 @@ function onTouchEnd() {
 
         <!-- Next Button -->
         <button
-          class="w-10 h-10 rounded-full transition bg-white/20 hover:bg-white/50 disabled:opacity-50 pointer-events-auto shadow-lg flex items-center justify-center"
+          :class="[
+            'w-10 h-10 rounded-full transition disabled:opacity-50 pointer-events-auto shadow-lg flex items-center justify-center',
+            navigationArrows === 'outside'
+              ? 'bg-brand-tertiary text-surface-on-surface'
+              : 'bg-white/20 hover:bg-white/50',
```
mdanilowicz:
```suggestion
              : 'bg-surface-surface/20 hover:bg-surface-surface/50',
```

packages/cms-base-layer/app/components/SwProductGallery.vue
```diff
@@ -1,30 +1,39 @@
...
+  CmsElementImageGallery,
+  SliderElementConfig,
+} from "@shopware/composables";
 import { ref, watch } from "vue";
 import type { Schemas } from "#shopware";
 
-const props = defineProps<{
-  product: Schemas["Product"];
-}>();
+const props = withDefaults(
```
mdanilowicz: Please define props with destructuring

packages/cms-base-layer/app/components/public/cms/element/CmsElementImageGallery.vue
mkucmus: yes, CmsElementImageGallery can be used independently (not just via SwProductGallery), so defaults make sense - moved to the const ✅ .

packages/cms-base-layer/app/components/SwProductGallery.vue
mkucmus: done

mdanilowicz (approved):

PR #2262, opened 2026-01-28
feat(api-gen): support for client_credentials grant type auth

mkucmus:
### Description

Added support for `client_credentials` grant type authentication when loading Admin API schema.

### DIFF
--- a/packages/api-gen/src/commands/loadSchema.ts
+++ b/packages/api-gen/src/commands/loadSchema.ts
@@ -26,7 +26,7 @@ const ADMIN_API_ENDPOINT = `/api/${SCHEMA_ENDPOINT}`;
  *
  * For `admin` API (client_credentials grant):
  * - OPENAPI_JSON_URL
- * - SHOPWARE_ADMIN_CLIENT_ID (optional, defaults to "administration")
+ * - SHOPWARE_ADMIN_CLIENT_ID
  * - SHOPWARE_ADMIN_CLIENT_SECRET
  *
  */


packages/api-gen/src/apiClient.ts
```diff
@@ -7,15 +7,24 @@ let storeApiClient: ReturnType<typeof createAPIClient<operations>>;
...
+    const credentials = process.env.SHOPWARE_ADMIN_CLIENT_SECRET
+      ? ({
+          grant_type: "client_credentials",
+          client_id: process.env.SHOPWARE_ADMIN_CLIENT_ID || "administration",
+          client_secret: process.env.SHOPWARE_ADMIN_CLIENT_SECRET,
+        } as const)
+      : ({
+          grant_type: "password",
+          client_id: "administration",
+          scope: "write",
```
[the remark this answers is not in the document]
mkucmus: its done via #2261

packages/api-gen/src/commands/loadSchema.ts
```diff
@@ -51,10 +56,14 @@ export async function loadSchema(args: {
     : `${args.apiType}ApiSchema.json`;
 
   const OPENAPI_JSON_URL = process.env.OPENAPI_JSON_URL;
-  const requiredEnvVars = [];
+  const requiredEnvVars: string[] = [];
   if (isAdminApi) {
-    requiredEnvVars.push("SHOPWARE_ADMIN_USERNAME");
-    requiredEnvVars.push("SHOPWARE_ADMIN_PASSWORD");
+    // Support both password and client_credentials grant types
+    const hasClientCredentials = !!process.env.SHOPWARE_ADMIN_CLIENT_SECRET;
```
patzick: `SHOPWARE_ADMIN_CLIENT_ID` is also required, so if someone put `SHOPWARE_ADMIN_CLIENT_ID` only they shouild see `SHOPWARE_ADMIN_CLIENT_SECRET`

and if someone just puts `SHOPWARE_ADMIN_CLIENT_SECRET` then this check will pass and authentication will fail

we need to test this part, we can extract it to separate function and pass process.env as its parameter, then it should be easy to add tests

packages/api-gen/src/apiClient.ts
```diff
@@ -7,15 +7,24 @@ let storeApiClient: ReturnType<typeof createAPIClient<operations>>;
...
+          username: process.env.SHOPWARE_ADMIN_USERNAME || "",
+          password: process.env.SHOPWARE_ADMIN_PASSWORD || "",
+        } as unknown as adminOperations["token post /oauth/token"]["body"]);
```
patzick: we have strong typing for a reason, this casting just claims to be in order with schema :)

patzick: I'll push the fix for this part

mkucmus: thanks

packages/api-gen/src/commands/loadSchema.ts
mkucmus: check this now 🔍

mkucmus: validator is now updated. in case of client_credentials auth - both envs are required.

PR #2263, opened 2026-01-28
feat(api-client): update `storeApiTypes.d.ts` based on the latest API release 

mkucmus:
### Description

- updates store api types exported by api-client
- removes obsolete entries from local overrides patch


packages/api-client/api-types/storeApiSchema.overrides.json
```diff
@@ -153,44 +153,31 @@
...
-        }
-      }
-    },
     "/role": {
       "get": {
         "requestBody": "_DELETE_"
       }
     },
-    "/shipping-method": {
+    "/dsr/customer/automatic-login": {
```
patzick: dsr is in commercial package, we should put that in another overrides file ans stack

mkucmus: good point! added 👍

patzick (approved):

PR #2290, opened 2026-02-19
feat(composables): add support for deep link code option in order loading

### LINKED ISSUE #1390: [BUG] DeepLink for Guests

### Current Behavior

When you order in the default Storefront as a Guest, you will receive an email with a `codeDeepLink` that lets you view all the details about your order.

Issues Composable Frontends have with that:
- The default link from the email template is not working or opening the orders in our demo store (there is a # missing in the URL)
- The link in total is not working for guests because you can not load an order without a context token

### Expected Behavior

- Our demo store opens the correct order without the need to change the link via admin
- Also, guests can view their order details when they use the link from email that is containing the `codeDeepLink` (order can be loaded via admin API but not via store API)

### Steps To Reproduce

_No response_

### Environment

_No response_

### Anything else?

_No response_

### LINKED ISSUE #2063: [BUG] useOrderDetails is not supporting deep link code
### Current Behavior

useOrderDetails composable should support loading orders by deepLinkCode. 
To consider new composable

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Anything else?

_No response_

mdanilowicz:
This pull request introduces a comprehensive order detail feature for the Vue demo store, supporting both guest and authenticated users. It adds new dedicated pages for order details, implements guest order lookup via deep link code, and introduces several new UI components for displaying order information. The `useOrderDetails` composable has been enhanced to support fetching orders either by ID or deep link code, and relevant tests have been added to ensure this functionality.

Order detail pages and guest order lookup:

* Added order detail page accessible via deep link code (`/account/order/[deepCode]`) with guest verification (email + postal code) and a dedicated order details page for logged-in users (`/account/order/details/[id]`).
* Enhanced `useOrderDetails` composable to support an `options` parameter with `isDeepLinkCode` flag, enabling order fetching by deep link code for guest users. [[1]](diffhunk://#diff-d51f6a9301348a9fab69330d17a95520f78846ce1416aeb3dcf6169e4aa177d9R1-R5) [[2]](diffhunk://#diff-981efcfefd2c9b942504ebd4454a182ee616bc9a9e1f9789994a1dd73879b1e0R7-R15) [[3]](diffhunk://#diff-981efcfefd2c9b942504ebd4454a182ee616bc9a9e1f9789994a1dd73879b1e0R139) [[4]](diffhunk://#diff-981efcfefd2c9b942504ebd4454a182ee616bc9a9e1f9789994a1dd73879b1e0L185-R217)
* Added and updated tests for `useOrderDetails` to cover deep link code and ID-based order fetching scenarios.

New order detail components:

* Introduced new components for order details: `Addresses.vue`, `LineItems.vue`, `PriceSummary.vue`, `PaymentMethodCard.vue`, `ChangePaymentModal.vue`, and `ShippingMethodInfo.vue`, each handling specific aspects of the order UI and i18n. [[1]](diffhunk://#diff-0b95e9972c207975682aa410121f043c6523bd8cb9072843a9dc0aaa949dfb7eR1-R48) [[2]](diffhunk://#diff-cc68d9fda7414a9055fc9198c21b2e9883a65257739e2f851174f70abde8b727R1-R124) [[3]](diffhunk://#diff-601417bb2f36f3654573ee141bf373ce65ec1247009613c877b9dac8e5c7c387R1-R58) [[4]](diffhunk://#diff-9cfe006d3d6fed343f9dd1979cb1635ead7646bc2159b58465198953e823302bR1-R50) [[5]](diffhunk://#diff-f5a2118910b1362241b666acc073d6535ef1acd1f9dee47028e16ff139ccd8dbR1-R97) [[6]](diffhunk://#diff-e045ccaf1ef040bd3701c139704a2c0f4ac61de08dae4fc22862a2ca8bf35fe9R1-R35)

Internationalization and UI improvements:

* Added i18n support for order authentication and messages, improving the user experience for both guest and logged-in users.

 closes #2063 
 closes #1390

### DIFF
--- a/.changeset/warm-pandas-glow.md
+++ b/.changeset/warm-pandas-glow.md
@@ -1,5 +1,0 @@
----
-"@shopware/composables": minor
----
-
-Add `options` parameter with `isGuestOrder` flag to `useOrderDetails` composable. When enabled, the order is fetched using a deep link code filter instead of an order ID, enabling guest order lookups without authentication.


packages/composables/src/useOrderDetails/useOrderDetails.ts
```diff
@@ -4,6 +4,15 @@ import type { ComputedRef, Ref } from "vue";
 import { useDefaultOrderAssociations, useShopwareContext } from "#imports";
 import type { Schemas, operations } from "#shopware";
 
+export type UseOrderDetailsOptions = {
+  /**
+   * When true, the first parameter is treated as a deep link code
+   * instead of an order ID. This enables fetching orders for guest
+   * users who are not logged in.
+   */
+  isDeepLinkCode?: boolean;
```
patzick: i think `isGuestOrder` would be more descriptive in what we're trying to do here, we definetly should propose an endpoint where it's resolved automatically, but first a few questions:
- Is the order structure exactly the same as with logged in user? Can we use the same methods like `cancel` or `changePaymentMethod`?
- Why you did not implemented this in this PR but you manually load these orders in `[order].vue` page?
- other methods are using orderId in invocations, this would use `deepLink` in the cover of `orderId` fact which would be very hard to debug

To sum that up, I'm more leaning toward creating dedicated `useGuestOrder` composable

patzick (changes requested):

packages/composables/src/useOrderDetails/useOrderDetails.ts
mdanilowicz: > i think isGuestOrder would be more descriptive

Done

> Is the order structure exactly the same as with logged in user? Can we use the same methods like cancel or changePaymentMethod?

We are receiving a new context API token, so the customer is logged in as a guest customer

> Why you did not implemented this in this PR but you manually load these orders in [order].vue page?

Sorry, but I don't understand. Can you add more context? 

> other methods are using orderId in invocations, this would use deepLink in the cover of orderId fact which would be very hard to debug 

`deepLink` code is used only on the init call to verify token or credentials (if form is displayed). All other calls are using orderID as an entity ID, the deepLink code is not used.

> To sum that up, I'm more leaning toward creating dedicated useGuestOrder composable 

There is no need to have a separate composable because of the context token for guest users. 

We could stop fetching data directly inside the composable and instead inject the data into it from the outside. However, this would be a breaking change, as it would alter the way the composable is currently used.

patzick: > We are receiving a new context API token, so the customer is logged in as a guest customer

So when you load order with deeplink you're also logging in as guest during that operation?

> Sorry, but I don't understand. Can you add more context?

Not sure how to describe it differently - you changed composable but haven't used that change anywhere :)

> deepLink code is used only on the init call to verify token or credentials (if form is displayed). All other calls are using orderID as an entity ID, the deepLink code is not used.

And this is where confusion will also start for other users - you're passing `orderId` creating composable but in fact it's `deepLInk` - any other method will use that as `orderId`, you added logic only in a single loadOrder method so it can actually recognise that it's not orderId. This speaks even more into having separate composable for that case

mdanilowicz: > So when you load order with deeplink you're also logging in as guest during that operation?

Yes

> Not sure how to describe it differently - you changed composable but haven't used that change anywhere :)

 Ah, useOrderDetails, sure, removed

>And this is where confusion will also start for other users - you're passing orderId creating composable but in fact it's deepLInk - any other method will use that as orderId, you added logic only in a single loadOrder method so it can actually recognise that it's not orderId. This speaks even more into having separate composable for that case

No, no, we are using deepLink only for the init request. After that, you use orderId as the main identifier key. The main problem is that we are fetching data inside the composable. Injecting data into the composable resolves all problems, one composable for all types of order data.

.changeset/warm-pandas-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@shopware/composables": minor
```
patzick: this file not relevant anymore, right?

PR #2406, opened 2026-04-17
feat: unoCSS nuxt layer 

mdanilowicz:
## ✨ Architectural Improvement: UnoCSS Design Tokens Layer

This branch introduces an important architectural change:
👉 **UnoCSS design tokens have been extracted from `@shopware/cms-base-layer` and moved into a dedicated Nuxt layer:**
`@shopware/unocss-design-tokens-layer`.

---

## 🧩 Why this change?

The `cms-base-layer` should stay focused on what it does best:

* 🧱 CMS rendering
* ♻️ Reusable storefront components

It **should not own styling decisions or a design system**.

Previously, keeping UnoCSS tokens inside `cms-base-layer` created tight coupling between:

* ⚙️ Functional CMS logic
* 🎨 Visual design choices

This meant every consumer of the CMS layer was **implicitly forced to adopt styling decisions**, which should instead be:

* optional
* composable
* explicit

---

### 1. Clear separation of concerns

* 🧠 CMS logic → stays in `cms-base-layer`
* 🎨 Design tokens → live in a dedicated layer

---

### 2. More flexibility for Nuxt layers

Teams can now:

* ➕ Include the token layer
* 🔁 Replace it
* 🛠️ Extend it

➡️ All **without modifying the CMS package**

---

### 3. Explicit styling dependencies

* ❌ No more hidden styling side effects
* ✅ Consumers (e.g. `vue-starter-template`) define styling consciously

---

## 🛠️ Maintainability improvements

* 📦 Centralized UnoCSS tokens (no duplication across templates)
* 🔄 Easier updates and theme changes
* 🧼 Cleaner architecture in layered storefronts

The updated `vue-starter-template` now:

* extends the dedicated token layer
* merges its local UnoCSS config with the base layer config

➡️ This is a **scalable and clean pattern** for layered projects.

---

## 🔍 Improved predictability

Previously:

* ⚠️ Some styling relied on runtime UnoCSS behavior in `cms-base-layer`
* 🤯 This could introduce hidden complexity

Now:

* 📍 Design tokens have a **clear, single source of truth**
* 🧩 CMS layer becomes more **framework-neutral** in terms of styling

---

## 🧠 In short

This change makes the codebase:

* 🧩 **More modular**
* 🪶 **Easier to extend**
* 🛠️ **Easier to maintain**
* 🏗️ **Better aligned with layered architecture principles**

---

A small structural change — but a big step toward a cleaner and more scalable frontend architecture 🚀


BrocksiNet:
Should that not be a major version for the CMS package?

packages/unocss-design-tokens-layer/app/plugins/unocss-runtime.client.ts
```diff
@@ -1,7 +1,8 @@
-import { presetWind3 } from "@unocss/preset-wind3";
 import initUnocssRuntime from "@unocss/runtime";
 import { defineNuxtPlugin, useAppConfig } from "#imports";
-import { cmsBaseLayerUnoTheme } from "../utils/unocss-theme";
+import config from "../../uno.config";
+
+initUnocssRuntime({ defaults: config });
```
mkucmus:
```suggestion
```

`initUnocssRuntime` is called at module load, outside defineNuxtPlugin. This bypasses the opt-out check below, runs the initializer even when unocssRuntime: false

packages/unocss-design-tokens-layer/app/plugins/unocss-runtime.client.ts
```diff
@@ -11,13 +12,9 @@ import { cmsBaseLayerUnoTheme } from "../utils/unocss-theme";
 // When disabled, CMS classes unknown at build time won't be resolved —
 // add them to the UnoCSS safelist in uno.config.ts to ensure they are generated.
 export default defineNuxtPlugin(() => {
+  console.log("unocss-runtime.client");
```
mkucmus:
```suggestion
```

packages/cms-base-layer/app/app.config.ts
```diff
@@ -15,5 +15,4 @@ export default defineAppConfig({
     3: "(max-width: 768px) 100vw, 33vw",
     default: "(max-width: 768px) 50vw, 25vw",
   },
-  unocssRuntime: true,
```
mkucmus, line 18: the `unocssRuntime: true` default was removed from `cms-base-layer/app/app.config.ts` but not added in new layer's `app.config.ts`. So `appConfig.unocssRuntime` is `undefined`, the plugin's `if (!appConfig.unocssRuntime) return;` always exits, and the runtime is off by default, the README says otherwise

add `app/app.config.ts` with `defineAppConfig({ unocssRuntime: true })`

packages/unocss-design-tokens-layer/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "license": "MIT",
+  "type": "module",
+  "main": "./nuxt.config.ts",
+  "files": [
+    "dist",
+    "app",
+    "nuxt.config.ts",
+    "index.cjs",
+    "index.d.ts",
+    "app.config.ts",
```
mkucmus, line 24: `app.config.ts` file is missing - mentioned in some comment above.

what about `index.cjs`?

templates/vue-starter-template/package.json
```diff
@@ -35,6 +35,7 @@
     "@shopware/composables": "canary",
     "@shopware/helpers": "canary",
     "@shopware/nuxt-module": "canary",
+    "@shopware/unocss-design-tokens-layer": "workspace:*",
```
mkucmus, line 38: please create a follow up issue to change "workspace:*" → "canary" once PR gets merged into main, because it will be a blocker for tiged-based scaffolding: `workspace:*` doesn't resolve for end users as we explain in docs.

packages/unocss-design-tokens-layer/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@shopware/unocss-design-tokens-layer",
+  "version": "0.0.0",
+  "description": "Empty Nuxt layer for UnoCSS design tokens",
```
mkucmus:
```suggestion
  "description": "Shared Shopware UnoCSS design tokens and runtime plugin as a Nuxt layer",
```

as it's not totally empty

packages/unocss-design-tokens-layer/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "dev": "unbuild --stub",
+    "lint": "biome check .",
+    "lint:fix": "biome check . --write && pnpm run typecheck",
+    "typecheck": "pnpm nuxt prepare && tsgo --noEmit"
+  },
+  "dependencies": {
+    "@unocss/nuxt": "66.6.0",
+    "@unocss/preset-wind3": "66.6.0",
+    "@unocss/runtime": "66.6.0",
+    "unocss": "66.6.0"
```
mkucmus, line 39: add `@unocss/core` to deps.

```suggestion
  "dependencies": {
    "@unocss/nuxt": "66.6.0",
    "@unocss/core": "66.6.0",
    "@unocss/preset-wind3": "66.6.0",
    "@unocss/runtime": "66.6.0",
    "unocss": "66.6.0"
```

`@unocss/core` is imported (as a type) by `packages/unocss-design-tokens-layer/uno.config.ts` but isn't declared in dependencies.

mkucmus (changes requested):
I like the extraction idea, there are few points to take care of and we are good to go with that.

mkucmus (approved):

PR #2417, opened 2026-04-23
chore: GHSA-j687-52p2-xcff GHSA-w5hq-g745-h8pq GHSA-v3rj-xjv7-4jmq

mkucmus:
fixes: GHSA-j687-52p2-xcff GHSA-w5hq-g745-h8pq GHSA-v3rj-xjv7-4jmq


package.json
```diff
@@ -60,6 +60,8 @@
       "semver": "^7.6.2",
       "qs": "^6.14.2",
       "undici": "^6.24.0",
+      "uuid": "^14.0.0",
```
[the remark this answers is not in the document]
mkucmus: I updated the example

mdanilowicz, line 63: Why in the main package? should be in example

mdanilowicz (changes requested):

package.json
mkucmus: hey, thanks for that, I added it explicitly here, however uuid has to stay at workspace root: pnpm only honors pnpm.overrides there in workspaces, and the advisory hits 2 chains unrelated to amazon-pay.

mdanilowicz (approved):

PR #2482, opened 2026-06-09
chore: disable amazon pay example

### LINKED ISSUE #2481: Disable the Amazon Pay example because of an audit issue
### Technical TODO

Disable the example as a fix for audit problems

mdanilowicz:
The Amazon Pay button example currently pulls in dependencies that produce a significant number of audit findings. Since this example is not part of the core framework or regular storefront templates, keeping it in the root workspace unnecessarily exposes the main install/build flow to security noise and potential supply-chain risk.

The example should remain available as reference material, but it should not be installed or built automatically from the repository root. Excluding it from the workspace keeps the default development setup safer and cleaner, while still allowing the integration to be reviewed or run manually when needed.

closes #2481


patzick (changes requested):
Let's just copy necessary code from src with adnotaiton and link to your pr or let's do pnpm patch

patzick (approved):

PR #2484, opened 2026-06-09
chore: update PNPM to 11

mdanilowicz:


pnpm-workspace.yaml
```diff
@@ -10,17 +10,43 @@ templates:
...
-- '@vue/test-utils'
-- '@adyen/adyen-web'
+  - '@shopware/*'
+  - '@typescript/native-preview'
+  - 'happy-dom'
+  - '@vue/test-utils'
+  - '@adyen/adyen-web'
 
 auditConfig:
   ignoreGhsas:
```
mkucmus, line 22: `ignoreCves` key must remain so please revert it. the only real GHSA was deleted, so the setting is obsolete.

mkucmus (changes requested):

pnpm-workspace.yaml
mdanilowicz: `ignoreCves` key is removed from PNPM11, `ignoreGhsas` is a main key now

mkucmus: good to know that 👍🏼 then the IDs must be converted into GHSA equivalent because now they won't be matched.

the migration guide says:
```CVE → GHSA. auditConfig.ignoreCves was renamed to auditConfig.ignoreGhsas. Replace each CVE-YYYY-NNNNN entry with the matching GHSA-xxxx-xxxx-xxxx ID (visible in the "More info" column of pnpm audit output).```

and we are good to go 🥇

patzick (approved):

pnpm-workspace.yaml
mdanilowicz: This advisory (GHSA-92fh-27vv-894w / CVE-2025-69874) affects nanotar versions <= 0.2.0. Our lockfile uses nanotar@0.3.0, so the vulnerable version is not present in this repository

PR #2506, opened 2026-06-22
feat(examples): sanity cms integration

mkucmus:
a result of circle day, the example of implementation https://www.sanity.io/

stackblitz preview: https://stackblitz.com/github/shopware/frontends/tree/feat/sanity-cms-example/examples/sanity-cms?file=README.md

image


patzick (commented):
Findings:

1. `examples/sanity-cms` is not covered by the root `pnpm typecheck` CI job because its `package.json` does not define a `typecheck` script. Running `pnpm --filter example-sanity-cms exec nuxi prepare && pnpm --filter example-sanity-cms exec vue-tsc --noEmit` locally fails with real example errors: `app.vue` compares notification type to `\"error\"` even though `useNotifications().pushError()` uses `\"danger\"`; `MiniCart.vue` passes nullable `item.cover` to `getSmallestThumbnailUrl`; and `SectionRichText.vue` passes `unknown[]` to `SanityContent`. CI passes because `nuxt build` does not run Vue template typechecking and Turbo skips packages without a `typecheck` script. Please add a `typecheck` script for this example and fix the reported errors.

2. The `pnpm-lock.yaml` diff is much larger than the Sanity example appears to require. It rewrites peer resolution sets for existing importers, for example moving existing Nuxt entries to `@types/node@25.9.1` / `tsx@4.22.4`, while the visible package changes only add the Sanity example plus two workspace overrides. Please reduce the lockfile churn to the new importer/intentional overrides, or call out why the broad refresh is required.

Verification I ran:
- `pnpm install --frozen-lockfile`
- `pnpm --filter example-sanity-cms lint` passed
- `pnpm --filter example-sanity-cms build` passed
- `pnpm --filter example-sanity-cms exec nuxi prepare && pnpm --filter example-sanity-cms exec vue-tsc --noEmit` failed as described above

mkucmus:
> Findings:
> 
> 1. `examples/sanity-cms` is not covered by the root `pnpm typecheck` CI job because its `package.json` does not define a `typecheck` script. Running `pnpm --filter example-sanity-cms exec nuxi prepare && pnpm --filter example-sanity-cms exec vue-tsc --noEmit` locally fails with real example errors: `app.vue` compares notification type to `\"error\"` even though `useNotifications().pushError()` uses `\"danger\"`; `MiniCart.vue` passes nullable `item.cover` to `getSmallestThumbnailUrl`; and `SectionRichText.vue` passes `unknown[]` to `SanityContent`. CI passes because `nuxt build` does not run Vue template typechecking and Turbo skips packages without a `typecheck` script. Please add a `typecheck` script for this example and fix the reported errors.
> 2. The `pnpm-lock.yaml` diff is much larger than the Sanity example appears to require. It rewrites peer resolution sets for existing importers, for example moving existing Nuxt entries to `@types/node@25.9.1` / `tsx@4.22.4`, while the visible package changes only add the Sanity example plus two workspace overrides. Please reduce the lockfile churn to the new importer/intentional overrides, or call out why the broad refresh is required.
> 
> Verification I ran:
> 
> * `pnpm install --frozen-lockfile`
> * `pnpm --filter example-sanity-cms lint` passed
> * `pnpm --filter example-sanity-cms build` passed
> * `pnpm --filter example-sanity-cms exec nuxi prepare && pnpm --filter example-sanity-cms exec vue-tsc --noEmit` failed as described above

hey, thanks! I've added `typecheck` and then I was able to fix appearing type errors ✅

patzick (approved):

PR #2522, opened 2026-06-30
feat(api-client): multipart content handling

### LINKED ISSUE #1799: Add information about multipart/form-data in api-client
### Describe the feature

the proper information must be displayed once content-type: multipart/form-data is being used in api-client (especially the client side)

### Description

the problem was reported in the community slack https://shopwarecommunity.slack.com/archives/C050L6NCMGQ/p1742555220417229

### Acceptance Criteria

- api-client is updated
- the exception is mentioned in the docs

### Additional information

- [x] Would you be willing to help implement this feature?

### Checks

mkucmus:
### 2. Revert the typecheck command in `vue-starter-template` and `vue-starter-template-extended` back to Nuxt's official typecheck:

Closes #1799

### Problem
The client seeds a default `Content-Type: application/json` on every request. The runtime (browser `fetch` / `undici`) only sets the correct content type for a non-JSON body when **no** `Content-Type` is present, so that default leaked onto uploads and broke them. The previous safeguard only stripped a manually-set `multipart/form-data`, and only in the browser - leaving server-side uploads (the Admin client's main environment) broken.

### What changed
- The client now drops the default `Content-Type` whenever the body is one the runtime must type itself: `FormData`, `Blob`/`File`, `URLSearchParams`, `ArrayBuffer`/typed arrays, and streams.
- Works on **both** the Store and Admin clients, in the browser **and** on the server.
- A `Content-Type` you set explicitly is preserved (e.g. an `image/png` for a typeless `Blob`) - except a boundary-less `multipart/form-data`, which is always replaced.
- Logic extracted into a shared `resolveRequestHeaders` helper used by both clients.
- README: new "Uploading files and other binary bodies" section documenting the behavior.

### Usage
Just pass the body and leave `Content-Type` alone:
```ts
const formData = new FormData();
formData.append("file", file);

await adminApiClient.invoke("uploadV2 post /_action/media/upload", {
  contentType: "multipart/form-data", // type metadata, ignored at runtime
  accept: "application/json",
  body: formData as unknown as { file: Blob },
});


patzick:
Improvement notes from review:

- `resolveRequestHeaders` currently treats only per-request `headers` as explicit. A client-level default such as `client.defaultHeaders.apply({ "Content-Type": "application/octet-stream" })` can still be dropped for `ArrayBuffer`/stream bodies because `callerContentType` only checks `callerHeaders`. Either document that only per-request headers are preserved, or treat non-JSON client defaults as explicit.
- Multipart boundary detection accepts any `boundary=` substring. A stricter check would avoid preserving invalid values such as `multipart/form-data; boundary=`. Worth adding a small test for empty/malformed boundary values.
- The Admin client token refresh request still sends `defaultHeaders` directly and bypasses the new resolver. Normal `/oauth/token` bodies are JSON-like, so this is probably fine, but it is now an exception to the centralized request-header path and could use either a comment or a test to make that intentional.

patzick (approved):

PR #2523, opened 2026-06-30
fix(templates): i18n config resolving

mkucmus:
### Description

Fix the @nuxtjs/i18n v10 config resolution in the vue-starter-template by setting vueI18n to a path relative to the i18n/ dir ("config.ts") instead of an absolute resolve("./config"), which pointed at the wrong directory and broke when the template is consumed as a Nuxt layer.

```[nuxt-i18n 11:25:44 AM]  WARN  Vue I18n configuration file /templates/vue-starter-template/config not found in /templates/vue-starter-template/i18n. Skipping...```


patzick (approved):
Approved. Reviewed the one-line i18n config resolution change and verified the starter template typecheck passes.

PR #2566, opened 2026-07-13
feat(vue-starter-template): URL-driven pagination abstraction for lists

### LINKED ISSUE #2544: Introduce a generic, URL-driven pagination abstraction for lists
Add a generic component for lists collection

mdanilowicz:
closes #2544 

Paginated lists in the starter template each wired up their own data fetching, page/limit state, and pagination UI by hand. This MR introduces a single, reusable abstraction so any list can get URL-driven, SSR-ready pagination with almost no boilerplate — and adopts it in the account orders list (which previously had no pagination controls at all).

**✨ What's new**

- 🧩 useOffsetPaginatedList composable — pass an api-client fetcher(({ page, limit }) => { elements, total }) and get back reactive elements, total, totalPages, currentPage, limit, loading, changePage, refresh. State lives in the URL query (p, limit), so pages are shareable, back/forward works, and the first page is fetched during SSR.
- 🎁 SharedPaginatedList wrapper — thin renderer around the composable, reusing the existing SharedElementsNavigation (pager + size selector). Slots: #default (items), #loading, #empty.
- 💀 AccountOrderLineSkeleton — loading placeholder that mirrors the order card's exact box model to avoid layout shift; the page renders limit skeletons so a full page swaps 1:1.
- 🔻 Dropdown variant prop (BaseDropdown / DropdownField) — new control variant matches neighbouring controls (bordered, white bg, rounded-md, appearance-none + custom chevron) so the page-size select is pixel-consistent with the pagination buttons. Forms (default) are untouched.
- 🌍 i18n — new listing namespace (loading, empty) added for en-GB, de-DE, pl-PL.

**♻️ Changed**

- 🛒 Account orders list now uses SharedPaginatedList (fetches via readOrder with total-count-mode: exact) — adds the previously missing pager, size selector, and loading skeletons.
- 🎨 SharedElementsNavigation layout: page numbers centered, page-size select pinned to the right.

**🔧 Technical notes**

- Filters are intentionally kept outside the wrapper — store them in the route query (or watchSources) and read them inside your fetcher; the query is watched, so filter changes refetch automatically.
- Slot items are typed via #default="{ items }: { items: Schemas['Order'][] }" for reliable typing across vue-tsc/nuxi setups.


templates/vue-starter-template/app/components/shared/PaginatedList.vue
```diff
@@ -0,0 +1,115 @@
...
+defineExpose({ refresh: () => refreshList?.() });
+
+const {
+  elements,
+  total,
+  totalPages,
+  currentPage,
+  limit,
+  loading,
+  isEmpty,
```
patzick, line 52: Could we preserve and render request errors here? `useOffsetPaginatedList` returns `error`, but this wrapper does not consume or expose it. When the fetcher rejects, `loading` becomes false and `elements` falls back to `[]`, so the component renders the empty state and tells the user there are no results. An API, authentication, or network failure should have a distinct error branch/slot, ideally with retry support.

templates/vue-starter-template/app/composables/useOffsetPaginatedList.ts
```diff
@@ -0,0 +1,109 @@
...
+    limitQueryKey = "limit",
+    watchSources = [],
+    scrollTarget,
+  } = options;
+
+  const route = useRoute();
+  const router = useRouter();
+
+  const currentPage = computed(
+    () => toNumber(firstQueryValue(route.query[pageQueryKey])) ?? defaultPage,
```
patzick: Could we normalize and bound URL-derived pagination values? `toNumber` accepts zero, negative, and fractional values, and a stale URL such as `?p=999` can return no elements while `total` is still nonzero. The wrapper then selects its empty branch and hides navigation, leaving no route back to valid results. Please validate positive integers/allowed limits and redirect or clamp pages beyond `totalPages`.

templates/vue-starter-template/app/components/shared/ElementsNavigation.vue
```diff
@@ -30,14 +30,15 @@ function handleChangeSize(size: number) {
...
-  <div class="flex items-center gap-2 justify-between">
+  <div class="relative flex items-center justify-center gap-2">
     <SharedPagination
       @change-page="handleChangePage"
       :total="pages"
       :current="currentPage"
     />
     <SharedSizeSelector
       v-if="showPageSizeSelector"
+      class="absolute right-0"
```
patzick: This absolute positioning does not reserve space for the size selector. With a multi-button pager, the centered pagination controls can overlap the selector at mobile and narrow tablet widths. Because this component is shared, the regression also affects the existing wishlist. Could this use a responsive stack/grid, or otherwise reserve dedicated space for both controls?

patzick (commented):
Three behavioral concerns from review; leaving these as comments only.

templates/vue-starter-template/app/composables/useOffsetPaginatedList.ts
```diff
@@ -0,0 +1,109 @@
...
+  });
+
+  const stateKey = computed(
+    () =>
+      `${key}:${JSON.stringify(route.query)}:${JSON.stringify(
+        watchSources.map((source) => toValue(source)),
+      )}`,
+  );
+
+  const { data, status, error, refresh } = await useAsyncData(
```
patzick: Could we avoid awaiting `useAsyncData` inside this custom composable, or deliberately opt into lazy behavior? With the default `lazy: false`, `SharedPaginatedList` has async setup and is suspended until the initial client request resolves, so its own `#loading` skeleton cannot render for that request; it only becomes useful for later reactive refreshes. Keeping the wrapper synchronous (the async-data refs are available immediately) or using `lazy: true` would let the component render its pending state while retaining SSR data registration. Avoiding the internal await also prevents future context-dependent composables added below it from running after Nuxt/Vue setup context has been unset.

patzick (approved):

PR #2574, opened 2026-07-15
feat(cms-base-layer): add filter by category

### LINKED ISSUE #2560: Support category aggregations for search result listings in the Store API
### Describe the feature

Expose **category aggregations** for search result listings in the Store API.

Currently, the Store API provides aggregations for common product filters such as manufacturers, properties, price, and rating. However, category aggregations are not available for product search results.

As a result, headless storefronts and custom frontends cannot offer category-based filtering on search result pages.

Example:

Search for:

```text
running
```

Expected category aggregation:

```text
Categories

Shoes (248)
Clothing (97)
Accessories (34)
```

Users should be able to refine their search by selecting one or more of these categories while preserving the current search term and other active filters.

This is a common feature in modern e-commerce search experiences and is supported by most major online retailers.

### WHY do we need that?

Category filtering is one of the most common ways users refine search results after entering a broad search term.

Without category aggregations, users can only narrow their search using manufacturers, properties, or other available filters. They cannot quickly limit results to a relevant product category, even though the information is already available in the search result set.

For headless storefronts, there is currently no practical way to implement this functionality using the Store API. Developers would need to build custom aggregation logic or perform additional processing outside of the standard product listing API.

Providing category aggregations would:

* improve the search experience for end users
* enable category-based refinement on search result pages
* reduce the need for custom backend implementations
* provide feature parity between Shopware-powered storefronts and modern e-commerce platforms
* make the Store API more suitable for headless storefront implementations

### Acceptance Criteria

* The Store API exposes category aggregations for product search result listings.
* Returned categories represent the categories of the current search result set.
* Each category contains the corresponding product count.
* Categories can be used as filters in subsequent product listing requests.
* The functionality is available through the Store API for headless storefronts and custom clients.
* Existing API behavior remains backwards compatible.

### Additional information

- [ ] I want to implement this feature.

### Checks

mkucmus:
### Description

closes: #2560

one note: there is a hidden count number next to a category's name due to issue that is already addressed in the backend. I will make sure the fix is provided afterwards.

### Unrelated change included

`examples/sanity-cms/package.json` (and the lockfile) gain `oxlint` and `oxfmt`. That example runs `oxlint . && oxfmt --check .` as its `lint` script but never declared either tool, so it only worked by hoisting. Not part of the category filter - kept here to keep the example's lint script honest.


patzick (changes requested):
# 🔍 Code Review

**Verdict: request changes** — 2 majors. CI is green (all 11 checks pass), no conflicts, `@beta` helpers are well tested.

## 🚫 Majors

**1. `SwFilterChips` crashes when `filters.categories` is absent** — `packages/cms-base-layer/app/components/SwFilterChips.vue:9,92`
`FilterState.categories` is a new *required* field and `Array.from(props.filters.categories)` runs unguarded. Any consumer of the published component that builds its own filter-state object (not via `useSelectedListingFilters`) now gets `TypeError: undefined is not iterable` and a dead chips row — on a `minor` bump. Fix: `categories?: Set<string>` + `Array.from(props.filters.categories ?? [])`.

**2. Category filter renders on non-search listings but does nothing** — `SwProductListingFilters.vue:66,88,159` and `SwProductListingFiltersHorizontal.vue:65,87,158`
All three write paths are gated on `isProductSearch` (post-filter, `?categories=` param, reset button), but the render loop at `SwProductListingFilters.vue:284` is not. On a category listing whose criteria include the category aggregations, checking a box mutates the set, `executeSearch` pushes a query without `categories`, `applyQueryToFilters` clears the set again → the checkbox flips back and the result set never changes. Either gate the render (skip `code === "categories"` when `!isProductSearch`) or drop the gates and let the post-filter apply everywhere.

## ⚠️ Minors

- `getListingFilters` silently changes shape for existing `categories` consumers — entities are reordered by count, `count` is injected, `apiAlias` is dropped, and a `categories-counts` aggregation is now omitted from the filter list (`packages/helpers/src/listing/filters.ts:61,114`). It is `@beta`, but the changeset should say so.
- Root-category hiding is duplicated in `SwFilterCategories.vue:60` and `templates/vue-demo-store/.../ListingFilter.vue:57`. Better as an opt-in helper next to `getCategoryFilterAggregations`.
- `examples/sanity-cms/package.json` + lockfile add `oxfmt`/`oxlint`. It fixes a genuinely undeclared dep for that example's `lint` script, but it is unrelated to this feature and unmentioned in the description — split it out or note it.
- `templates/vue-demo-store/app/pages/search.vue:157`: `p: route.query.p ? Number(route.query.p) : 1` skips the `firstQueryValue`/`toNumber` pair used two lines above; `?p=1&p=2` sends `p: NaN`.
- `CATEGORY_PARENTS_AGGREGATION_NAME` is not re-exported from `packages/helpers/src/listing/index.ts`, so consumers following the new doc section must hardcode `"categories-parents"`.
- No component-level test for `SwFilterCategories` or for the `isProductSearch` gating — behavioral logic landed in three components with helper-level tests only.

## 💡 Nits

- `SwFilterCategories.vue:139`: `:key="`${option.id}-${isChecked(option.id)}`"` tears down and recreates each label on every toggle; `:key="option.id"` suffices since `SwCheckbox` is reactively bound.
- `SwFilterChips.vue:93`: `availableFilters.find(...)` is re-run per selected category inside the loop; hoist it (same pre-existing pattern for manufacturer).
- `vue-starter-template` has no `listing.categories` i18n entry, so the heading reads "Categories" in de/pl. Consistent with the existing gap for `listing.filters` there, so only worth knowing.

## ✅ Good

- Post-filter choice for faceted behavior is correct and well documented, including the ES nested-aggregation caveat.
- `resolveCategoryBucketCount` variant collapsing and the `getListingFilters` merge are covered by focused unit tests, including the missing-counts and counts-without-entities paths.
- Removing the `search()` / `changeCurrentPage()` double-fetch in favor of the `useAsyncData` `watch: [() => route.query]` refetch is a real fix; the demo-store's `limit` still round-trips via `v-model` + query.

## 🔁 Inherited feedback

Both Copilot inline comments (`string[]` query params in `search.vue` and `ListingFilters.vue`) are already addressed — the current head uses `firstQueryValue(...)` in `search.vue:148` and `ListingFilters.vue:65`.

## 🧪 Validation

Not run locally by request — verdict is based on CI: `Build`, `Test`, `lint`, `Check audit`, CodeQL and all Vercel deployments pass on `fb3205c`.

apps/docs/src/guides/e-commerce/product-listing.md
```diff
@@ -365,6 +365,105 @@ const ColorFilter: ListingFiler = {
...
+  entities: [
+    { id: "...", name: "Shoes", count: 248, /* other category props */ },
+    { id: "...", name: "Clothing", count: 97 },
+  ],
+}
+```
+
+The entities are sorted by `count`, highest first. Note that `label` is the raw aggregation name, so give the filter your own translated heading when you render it.
+
+On the `/store-api/search` route the `count` counts every matching **variant**, not every product. The filter UIs in this repo sort by it but do not display it. The `cms-base-layer` filters show no counts at all, and the demo store hides this one because its manufacturer and property filters do collapse variants, so an uncollapsed category count next to them would read as a bug.
```
patzick, line 400: Nit (inherited from Copilot): "The filter UIs in this repo sort by it" points at the wrong place — the sort happens in `getListingFilters`, so the entities already arrive sorted by `count` and the UIs just render them in order. Suggest "The entities arrive sorted by it, but no UI in this repo displays it."

patzick (approved):
# 🔍 Re-review: Code Review

**Verdict: approve.** Both majors from the previous round are fixed and every minor/nit was picked up. CI is green on `effb7961` (all 11 checks), no conflicts.

## ✅ Resolved

- **Major 1** — `SwFilterChips`: `categories` is now optional and `?? []`-guarded, so a pre-existing filter-state object no longer throws.
- **Major 2** — new `getVisibleListingFilters` drops the `categories` filter on non-search listings, applied to both the render loop and the chips in `SwProductListingFilters` / `SwProductListingFiltersHorizontal`, with unit tests for all four branches. `isProductSearch` is a plain boolean here, so the object literal passes the real value — no ref-truthiness trap.
- Root-category hiding extracted to `excludeRootCategory()` and reused in both components; `CATEGORY_*_AGGREGATION_NAME` re-exported and documented with a copy-pasteable example; `search.vue` now routes `search`/`p`/`order` through `firstQueryValue`/`toNumber`; `:key` reduced to `option.id`; the `find()` hoisted out of both chip loops; starter-template `listing.categories` added for en/de/pl; the changeset now spells out the `getListingFilters` shape change and the search-only scope; the `oxlint`/`oxfmt` addition is explained in the description.

## 💡 Nits (non-blocking)

- One inline comment on the docs wording (inherited from Copilot's suppressed comment).
- Still no component-level test for `SwFilterCategories` itself — the helper-level coverage is good enough to ship, worth a follow-up.

## 🧪 Validation

Not run locally by request — verdict is based on CI: `Build`, `Test`, `lint`, `Check audit`, CodeQL and all Vercel deployments pass on the head commit.

PR #2596, opened 2026-08-03
fix: quiet remaining Rolldown and Nuxt build warnings

patzick:
### Description

Reduce leftover Rolldown/Vite/Nitro build noise after the Lightning CSS `@apply` work: scope Nuxt `components.global` to string-resolved components, keep UnoCSS runtime free of full carbon icon JSON, bump `@vueuse` to 14.4.0, and accept intentional large async chunks (three.js / Braintree).

### ToDo's

- [ ] Documentation added/updated
- [ ] E2E-Tests added/updated
- [ ] Related Issue updated

### Additional context

`INEFFECTIVE_DYNAMIC_IMPORT` came from `components.global: true` wrapping static imports; express-checkout also double-imported `App`. Nitro may still warn about `@__NO_SIDE_EFFECTS__` inside `@vueuse/shared` JSDoc; that is filtered narrowly in cms-base-layer until upstream fixes the annotation placement.

Made with [Cursor](https://cursor.com)


mkucmus:
one observation so far is that vue-starter-template-extended has a top menu and footer sections missing 👀 
image

patzick:
Good catch — confirmed and fixed.

Root cause: Nuxt’s `scanComponents` records each scanned dir path and skips later files under that path. Our attempt to register `Frontend*.vue` as `global: true` and everything else as non-global used the **same** `components/` path twice, so only the `Frontend*` scan stuck and Layout/Header/Footer (etc.) never registered. That’s why the extended template lost the top menu and footer.

Fix: move `resolveComponent` targets into `components/global` (distinct path, `pathPrefix: false`) and keep the rest of `components/` as a normal non-global scan. Same approach for `vue-demo-store` (`components/errors` stays a separate global dir for `ErrorsRoutingNotFound` / `ErrorsMaintainMode`).

mkucmus:
cool! looks as it should be now :)

mkucmus (approved):
quite a big fix 👍🏼 

before merging: please make sure if our AGENTS, *.md files and corresponding docs are up to date with the last change as some files were moved (components to global).

patzick:
Addressed the docs follow-up from the approval review.

Updated `AGENTS.md`, `packages/cms-base-layer/AGENTS.md`, template READMEs, and the Vue Starter / CMS docs to describe:

- `app/components/global/` for `Frontend*` SEO page resolvers
- `app/components/cms/` for CMS block/element overrides (also global, `pathPrefix: false`)
- remaining `app/components/**` as normal auto-imports

Also wired `components/cms` as its own global dir in the starter/demo `nuxt.config.ts` so overrides keep resolving via `resolveComponent` with the correct `Cms*` names.

PR #2597, opened 2026-08-03
fix(templates): skip checkout method icons without media

mkucmus:
### Description

Shipping and payment methods without an assigned icon rendered a broken image in the checkout.

`getShippingMethodIcon()` and `getPaymentMethodIcon()` return an empty string when the method has no media. That empty string was passed straight to `NuxtImg`, which still built its density srcset from it. The result was ``. A browser resolves an empty `src` against the current document, so it requested the checkout page itself as an image and showed the broken image placeholder.

The fix renders the icon only when the method actually has media. This matches how `vue-demo-store` already handles it in `app/pages/checkout/index.vue`.

This is not a missing association. Shopware core adds the media association server side in both `ShippingMethodRoute` and `PaymentMethodRoute`, so the data is already requested correctly. The demo store's "Standard" shipping method simply has `mediaId: null`. Any merchant can leave a method icon empty, so the frontend has to handle it.

Changed files:

- `templates/vue-starter-template/app/components/checkout/ShippingMethods.vue`
- `templates/vue-starter-template/app/components/checkout/PaymentMethods.vue`

`vue-starter-template-extended` needs no change. It extends `vue-starter-template` as a Nuxt layer and uses these same components, so it picks up the fix automatically.

### Type of change

image

Before: the shipping method row shows a broken image placeholder. DevTools shows `srcset="1x, 2x"` with an empty `src`.

After: no image element is rendered when the method has no icon. Methods that do have an icon are unchanged.

### Additional context

Verified against both demo sales channels with and without an explicit `media` association:

| endpoint | request body | `mediaId` | `media` |
| --- | --- | --- | --- |
| `/payment-method` | `{}` | `b5da4638...` | full object with `url` |
| `/payment-method` | `{"associations":{"media":{}}}` | `b5da4638...` | full object with `url` |
| `/shipping-method` | `{}` | `null` | `null` |
| `/shipping-method` | `{"associations":{"media":{}}}` | `null` | `null` |

Payment method media arrives fully loaded with no associations requested. Adding the association to the shipping request changes nothing.

The public helpers keep returning an empty string. That is their documented and unit tested contract in `@shopware/helpers`, so changing it to `undefined` would be a breaking change for a problem the call sites can handle.

`oxlint` and `oxfmt --check` pass on both changed files.

### DIFF
--- a/templates/vue-starter-template/app/components/checkout/PaymentMethods.vue
+++ b/templates/vue-starter-template/app/components/checkout/PaymentMethods.vue
@@ -52,7 +52,7 @@ function handleChange(id: string) {
             {{ paymentMethod.description }}
           </div>
         </div>
-        <div class="ml-auto">
+        <div v-if="getPaymentMethodIcon(paymentMethod)" class="ml-auto">
           <NuxtImg
             :src="getPaymentMethodIcon(paymentMethod)"
             height="32"
--- a/templates/vue-starter-template/app/components/checkout/ShippingMethods.vue
+++ b/templates/vue-starter-template/app/components/checkout/ShippingMethods.vue
@@ -54,7 +54,7 @@ function handleChange(id: string) {
             {{ getShippingMethodDeliveryTime(shippingMethod) }}
           </div>
         </div>
-        <div class="ml-auto">
+        <div v-if="getShippingMethodIcon(shippingMethod)" class="ml-auto">
           <NuxtImg
             :src="getShippingMethodIcon(shippingMethod)"
             height="32"


patzick:
🤖 `om-auto-review-pr` started by @patzick at 2026-08-03T13:20:14Z. Other auto-skills will skip this PR until the lock is released. (Label operations are disabled for this run by explicit user instruction, so the claim is the assignee plus this comment.)

patzick (approved):
# 🔍 Code Review

**PR:** #2597 — `fix(templates): skip checkout method icons without media`
**Head reviewed:** `9f59983d55cd5ba97f57073cef80b444fad57c77`
**Base:** `main`
**Mode:** first review (no earlier review by this reviewer on this PR)

## 🎯 Summary

This PR fixes a real, user-visible rendering defect in the `vue-starter-template` checkout: shipping and payment methods that have no assigned media rendered an `` with an empty `src`, which a browser resolves against the current document, so the checkout page requested itself as an image and painted the broken-image placeholder. The change wraps each icon container in a `v-if` guard on the icon helper's return value, so the `NuxtImg` is only created when the method actually has media.

I verified the causal chain in the code rather than taking the description on trust. `getShippingMethodIcon()` and `getPaymentMethodIcon()` (`packages/helpers/src/checkout/getShippingMethodIcon.ts`, `packages/helpers/src/checkout/getPaymentMethodIcon.ts`) both end in `return method.media?.url ?? ""`, and both have unit tests that pin the empty string as their documented contract (`getShippingMethodIcon.test.ts`, `getPaymentMethodIcon.test.ts`). That empty string was flowing straight into `:src`, so the diagnosis and the decision to fix it at the call site rather than change the helper's return type are both correct: changing `""` to `undefined` would break a tested public contract of `@shopware/helpers` for a problem two call sites can handle in one line each.

I also confirmed the two secondary claims in the description. `templates/vue-starter-template-extended` genuinely needs no change — its `app/` directory contains only `app.config.ts` and it holds no override of either component, so it inherits the fix through the Nuxt layer. And the fix's shape is consistent with `vue-demo-store`, which guards the same markup at `templates/vue-demo-store/app/pages/checkout/index.vue:780` and `:843`, although it guards on `media?.url` rather than on the helper (see the nit below).

As a side benefit the change is also a small accessibility improvement: the removed element previously exposed a broken image whose `alt` merely duplicated the method name that is already rendered as text next to it.

**Verdict: ✅ Approve.** The change is minimal, correctly scoped, fixes the defect at the right layer, and preserves the public helper contract. Nothing in the diff is a blocker or a major. Three minors and three nits are listed below for the author to pick up at their discretion — none of them needs to hold this PR.

## 🧪 Validation Gate

The gate was run inside an isolated worktree checked out at the PR head, after a clean `pnpm i --frozen-lockfile`.

| Command | Scope | Result |
| --- | --- | --- |
| `pnpm i --frozen-lockfile` | monorepo | ✅ pass — lockfile is in sync, package prebuild succeeded (6/6 tasks) |
| `oxlint` | the two changed files | ✅ pass — no diagnostics |
| `oxfmt --check` | the two changed files | ✅ pass — both files already correctly formatted |
| `pnpm run typecheck --filter=vue-starter-template` | `vue-starter-template` and its dependency graph | ✅ pass — 8/8 turbo tasks successful, `nuxi typecheck` clean |
| `pnpm run test --filter=@shopware/helpers` | `@shopware/helpers` | ✅ pass — the icon-helper suites that pin the `""` contract still pass, confirming the fix did not need a helper change |
| Full monorepo `pnpm run test` | monorepo | ⏭️ delegated to CI — the `Test`, `Build`, `lint`, `Analyze`, and `CodeQL` checks are all green on this exact head SHA, so the unscoped run was not repeated locally |

## 🔍 Findings

No blockers. No majors.

### Minor

- `templates/vue-starter-template/app/components/checkout/ProductTile.vue:18` — The identical defect this PR fixes still exists one file over, in the same checkout flow. `cover` is computed as `item.cover ? getSmallestThumbnailUrl(item.cover) : ""`, and that value is passed straight to `NuxtImg` at line 25. Any cart line item without a cover — a promotion line, a custom line item, or a product with no media — therefore renders exactly the `` that this PR is removing two components away, on the same checkout page. It is a pre-existing bug and not something this diff introduced, so I am not blocking on it, but the one-line guard is the same guard you already wrote here. Either fold it into this PR while the context is fresh, or open a follow-up so the class of defect is closed rather than half-closed.

- `templates/vue-starter-template/app/components/wishlist/ProductTile.vue:47`, `templates/vue-starter-template/app/components/account/order/Product.vue:18`, `templates/vue-starter-template/app/components/account/order/LineItemProduct.vue:35` — The same unguarded pattern appears in three more places, though in a milder form: `getSmallestThumbnailUrl()` returns `string | undefined`, so a missing cover yields `:src="undefined"`, Vue drops the attribute, and no bogus request is made. These are worth sweeping in the same follow-up as the finding above, because a reader cannot tell from the call site which of the two helper contracts (`""` or `undefined`) is in play, and that ambiguity is precisely what produced this bug.

- No changeset accompanies the fix. `vue-starter-template` is a changesets-versioned package in this repo — `.changeset/smooth-starter-breadcrumbs.md` bumps `"vue-starter-template": patch` for an equivalent template-only change — and `.github/pull_request_template.md` asks for `pnpm run changeset`. I checked recent history before raising this, and practice is genuinely inconsistent for template-only PRs: #2594 shipped a changeset, while #2566, #2576, and #2563 did not, and no CI job enforces it. That is why this is a minor rather than a major. Since this is a user-visible bug fix in a released template, a `patch` changeset would be the right call.

### Nit

- `templates/vue-starter-template/app/components/checkout/ShippingMethods.vue:57` and `templates/vue-starter-template/app/components/checkout/PaymentMethods.vue:55` — The PR description says the fix "matches how `vue-demo-store` already handles it". The two are equivalent in effect but not identical in form: `vue-demo-store` guards on `singleShippingMethod.media?.url` / `singlePaymentMethod.media?.url` (`templates/vue-demo-store/app/pages/checkout/index.vue:780` and `:843`), whereas this PR guards on the helper's return value. That difference matters for the inherited review comments below, so it is worth stating accurately rather than as a straight match.

- Two inherited nits from the Copilot reviewer are recorded in the inherited-feedback section below; both are declined with a reason, and neither affects the verdict.

## 🔍 Inherited reviewer feedback (2 asks already on the PR)

- **@Copilot** on `templates/vue-starter-template/app/components/checkout/ShippingMethods.vue:59` ([discussion_r3704119579](https://github.com/shopware/frontends/pull/2597#discussion_r3704119579)) — Suggests avoiding the duplicate `getShippingMethodIcon(shippingMethod)` call by checking the underlying `media`/`url` in `v-if` and calling the helper only for `:src`, both to skip duplicate work and to "guard against future helper changes". → **Declined, with reason.** The duplicate work is a single optional-chained property read plus a `??`, evaluated once per method row in a list that has a handful of entries; there is nothing measurable to reclaim, and hoisting it would mean introducing a computed map or a wrapper component for a property access. More importantly, the second half of the rationale points the other way: guarding on the exact expression that feeds `:src` is what keeps the guard and the rendered value from ever diverging. Guarding on `media?.url` while rendering `getShippingMethodIcon()` is the form that would silently break if the helper's logic changed — for instance if it ever gained a placeholder fallback. Marked as a nit, not carried as an actionable finding.

- **@Copilot** on `templates/vue-starter-template/app/components/checkout/PaymentMethods.vue:57` ([discussion_r3704119624](https://github.com/shopware/frontends/pull/2597#discussion_r3704119624)) — The same suggestion for `getPaymentMethodIcon(paymentMethod)`. → **Declined, with reason** — identical reasoning to the finding above; the two call sites are structurally the same.

No other unaddressed reviewer feedback was on the PR: the Copilot review body itself is a descriptive overview with no ask, and the conversation thread carried no comments at the time of this review.

## 💥 Breaking-Changes checklist

- **Public exports and APIs:** unchanged. The diff touches only two `.vue` files inside a `private: true` template; no file under `packages/` is modified, and `getShippingMethodIcon` / `getPaymentMethodIcon` keep their signatures, their `""` return value, and their tests. This is exactly the right call — the description's reasoning that changing the helpers to `undefined` would be the breaking option is correct.
- **HTTP routes, request shapes, and response shapes:** unchanged. No store-API call, criteria, or association is added or removed, which matches the author's finding that Shopware core already ships the media association in both `ShippingMethodRoute` and `PaymentMethodRoute`.
- **Event names, CLI flags, config formats, DB schema:** not touched.
- **Rendered contract for consumers:** the only observable change is that the `<div class="ml-auto">` icon wrapper is now absent instead of holding a broken image when a method has no media. Nothing else in the row depends on that element — `ml-auto` is a self-applied utility, and the surrounding `label` is a plain `flex items-center gap-4` container — so the layout of the remaining content is unaffected. A downstream project that had CSS-selected that always-present wrapper would see it disappear, but it only ever contained a broken image in that state, so this is the intended correction rather than a regression.
- **`BACKWARD_COMPATIBILITY.md`:** no such file exists at the repository root, so there is no protected-surface list to check this diff against. Nothing here would have violated one.

**No breaking changes. No warning needed.**

## 🧪 Test Coverage

No test files appear in this diff, and I want to be explicit about why that is acceptable here rather than waving it through. `vue-starter-template` has no unit-test harness at all: its `package.json` defines no `test` script, there is no `vitest.config.*` in the template, and the template contains no `*.test.*` or `*.spec.*` files anywhere. There is therefore no existing place to put a component test for this behaviour, and adding a Vitest plus `@vue/test-utils` setup to the template would be a substantially larger change than the one-line fix it would be testing — well outside this PR's scope.

The behaviour underneath the fix is already covered where it can be: `getShippingMethodIcon.test.ts` and `getPaymentMethodIcon.test.ts` both assert the empty-string return for a method with no media, which is the exact input this guard now handles, and both suites pass on this head. The author's manual verification is also unusually thorough — the before/after screenshots, the DevTools evidence of `srcset="1x, 2x"` with an empty `src`, and the endpoint-by-endpoint table showing that adding the `media` association changes nothing all substantiate the diagnosis.

If the team wants regression protection for this, the natural home is the Playwright suite that already walks the checkout (`apps/e2e-tests/tests/createOrder.spec.ts`): an assertion that no `img` on the checkout page has an empty `src` would catch this defect and the adjacent `ProductTile.vue` instance in the minor findings above, in one check. That is a follow-up suggestion, not a condition of this approval.

patzick:
🤖 `om-auto-review-pr` completed: APPROVED. Lock released.

The review passed with no blockers and no majors. The fix correctly resolves the empty-`src` icon rendering at the call sites instead of changing the tested `""` return contract of `getShippingMethodIcon()` / `getPaymentMethodIcon()` in `@shopware/helpers`, and the claims about `vue-starter-template-extended` needing no change and about `vue-demo-store` already guarding this markup both check out against the code. Validation ran in an isolated worktree at head `9f59983d`: `pnpm i --frozen-lockfile`, `oxlint` and `oxfmt --check` on both changed files, `pnpm run typecheck --filter=vue-starter-template` (8/8 tasks), and `pnpm run test --filter=@shopware/helpers` all pass, and every CI check is green on this SHA.

Three minors and three nits are recorded in the review for the author to take or leave: the same empty-`src` defect still exists at `templates/vue-starter-template/app/components/checkout/ProductTile.vue:18` in the same checkout flow (pre-existing, worth a follow-up or a one-line addition here), three milder `getSmallestThumbnailUrl()` call sites share the pattern, and a `patch` changeset for `vue-starter-template` would be appropriate for a user-visible bug fix. Both inherited @Copilot nits about the duplicated helper call are declined with a stated reason — guarding on the exact expression that feeds `:src` is what keeps the guard and the rendered value from diverging.

autofix: skipped (not my PR — re-run with --autofix to fix it here). No label operations were performed: the invoking user explicitly disabled labels for this run, so no pipeline, priority, or risk label was added, changed, or removed. The reviewer assignee added for the claim has been removed again, leaving the PR assignees as they were found.

patzick (approved):

PR #2600, opened 2026-08-03
fix(api-client): ignore context token from public cache responses

patzick:
### Description

CDN-cached `cacheableReads` Store API GETs (`Cache-Control: public`) can replay a stale guest `sw-context-token`. The API client was adopting that token and overwriting the logged-in session, so navigating to `/account` sent `account/newsletter-recipient` with the wrong token and logged the user out.

This change ignores `sw-context-token` updates from publicly cacheable responses. Session-changing routes (login/logout/register/context) remain `private` and still rotate the token.

### Type of change

### ToDo's

### Additional context

Reproduced against the demo Store API: a `GET /language` cache HIT returned a different `sw-context-token` than the one sent on the request. With `cacheableReads: true` in the vue-starter-template, that token was written to the cookie before account requests ran.

Made with [Cursor](https://cursor.com)


apps/docs/src/best-practices/caching.md
mkucmus:
```suggestion
This is where `cacheableReads` pays off. By switching reads to GET with a deterministic `_criteria` URL, the request layer produces cacheable requests; the backend reverse proxy then applies `Cache-Control`, cache tags, `sw-cache-hash`, and invalidation. None of that is handled by the frontend `@shopware/api-client` - it only forwards Shopware Store API headers (`sw-access-key`, `sw-context-token`, `sw-language-id`, and so on) and refreshes the context token from non-public response headers. It reads `Cache-Control` only to ignore `sw-context-token` on publicly cacheable responses; it does not set `Cache-Control` or handle cache tags or `sw-cache-hash`.
```

apps/docs/src/best-practices/caching.md
```diff
@@ -260,7 +260,7 @@ Beyond HTTP, the storefront avoids redundant work by sharing in-memory state acr
...
 - **Session, cart, user, listings.** State is held in shared refs under named injection keys (`swSessionContext`, `swCart`, `customer`, `swNavigation-${type}`, listing keys) via a `useContext` helper built on VueUse's `provideLocal`/`injectLocal`. Mutations reassign the shared ref so every consumer sees consistent state without refetching.
-- **Single API client.** One Store API client is created per app and provided via `provide`/`inject`, so all composables share its `defaultHeaders` - including the `sw-context-token` it captures from response headers and reuses on subsequent requests.
+- **Single API client.** One Store API client is created per app and provided via `provide`/`inject`, so all composables share its `defaultHeaders` - including the `sw-context-token` it captures from private response headers and reuses on subsequent requests. Tokens on `Cache-Control: public` responses are ignored, so a CDN hit cannot overwrite a logged-in session with a stale guest token.
```
mkucmus:
```suggestion
- **Single API client.** One Store API client is created per app and provided via `provide`/`inject`, so all composables share its `defaultHeaders` - including the `sw-context-token` it captures from non-public (private or uncached) response headers and reuses on subsequent requests. Tokens on `Cache-Control: public` responses are ignored, so a CDN hit cannot overwrite a logged-in session with a stale guest token.
```

packages/api-client/src/createApiClient.test.ts
```diff
@@ -178,6 +178,75 @@ describe("createAPIClient", () => {
...
+    // still carry a guest sw-context-token from when the entry was stored.
+    // Adopting that token would log the user out.
+    const app = createApp().use(
+      "/language",
+      eventHandler(async (event) => {
+        setHeader(event, "sw-context-token", "cached-guest-token");
+        setHeader(
+          event,
+          "cache-control",
+          "max-age=0, public, s-maxage=1800, stale-while-revalidate=86400",
```
mkucmus: with a lowercase fixture the flag is never exercised

```suggestion
          "max-age=0, Public, s-maxage=1800, stale-while-revalidate=86400",
```

mkucmus (changes requested):
Solid fix and correctly scoped - the mechanism is right, the regression test genuinely fails on `main`, guest session bootstrap via `/context` (private) is unaffected, and the admin client needs no matching guard. Requesting changes only for two small things:

1. **`apps/docs/src/best-practices/caching.md` line 245** - this line still says the client "does not read or set `Cache-Control`" and refreshes the token unconditionally. After this PR both claims are outdated, so the page contradicts the updated "Single API client" bullet below. Please reword it to match the new behavior (inline suggestion attached).

2. **`createApiClient.test.ts` line 192** - use `Public` (title-case) in the cache-control fixture. Directive names are case-insensitive (RFC 9111) and only the `/i` flag covers that; a lowercase fixture never exercises the flag, so dropping it in a refactor would silently bring the logout bug back with a green suite (inline suggestion attached).

Optional nitpick: in the updated bullet (line 263), "captures from private response headers" over-narrows the rule - responses with no `Cache-Control` header also update the token. "non-public (private or uncached)" would be precise.

Note for the record, no change requested: the guard intentionally matches only the `public` directive. That's fine - Shopware can only emit `s-maxage` together with `public` (Symfony's `setSharedMaxAge()` calls `setPublic()`). If anyone hardens this later, matching bare `s-maxage` would break login, since the private policy is `private, no-cache, max-age=0, s-maxage=0` - the second test here pins exactly that header. Any extended check must exclude `private` first.

patzick:
Addressed the review feedback:

1. Reworded the backend HTTP cache paragraph so it no longer claims the client ignores `Cache-Control` / refreshes the token unconditionally.
2. Updated the Single API client bullet to say non-public (private or uncached).
3. Changed the public-cache test fixture to `Public` so the `/i` flag is exercised.

mkucmus (approved):
Image

PR #2609, opened 2026-08-05
refactor(cms-base-layer)!: make the package a plain Nuxt layer

### LINKED ISSUE #2163: Refactor cms-base-layer to be officially a nuxt layer
Reflect the actual state of the package.

### Describe the feature

- update package.json
- refactor module structure
- cleanup tsconfig.json and types augmentation

### Description

--

### Acceptance Criteria

_No response_

### Additional information

- [x] Would you be willing to help implement this feature?

### Checks

mkucmus:
### Description

Makes `@shopware/cms-base-layer` a plain Nuxt layer. Closes #2163

- Removed the dead Nuxt module (`src/index.ts`, `build.config.ts`, `index.cjs`, `dist`). It registered a component folder that moved to `app/` long ago, so it could not work.
- Dropped the `build` and `dev` scripts and the `unbuild` dev dependency. Nothing is left to bundle.
- Fixed `files` (listed `helpers` and `app.config.ts` that do not exist) and added an `exports` map.
- Deleted the stale `components.d.ts` and two vitest aliases pointing at missing files.
- Updated README and AGENTS.md to match.

**Breaking:** the `exports` map blocks deep imports like `@shopware/cms-base-layer/app/components/SwProductCard.vue`, and `index.cjs`/`dist` are no longer published. Consume the package with `extends: ["@shopware/cms-base-layer"]`.

### Type of change

Breaking change (fix or feature that would cause existing functionality to not work as expected)

### ToDo's

- [x] Changeset file provided (major bump)

### Additional context

`types/imports.d.ts` keeps resolving `@shopware/composables` from the package root. An earlier revision of this PR pointed it at `@shopware/composables/lib`, which checked the layer against `dist/index.d.mts` while consumers resolve `exports["."]` to `src/index.ts`. That was reverted after review: the two surfaces can drift, it needs `packages/composables/dist` to exist, and it fails silently when it does not, because `skipLibCheck` hides a broken `export *` inside a `.d.ts`.

This branch also carried a turbo change that cached `.nuxt` as a build output. That is a latent CI issue on `main` rather than something this refactor causes, so it moved to its own PR.

`packages/api-client` gains `h3` as a dev dependency. Unrelated to the layer refactor, but merging `main` exposed it: those tests import `h3` without declaring it, and the merge flipped the hoisted version from 1.15.11 to 2.0.1-rc.26, which drops the `App` type and broke the Test job. Declaring it makes the resolution deterministic. No changeset, as it is test-only.

Verified: layer tests/lint/typecheck, repo-wide build (35 tasks) and typecheck (69 tasks), `vue-demo-store` vue-tsc and production build, CI build command, and `npm pack` diff vs main (only `index.cjs` leaves the tarball).


patzick (changes requested):
## 🔍 Code Review

Solid, well-evidenced refactor — the `main` + `exports` + no-build shape now matches the sibling `@shopware/unocss-design-tokens-layer`, the deletions check out (no deep imports of `@shopware/cms-base-layer/...` anywhere in the repo, no `dataTestid`/`AllowedComponentProps` consumers, the two removed vitest aliases really did point at non-existent files, `check-unused-colors.ts` had no remaining references), and the major changeset documents the breaks. All required checks are green.

Two things to resolve first — details inline:

### 🚨 Major

- **`types/imports.d.ts:5`** — switching the `#imports` shim to `@shopware/composables/lib` makes the layer type-check against the *rolled-up `dist` d.ts* while every consumer resolves `@shopware/composables` → `src/index.ts`. Two type surfaces that can drift, and a fresh checkout can no longer type-check this package standalone — which the Quick Start added in this same PR now advertises. (Also raised by Copilot as a suppressed comment.)

### 🧹 Minor

- **`turbo.json:8`** — the `vue-starter-template#build` override is unrelated to "make cms-base-layer a plain Nuxt layer", is not mentioned in the description or changeset, and caches `.nuxt/**` (a `nuxt prepare` artifact) as a build output.

### ✅ Verification

No commands were run locally for this review; the verdict is based on the diff, the repo state at the PR head, and the CI results (Build, Test, lint, Analyze, Check audit — all passing).

packages/cms-base-layer/types/imports.d.ts
```diff
@@ -1,2 +1,6 @@
-export * from "@shopware/composables";
+// `#imports` shim for type-checking this layer on its own. Nuxt only generates its
+// own auto-imports into `.nuxt/imports.d.ts`, so the @shopware/composables ones the
+// components rely on have to be added here. Built types, not `src`, so this does not
+// pull the composables sources into the program.
+export * from "@shopware/composables/lib";
```
patzick: **Major** — `@shopware/composables/lib` maps to `./dist/index.d.mts`, but every consumer of `@shopware/composables` resolves `exports["."].types` → `./src/index.ts`. So this layer would be validated against the unbuild dts rollup while its consumers see the source types; the two surfaces can diverge and a break would only surface downstream.

It also makes the package non-typecheckable standalone: on a fresh checkout `packages/composables/dist` does not exist, so `cd packages/cms-base-layer && pnpm run typecheck` fails until `@shopware/composables` is built. `turbo typecheck` masks this via `dependsOn: ["build", "^build"]`, but AGENTS.md:21 (added in this PR) presents the bare command as a working entry point.

Suggestions, in order of preference: keep `export * from "@shopware/composables"` and solve the "src in the program" cost via the project-references layout from #2606; or, if `lib` is deliberate, make it self-sufficient (`"typecheck": "pnpm --filter @shopware/composables build && pnpm nuxt prepare && tsgo --noEmit"`) and say in the comment why the dist surface is preferred over the one consumers resolve.

turbo.json
```diff
@@ -5,6 +5,10 @@
       "dependsOn": ["^build"],
       "outputs": ["dist/**", ".output/**", ".vercel/**", ".vitepress/dist/**"]
     },
+    "vue-starter-template#build": {
+      "dependsOn": ["^build"],
+      "outputs": [".nuxt/**", "dist/**", ".output/**", ".vercel/**"]
```
patzick: **Minor** — two concerns with this override.

Scope: it is unrelated to making `cms-base-layer` a plain Nuxt layer, and it is in neither the description nor the changeset, so it will not be obvious later why a breaking packaging PR changed the starter template's build caching. Worth splitting into its own PR, or at least documenting.

Mechanism: `.nuxt` is a `nuxt prepare` artifact, not a build output, and it contains generated tsconfig/type files. Declaring it a cached output means a cache hit can restore it instead of regenerating it — with remote cache that means one machine's generated paths landing in another's build. If `vue-starter-template-extended` needs the base template's tsconfig, having its build depend on a `prepare`/`postinstall` task that regenerates `.nuxt` is more predictable than caching it.

packages/cms-base-layer/types/imports.d.ts
mkucmus: Went with the first suggestion: the shim now re-exports `@shopware/composables` (the package root) instead of `@shopware/composables/lib`.

That is the same entry consumers resolve, and its `exports["."].types` points at `./src/index.ts`, so the layer is validated against the source surface apps actually see. The dist/src divergence you described cannot happen now.

This also fixes the standalone typecheck. Verified by moving `packages/composables/dist` out of the way and running `cd packages/cms-base-layer && pnpm run typecheck` on its own: exits 0, no build of `@shopware/composables` needed. So the bare command in AGENTS.md holds on a fresh checkout.

The reasoning is recorded in the comment above the export in `types/imports.d.ts`.

packages/cms-base-layer/package.json
```diff
@@ -67,7 +65,6 @@
     "nuxt": "4.5.1",
     "tsconfig": "workspace:*",
     "typescript": "5.9.3",
```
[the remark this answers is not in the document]
mkucmus: Resolved by removal. This PR drops the `check-colors` script, the `tsx` devDependency, and the whole `scripts/` directory, so nothing in the package relies on an undeclared binary any more.

The general point was right, and the same class of problem bit this branch elsewhere: `packages/api-client` tests import `h3` without declaring it and relied on `shamefullyHoist: true` to supply it. After merging `main`, the dependency graph shifted and the hoisted `h3` flipped from 1.x to 2.x, which removed the `App` type export and broke the Test job. Fixed in b7b446bf by declaring `h3` in that package.

turbo.json
mkucmus: Both points are handled, and the mechanism you proposed is now what the repo does.

The `.nuxt` output override is gone from `turbo.json`, so no `nuxt prepare` artifact is cached any more. Instead, #2629 on `main` added a dedicated `nuxt-prepare` task with `cache: false`, and `build` and `generate` now list it in `dependsOn`. That means `.nuxt` is regenerated on each machine rather than restored from a remote cache, which is exactly the predictable-regeneration approach you described.

That change arrived here through the merge of `main`, so the scope concern is moot: this PR no longer touches the starter template's build caching at all.

packages/api-client/package.json
```diff
@@ -70,6 +70,7 @@
     "@typescript/native-preview": "7.0.0-dev.20260205.1",
     "@vitest/coverage-v8": "4.1.10",
     "fflate": "0.8.3",
+    "h3": "1.15.11",
```
patzick, line 73: **Nit** — good catch and the right fix (these tests do `import { createApp, … } from "h3"`, so relying on `shamefullyHoist` was always fragile), but it is invisible from the outside: the PR description and the changeset both describe only the `cms-base-layer` refactor, so a reader later will not know why a packaging PR touched `packages/api-client`. Worth one line in the description.

patzick (approved):
## 🔍 Re-review

Both findings from the previous round are resolved, verified at the head commit:

- **Major — `types/imports.d.ts`**: the shim re-exports `@shopware/composables` (package root) again, so the layer type-checks against the same `exports["."].types` → `src/index.ts` surface consumers resolve. No dist/src divergence, and no dependency on `packages/composables/dist` existing.
- **Minor — `turbo.json`**: the `.nuxt` output override is gone from the diff; `turbo.json` is no longer touched by this PR.

The packaging shape matches the sibling `@shopware/unocss-design-tokens-layer` (`main` + `exports` + no build). Re-checked the deletions still hold: no package-subpath imports of `@shopware/cms-base-layer/…` anywhere in the repo (only filesystem-relative references in `apps/docs/.vitepress/config.hub.ts` and a `CmsNoComponent.vue` hint string, both unaffected by the `exports` map), `resolveLayer()` in `nuxt.config.ts` uses filesystem paths so the `app/assets/css` and `app/providers` references still work, and everything `files` now lists exists.

One nit inline. Nothing blocking.

### ⏳ CI

`Build` is still pending; lint, Test, Check audit, CodeQL and the Vercel deployments pass. This approval is based on the diff and the repo state at the head commit — required checks still gate the merge.

packages/api-client/package.json
mkucmus: Fair point. Added a line to the description covering why this PR touches `packages/api-client`: the merge flipped the hoisted `h3` from 1.15.11 to 2.0.1-rc.26, `App` is gone in 2.x, and the tests never declared the package. No changeset, since it is test-only.

`listhen` in the same file is undeclared for the same reason. Declaring it shifts `srvx` across the nitro/nuxt graph, so I left it for its own PR.

Isengo1989:
@mkucmus this broke the current deploy pipeline from developer-portals. Since @bojanrajh is on vacation, could you take a look at how we could resolve this?

https://github.com/shopware/developer-portal/actions/runs/32147240203/job/95744102530#step:8:194

mkucmus:
> Maciek Kucmus (@mkucmus) this broke the current deploy pipeline from developer-portals. Since Bojan Rajh (@bojanrajh) is on vacation, could you take a look at how we could resolve this?
> 
> https://github.com/shopware/developer-portal/actions/runs/32147240203/job/95744102530#step:8:194

hey, thanks for reporting, I'm looking into it. the broken mechanism relies on generated ts types which have been moved to the other place 🙌🏼

PR #2629, opened 2026-08-12
ci: run nuxt prepare as its own turbo task

mkucmus:
### Description

`templates/vue-starter-template/tsconfig.json` uses the Nuxt 4 layout from #2606: `files: []` plus `references` into `./.nuxt`. Vite loads the nearest `tsconfig.json` when it transforms a `.vue` file, so anything compiling this template needs `.nuxt` first. `lumora-demo-store` extends it, so it compiles those files too.

`.nuxt` is created by `nuxt prepare`, which only lives inside the template's `build` script. A turbo cache hit does not run the script. It restores the files listed in `outputs` and replays the logs. `.nuxt` is not in `outputs`, so it is neither generated nor restored:

```
[TSCONFIG_ERROR] Failed to load tsconfig '../vue-starter-template/.nuxt/tsconfig.app.json': Tsconfig not found
```

This PR moves `nuxt prepare` into its own uncacheable turbo task, so it always runs.

### When it fails

Three things have to line up:

- The machine has no `.nuxt` on disk. It is gitignored, so every fresh checkout starts this way.
- `vue-starter-template#build` is a cache hit, so `nuxt prepare` is skipped.
- `lumora-demo-store` is a cache miss, so it still has to compile the template's `.vue` files.

The third point is why this is not a corner case. A change touching only the extended template splits the hashes:

```
lumora-demo-store            cache=MISS
vue-starter-template         cache=HIT
```

When both hit or both miss, nothing breaks. Hence the intermittency. Remote caching is on in every workflow, so the hit can come from another machine.

### Changes

- `vue-starter-template` gets a `nuxt-prepare` script.
- `turbo.json` gets a `nuxt-prepare` task with `cache: false`, so it always runs.
- `build` and `generate` depend on `nuxt-prepare` and `^nuxt-prepare`, covering both the template being built and the one being extended.
- `typecheck` depends on `^nuxt-prepare`.

No package names are hardcoded. Turbo skips the task where the script is not defined. Only `vue-starter-template` needs it today: it is the one package that is both extended by another and uses `references` into its own `.nuxt`.

### Why not cache `.nuxt`

Listing `.nuxt/**` as a build output also hides the error, and was the first attempt on the #2609 branch. It is the wrong tool:

- `.nuxt` is a `nuxt prepare` artifact, not a build output.
- 5 of its 47 files contain absolute paths (`shopware.d.ts`, `uno.config.mjs`, `app.config.mjs`, `types/layouts.d.ts`, `types/nitro-imports.d.ts`), and remote cache would carry one machine's paths onto another.

That workaround has since been reverted, so neither fix is on `main` today. Thanks @patzick for catching it in review.

### Type of change

Bug fix (non-breaking change which fixes an issue)

### ToDo's

- [x] Changeset file provided (not needed, only private packages and root config change)
- [x] Documentation added/updated (not needed)

### Additional context

Reproduced against current `main` config:

1. `turbo run build --filter=vue-starter-template` to populate the cache.
2. `rm -rf templates/vue-starter-template/.nuxt`.
3. Re-run the build: `6 cached, 6 total` in 46ms, `>>> FULL TURBO`, and `.nuxt` is not recreated.
4. `turbo run typecheck --filter=lumora-demo-store` fails with the error above.

After the fix, both paths are covered and caching is preserved (7 of 8 tasks still cached in 1.9s, only `nuxt-prepare` bypassing). A repo-wide run on the #2609 branch was clean: `pnpm build` 35/35, `pnpm typecheck` 69/69, `pnpm lint` 31/31.

Not related to #2632. That was a client bundle resolve error for `@oxc-parser/binding-wasm32-wasi`. This is a tsconfig lookup during transform.


templates/vue-starter-template/package.json
```diff
@@ -18,6 +18,7 @@
   "scripts": {
     "dev": "nuxt prepare && nuxt dev",
     "build": "nuxt prepare && nuxt build",
+    "nuxt-prepare": "nuxt prepare",
     "generate": "nuxt generate",
```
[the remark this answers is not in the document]
mkucmus: The inline nuxt prepare keeps these scripts package-manager agnostic, which matters because this template is scaffolding users copy and run outside the monorepo. nuxt prepare takes no arguments, so there is nothing to drift. The turbo task exists to cover the cache-hit path, not to replace the inline call.

turbo.json
```diff
@@ -2,13 +2,18 @@
   "$schema": "https://turborepo.org/schema.json",
   "tasks": {
     "build": {
-      "dependsOn": ["^build"],
+      "dependsOn": ["^build", "nuxt-prepare", "^nuxt-prepare"],
```
patzick: `^nuxt-prepare` is redundant here (same in `generate` line 9 and `typecheck` line 30): `^build` already resolves each dependency's `build`, which now itself `dependsOn` `nuxt-prepare`, so the dependency's prepare runs either way. In `typecheck` the local `nuxt-prepare` is likewise implied by its `build` dep.

It does buy one thing — a package that defines `nuxt-prepare` but no `build` script would otherwise be missed — so this is fine to keep if that's deliberate. Just noting it's now three places to keep in sync for one behaviour.

turbo.json
```diff
@@ -51,6 +56,7 @@
       "outputs": []
     },
     "dev": {
+      "dependsOn": ["^nuxt-prepare"],
```
patzick, line 59: Worth calling out: `dev` had no `dependsOn` before. Via `^nuxt-prepare` → `nuxt-prepare.dependsOn: ["^build"]`, starting dev for `lumora-demo-store` (or a root `turbo dev`) now builds the six workspace-linked `@shopware/*` packages first — they resolve as `link:../../packages/*` in the lockfile, so `^build` is not a no-op here.

Correct in principle (lumora's dev compiles the template's `.vue` files and needs `.nuxt`), but it changes dev startup from instant to a full package build on a fresh checkout. If that's unwanted, `nuxt-prepare` may not actually need the whole `^build` chain — `nuxt prepare` mainly needs the module packages resolvable, not every dist rebuilt.

patzick (approved):
Approving — diagnosis and fix match: `cache: false` is the only way to make `nuxt prepare` survive a `build` cache hit, and keeping `.nuxt` out of `outputs` is right given the absolute paths in it. CI is green on all required checks.

Two non-blocking notes inline (redundant `^nuxt-prepare` entries, and `dev` now pulling a full `^build`).

turbo.json
mkucmus: You are right, fixed.

I compared what runs before each task, with and without `^nuxt-prepare`. Same list both ways, so it did nothing.

The side note does not hold either. I added a test `nuxt-prepare` script to `@shopware/unocss-design-tokens-layer`, which has no `build` script. Turbo still ran it without `^nuxt-prepare`. It creates the `#build` node even when the script is missing, and follows its `dependsOn`.

So the rule is simple: `^nuxt-prepare` does nothing where `^build` already is. Only `dev` needs it, because `dev` has no `^build`.

```
build:        ["^build", "nuxt-prepare"]
generate:     ["^build", "nuxt-prepare"]
typecheck:    ["build", "^build"]
dev:          ["^nuxt-prepare"]
nuxt-prepare: ["^build"]
```

One risk to know: `nuxt-prepare` in `build.dependsOn` is now the only thing holding this up. Remove it and all three tasks break quietly.

turbo.json
mkucmus: I measured it. `dev` for lumora waits for 17 tasks: 3.3s with a warm cache (6 of 7 cached), 45s with an empty one.

The 45s is not new work. `postinstall` already runs `pnpm run build --filter='./packages/*'`, which builds the same six packages and fills the same cache. After install, `pnpm dev` gets cache hits. The real extra cost at dev time is about 3s.

On "resolvable, not rebuilt": prepare does need the dist. I moved `packages/nuxt-module/dist` away and got:

```
ERROR  The module @shopware/nuxt-module could not be loaded.
```

It points to `./dist/index.mjs`. Depending on only that one package means hardcoding its name in `turbo.json`, which this PR avoids.

And dev really does break. With the template's `.nuxt` deleted, `nuxt dev` starts fine, but the first request returns HTTP 500:

```
[TSCONFIG_ERROR] Failed to load tsconfig
'../vue-starter-template/.nuxt/tsconfig.app.json': Tsconfig not found
```

Vite only transforms files when asked, so the server looks healthy until you open a page.

PR #2641, opened 2026-08-18
feat(vue-starter-template): wishlist pagination

### LINKED ISSUE #2640: Move wishlist page to SharedPaginatedList + wishlist tiles do not link to the product
### Technical TODO

Rewrite `app/pages/wishlist.vue` to use `SharedPaginatedList` + `useOffsetPaginatedList`

and fix not clickable header wishlist icon

mdanilowicz:
closes #2640

https://github.com/user-attachments/assets/29b25065-0743-4ac9-8e80-5a5580284a0f

This pull request refactors and improves the wishlist page and related components in the Vue starter template. The main focus is on simplifying data fetching and pagination for the wishlist, enhancing the user experience with better loading states, accessibility, and internationalization, and improving the maintainability of the codebase.

**Wishlist Page Refactor and Feature Improvements:**

* The wishlist page (`wishlist.vue`) now uses a generic, composable-based paginated list (`SharedPaginatedList`) for fetching and displaying wishlist products, replacing custom fetching and pagination logic. This results in cleaner, more maintainable code and consistent pagination behavior across the app. [[1]](diffhunk://#diff-fb5e4d1140ef3a0dd2ac2148e51ef7bceaa522b1a62776c00a938c564615d8d0R2-R33) [[2]](diffhunk://#diff-fb5e4d1140ef3a0dd2ac2148e51ef7bceaa522b1a62776c00a938c564615d8d0R45-R88)
* Wishlist product tiles now support removal with immediate UI feedback and automatic refresh of the list when a product is removed. [[1]](diffhunk://#diff-0b22146ef3e79ff4341c7366efcff6a06ca77261892a6854a4839706a0ceecc2R13-R23) [[2]](diffhunk://#diff-0b22146ef3e79ff4341c7366efcff6a06ca77261892a6854a4839706a0ceecc2R39-R63)

**Pagination and Loading State Enhancements:**

* The `Pagination.vue` component has been rewritten to always show a maximum of 7 visible pages with ellipsis for large lists, improved accessibility (ARIA labels, disabled states), and a more robust structure.
* The paginated list logic (`useOffsetPaginatedList.ts` and `PaginatedList.vue`) now distinguishes between initial and subsequent loading states, disables navigation when appropriate, and improves scroll-to-target behavior. [[1]](diffhunk://#diff-991af0f58595edadf094b3d01c00de54dee374d19aa9e7f5c2aebafc2813dc2eR52) [[2]](diffhunk://#diff-991af0f58595edadf094b3d01c00de54dee374d19aa9e7f5c2aebafc2813dc2eL90-R91) [[3]](diffhunk://#diff-991af0f58595edadf094b3d01c00de54dee374d19aa9e7f5c2aebafc2813dc2eR107-R119) F19d39aeL19R19, [[4]](diffhunk://#diff-f11a27f73078088eef09ee288be6a94e87b3f4debcdef6fc60c416cd43fe50ecL83-R88) [[5]](diffhunk://#diff-f11a27f73078088eef09ee288be6a94e87b3f4debcdef6fc60c416cd43fe50ecL96-R120) [[6]](diffhunk://#diff-f11a27f73078088eef09ee288be6a94e87b3f4debcdef6fc60c416cd43fe50ecR159)

**Wishlist Navigation and Routing:**

* The wishlist icon in the header now links to the wishlist page, improving navigation.
* A bug fix ensures that wishlist data is only fetched when not already on a wishlist-related route.

**Internationalization:**

* Added and improved translations for pagination controls and empty wishlist states in English, German, and Polish, providing a better experience for international users. [[1]](diffhunk://#diff-a9f45f62d49397b02c597421c1e9085e50792509b4c9cecc44d6e727fba2bb06R32-R33) [[2]](diffhunk://#diff-b967764f7a8a7e1a3b73284fdca6e2cd8ac9a9f60679e18465aa8235ff2b19a3R32-R33) [[3]](diffhunk://#diff-6e1ae4030c33a0b759fea04a6d8c1c2d755d31eaba4bee9acb3dee3da3fee82dR32-R33) [[4]](diffhunk://#diff-d129551742269b77ee3b9cdb7a5b05bbad92b9f183bba5cf57a958f85339ecaaR5-R9) [[5]](diffhunk://#diff-be2b8786011e66ff57f04ec9c3e94d105ef744a5773c02a222884d9b33edb012R5-R9) [[6]](diffhunk://#diff-dfbef6fa691a34e2c3d9ba117a2315e1a10140539bc6fde6040945edec283f92R5-R9)

---

**Most important changes:**

**Wishlist Page Refactor:**
- Refactored `wishlist.vue` to use a generic paginated list component, simplifying data fetching and pagination logic for the wishlist. Also added empty state handling and loading skeletons. [[1]](diffhunk://#diff-fb5e4d1140ef3a0dd2ac2148e51ef7bceaa522b1a62776c00a938c564615d8d0R2-R33) [[2]](diffhunk://#diff-fb5e4d1140ef3a0dd2ac2148e51ef7bceaa522b1a62776c00a938c564615d8d0R45-R88)
- Wishlist product tile now emits a `removed` event and triggers a list refresh when an item is removed from the wishlist, improving UX. [[1]](diffhunk://#diff-0b22146ef3e79ff4341c7366efcff6a06ca77261892a6854a4839706a0ceecc2R13-R23) [[2]](diffhunk://#diff-0b22146ef3e79ff4341c7366efcff6a06ca77261892a6854a4839706a0ceecc2R39-R63)

**Pagination Improvements:**
- Rewrote `Pagination.vue` to use a more robust, accessible, and user-friendly pagination system with ellipsis and ARIA labels.
- Enhanced paginated list loading states and navigation, distinguishing initial loading, disabling navigation during loading, and improving scroll-to-target logic. [[1]](diffhunk://#diff-991af0f58595edadf094b3d01c00de54dee374d19aa9e7f5c2aebafc2813dc2eR52) [[2]](diffhunk://#diff-991af0f58595edadf094b3d01c00de54dee374d19aa9e7f5c2aebafc2813dc2eL90-R91) [[3]](diffhunk://#diff-991af0f58595edadf094b3d01c00de54dee374d19aa9e7f5c2aebafc2813dc2eR107-R119) F19d39aeL19R19, [[4]](diffhunk://#diff-f11a27f73078088eef09ee288be6a94e87b3f4debcdef6fc60c416cd43fe50ecL83-R88) [[5]](diffhunk://#diff-f11a27f73078088eef09ee288be6a94e87b3f4debcdef6fc60c416cd43fe50ecL96-R120) [[6]](diffhunk://#diff-f11a27f73078088eef09ee288be6a94e87b3f4debcdef6fc60c416cd43fe50ecR159)

**Navigation and Routing:**
- Fixed logic to only fetch wishlist data when not already on a wishlist route and added a wishlist link to the header. [[1]](diffhunk://#diff-d8cde668b96335b72d17c432830209ffe782ffb41fde77482eb038aee09e11e0L132-R132) [[2]](diffhunk://#diff-e4491e574ec21ba09288f04474133d1fd6a249fc9899756f8187d69caa5408d0R98-R100)

**Internationalization:**
- Added translations for pagination controls and empty wishlist states in English, German, and Polish. [[1]](diffhunk://#diff-a9f45f62d49397b02c597421c1e9085e50792509b4c9cecc44d6e727fba2bb06R32-R33) [[2]](diffhunk://#diff-b967764f7a8a7e1a3b73284fdca6e2cd8ac9a9f60679e18465aa8235ff2b19a3R32-R33) [[3]](diffhunk://#diff-6e1ae4030c33a0b759fea04a6d8c1c2d755d31eaba4bee9acb3dee3da3fee82dR32-R33) [[4]](diffhunk://#diff-d129551742269b77ee3b9cdb7a5b05bbad92b9f183bba5cf57a958f85339ecaaR5-R9) [[5]](diffhunk://#diff-be2b8786011e66ff57f04ec9c3e94d105ef744a5773c02a222884d9b33edb012R5-R9) [[6]](diffhunk://#diff-dfbef6fa691a34e2c3d9ba117a2315e1a10140539bc6fde6040945edec283f92R5-R9)


templates/vue-starter-template/app/app.vue
```diff
@@ -129,7 +129,7 @@ if (languages && router.currentRoute.value.name) {
 
 onMounted(() => {
   refreshCart();
-  const isWishlistPage = route.name?.toString().endsWith("wishlist") ?? false;
+  const isWishlistPage = route.name?.toString().startsWith("wishlist") ?? false;
```
patzick: **Blocker** — this now actually skips the fetch, and nothing else fills the store.

With `strategy: "prefix_except_default"` route names are `wishlist___en-GB` / `wishlist___de-DE`, so `endsWith("wishlist")` was always `false` and `getWishlistProducts()` *did* run on the wishlist page. `startsWith` fixes the predicate — but the page no longer calls `useWishlist().getWishlistProducts()` (it hits `readCustomerWishlist` directly), so `useSyncWishlist`'s `totalWishlistItemsCount` stays `0`.

`useWishlist().count` → `countSync` → `totalWishlistItemsCount`, and `LayoutHeaderWishlistIcon` only renders the badge when `counter > 0`. Result: load `/wishlist` directly (or hard-refresh) as a logged-in user with items → **header badge disappears**. It only reappears after a removal, because `useProductWishlist.removeFromWishlist()` internally calls `getWishlistProducts()`.

`itemsSync` is empty for the same reason, so `isInWishlist` is wrong on that page too — masked here only because the tile hardcodes `:isSelected="true"`.

Two ways out: drop the special case entirely (the duplicate request is cheaper than a broken counter), or have the page fetcher go through `useWishlist` so one fetch serves both.

templates/vue-starter-template/app/composables/useOffsetPaginatedList.ts
```diff
@@ -93,12 +98,26 @@ export function useOffsetPaginatedList<T>(
...
 
   function scrollToTarget() {
     if (!scrollTarget) return;
-    toValue(scrollTarget)?.scrollIntoView({ behavior: "smooth" });
+
+    const target = toValue(scrollTarget);
+    if (!target) return;
+
+    const top = target.getBoundingClientRect?.().top;
+    if (top !== undefined && top >= 0) return;
```
patzick: **Minor** — `top >= 0` also matches "list starts below the fold".

On a page where the list top sits at e.g. `y = 900`, changing page returns early and never scrolls, so the user stays put while the content they asked for swaps out off-screen. The intent looks like "don't scroll if already visible" — that needs an upper bound too:

```ts
if (top !== undefined && top >= 0 && top <= window.innerHeight) return;
```

templates/vue-starter-template/app/components/wishlist/ProductTile.vue
```diff
@@ -10,12 +10,20 @@ const { product } = defineProps<{
...
+const localePath = useLocalePath();
+const { formatLink } = useInternationalization(localePath);
+
 const { addProduct } = useCart();
 const { removeFromWishlist } = useProductWishlist(product.id);
 const { pushError, pushSuccess } = useNotifications();
 const { t } = useI18n();
 
 const addingProducts = ref(false);
+const removing = ref(false);
```
patzick, line 30: **Minor** — `removing` guards re-entry but is never bound in the template, so there is no disabled state or spinner on the heart while the request is in flight. The PR description promises "immediate UI feedback"; right now the tile looks inert for the whole round-trip. Consider `:disabled="removing"` / an opacity class on `ProductWishlistIcon` (compare `addingProducts`, which is surfaced on the add-to-cart button).

templates/vue-starter-template/app/components/wishlist/ProductTile.vue
```diff
@@ -31,31 +39,51 @@ async function handleAddToCart() {
...
+
+async function handleRemoveFromWishlist() {
+  if (removing.value) return;
+
+  try {
+    removing.value = true;
+
+    await removeFromWishlist();
+
+    emit("removed");
```
patzick, line 55: **Minor** — three requests per removal. `removeFromWishlist()` already does `DELETE` + a full `getWishlistProducts()` (that inner call is what keeps the header count in sync, so it can't just be dropped), and then `removed` triggers `list.refresh()` for a third round-trip. If the page fetcher read from the same store, the refresh would be free.

templates/vue-starter-template/app/components/wishlist/ProductTile.vue
```diff
@@ -31,31 +39,51 @@ async function handleAddToCart() {
...
+    </NuxtLink>
 
-    <div
+    <NuxtLink
+      :to="formatLink(getProductRoute(product))"
       class="text-surface-on-surface text-2xl font-normal font-['Noto_Serif'] leading-9"
     >
-      {{ product.name }}
-    </div>
+      {{ product.translated.name }}
```
patzick: **Minor** — inconsistent with the `alt` a few lines up, which still uses `product.name`. Pick one (`product.translated?.name ?? product.name` is the safer read) and use it in both places.

templates/vue-starter-template/app/components/shared/Pagination.vue
```diff
@@ -1,92 +1,92 @@
...
+        v-else
+        type="button"
+        class="relative inline-flex items-center justify-center min-w-[3rem] px-4 py-2 border text-sm"
+        :class="
+          cell === current
+            ? 'bg-brand-primary text-brand-on-primary'
+            : 'bg-white border-outline-outline-variant outline outline-1 outline-offset-[-1px] outline-outline-outline-variant'
+        "
+        :aria-current="cell === current ? 'page' : undefined"
+        :disabled="cell === current"
```
patzick: **Minor (a11y)** — `disabled` on the current-page button removes it from the tab order, so keyboard users tabbing through pagination lose their position marker. `aria-current="page"` already conveys the state; the standard pattern is to keep the button focusable and just not act on the click. (The prev/next `disabled` are correct — those are genuinely unavailable.)

templates/vue-starter-template/app/pages/wishlist.vue
```diff
@@ -1,78 +1,36 @@
...
-  await router.push({
-    query: {
-      ...route.query,
-      p: page,
-      limit: limit.value,
-    },
-  });
-  getWishlistProductsProxy({ page });
-}
+const list = ref<{ refresh: () => unknown } | null>(null);
```
patzick: **Nit** — hand-rolled structural type for the template ref. `InstanceType<typeof SharedPaginatedList>` (or `ComponentExposed<...>` from `vue-component-type-helpers`) keeps it in sync with `defineExpose` if the exposed surface ever grows.

templates/vue-starter-template/app/components/layout/Header.vue
```diff
@@ -30,7 +30,9 @@ function toggleMobileSearch() {
 function handleMyAccountClick() {
   if (!isLoggedIn.value) {
     push(
-      `${formatLink("/account/login")}?redirect=${encodeURIComponent(route.fullPath)}`,
+      `${formatLink("/account/login")}?redirect=${encodeURIComponent(
```
patzick: **Nit** — formatting-only churn on `handleMyAccountClick`, unrelated to wishlist pagination. Looks like a local formatter disagreeing with the repo config; worth reverting to keep the diff focused.

patzick (changes requested):
Nice cleanup — the page is much smaller and pagination behaviour is now consistent with the rest of the app. CI is green.

One blocking regression: the `endsWith` → `startsWith` change in `app.vue` finally makes the "skip wishlist fetch" branch fire (route names are `wishlist___en-GB` under `prefix_except_default`, so `endsWith("wishlist")` was always `false`). Since the page now fetches through `apiClient.invoke` instead of `useWishlist`, nothing populates `useSyncWishlist`'s store any more — so the header wishlist badge reads 0 on a direct load of `/wishlist`. Details inline.

Rest is minors/nits.

templates/vue-starter-template/app/components/layout/Header.vue
mdanilowicz: done

templates/vue-starter-template/app/pages/wishlist.vue
mdanilowicz: SharedPaginatedList is a generic component, so we cannot do it as you suggest. I did it a bit differently

templates/vue-starter-template/app/components/shared/Pagination.vue
mdanilowicz: true, done

templates/vue-starter-template/app/components/wishlist/ProductTile.vue
mdanilowicz: done

templates/vue-starter-template/app/components/wishlist/ProductTile.vue
mdanilowicz: An optimistic UI pattern was used for this interaction. But I agree that we should not mix patterns. We have loaders (`addingProducts`), so we should stay with them - changed

templates/vue-starter-template/app/composables/useOffsetPaginatedList.ts
mdanilowicz: This scenario can't happen now 

`SharedPaginatedList` is the only consumer of `useOffsetPaginatedList` and it
hardcodes `scrollTarget: listEl`. `SharedElementsNavigation` — the only source of
`changePage` and of the `limit` setter, i.e. the only two paths that reach
`scrollToTarget` — renders *inside* `listEl`, below the item grid, with no sticky or
fixed positioning. So `listTop < paginationTop` always holds, and clicking the control
requires `paginationTop` to be on screen, which forces `listTop < viewportHeight`. (The
third caller of `navigate`, the out-of-range watcher, passes `replace: true` and never
scrolls.) `top >= 0 && top > innerHeight` is unreachable at the moment the function
runs.

Your own example shows it: at `top = 900` the proposed condition also returns early on
any viewport taller than 900px. For the two branches to diverge you'd need
`innerHeight < 900 < top` — precisely the case where the pagination control you just
clicked is off-screen.

This would only be possible with a different layout structure. So I agree that it might be worth having in that case, but for our current layout, this would be dead code.

templates/vue-starter-template/app/app.vue
mdanilowicz: Right, I removed the check for whether it’s a wishlist page and also trimmed down the response.

templates/vue-starter-template/app/components/shared/Pagination.vue
```diff
@@ -1,92 +1,98 @@
...
   <nav
     class="relative z-0 inline-flex rounded-md shadow-sm space-x-px"
     :aria-label="$t('layout.ariaLabels.pagination')"
   >
     <button
-      v-if="current - 1 >= 2"
-      class="relative inline-flex items-center px-2 py-2 rounded-l-md border border-outline-outline-variant bg-white text-sm outline outline-1 outline-offset-[-1px] outline-outline-outline-variant"
+      type="button"
+      class="relative inline-flex items-center px-2 py-2 rounded-l-md border border-outline-outline-variant bg-white text-sm outline outline-1 outline-offset-[-1px] outline-outline-outline-variant disabled:opacity-40 disabled:cursor-not-allowed"
+      :disabled="current <= 1"
```
patzick, line 59: **Minor** — the arrows are now unconditional, so a single-page list renders two permanently-disabled buttons.

Old markup gated every arrow (`v-if="current - 1 >= 2"` / `v-if="total > current + 1"`), so at `total === 1` it emitted just the `1` cell. Now `[<] [1] [>]` renders with both arrows dead. That's reachable on the wishlist itself: `PaginatedList` keeps `SharedElementsNavigation` mounted whenever `showPageSizeSelector` is true (the default), so any wishlist with ≤ 15 items shows the control. Same for `account/order`.

Cheapest fix is an early return in the component: `v-if="total > 1"` on the `<nav>`, leaving the size selector alone.

templates/vue-starter-template/app/pages/wishlist.vue
```diff
@@ -1,78 +1,36 @@
...
+const fetchWishlistProducts: OffsetPaginationFetcher<
+  Schemas["Product"]
+> = async ({ page, limit }) => {
+  const { data } = await apiClient.invoke(
+    "readCustomerWishlist post /customer/wishlist",
+    {
+      body: {
+        page,
+        limit,
+        "total-count-mode": "exact",
```
patzick, line 24: **Minor** — no `associations: { seoUrls: {} }`, so the new product links likely fall back to technical URLs.

`getProductRoute` → `getProductUrl` reads `product.seoUrls?.[0]?.seoPathInfo` and returns `/detail/${id}` when it's absent. `searchSuggest` (the tile you modelled this on) gets `seoUrls` from the listing route's criteria; `readCustomerWishlist` builds its criteria from the request body, so unless the platform adds the association for you, every tile links to `/detail/<uuid>`. That still resolves through `[...all].vue`, but it bypasses the SEO URL.

Worth confirming against a response from the preview — if `seoUrls` is missing, add the association here.

templates/vue-starter-template/app/app.vue
```diff
@@ -129,10 +128,12 @@ if (languages && router.currentRoute.value.name) {
 
 onMounted(() => {
   refreshCart();
-  const isWishlistPage = route.name?.toString().endsWith("wishlist") ?? false;
-  if (!isWishlistPage) {
-    getWishlistProducts();
-  }
+  // hydrate the wishlist store with the product ids only
+  getWishlistProducts({
```
patzick, line 136: **Minor** — id-only hydration is still capped at the default `limit: 15`.

`useWishlist.getWishlistProducts` defu's your criteria over `{ limit: limit.value }` (15), so `_wishlistItems` only ever holds the first page of ids. `useProductWishlist.isInWishlist` reads exactly that array, so on a wishlist of 20+ items the heart on the 16th product's detail page renders unfilled even though it *is* wishlisted.

Pre-existing (the old `endsWith` predicate never fired, so this path already ran with limit 15) — but this line now owns the intent, and the comment above says "the product ids only". Since the payload is one field per row, passing a generous `limit` here is nearly free and makes the comment true.

templates/vue-starter-template/app/components/wishlist/ProductTile.vue
```diff
@@ -34,31 +43,53 @@ async function handleAddToCart() {
...
+
+async function handleRemoveFromWishlist() {
+  if (removing.value) return;
+
+  try {
+    removing.value = true;
+
+    await removeFromWishlist();
+
+    emit("removed");
```
patzick, line 55: **Minor** (carried over, no reply last round) — three round-trips per removal.

`removeFromWishlist()` does `DELETE` + a full `getWishlistProducts()` — and that inner call is what keeps the header count in sync, so it can't just be dropped — then `removed` fires `list.refresh()` for a third. Note the inner call passes no criteria, so it also re-fetches 15 *full* products and resets `_limit`, undoing the id-only hydration you just set up in `app.vue`.

Not blocking, and hard to fix without touching the composable package. Flagging so it's a conscious trade-off rather than an oversight.

templates/vue-starter-template/app/components/wishlist/ProductTile.vue
```diff
@@ -34,31 +43,53 @@ async function handleAddToCart() {
...
-      :src="getSmallestThumbnailUrl(product.cover?.media)"
-      :alt="`${getTranslatedProperty(product, 'name')} item`"
-      fit="inside"
-      class="object-cover"
-      :style="`height: ${ELEMENT_HEIGHT}px; width: ${ELEMENT_WIDTH}px;`"
-    />
+    <NuxtLink :to="formatLink(getProductRoute(product))">
+      <NuxtImg
+        :src="getSmallestThumbnailUrl(product.cover?.media)"
+        :alt="`${getTranslatedProperty(product, 'name')} item`"
```
patzick, line 80: **Minor (a11y)** — image and name are two sibling links to the same target, so the tile announces two links with near-identical names (`"<name> item"`, then `"<name>"`).

Either wrap image + name in one `NuxtLink`, or keep both and give the image `alt=""` so it's decorative and only the text link carries the accessible name.

templates/vue-starter-template/app/composables/useOffsetPaginatedList.ts
```diff
@@ -80,7 +92,7 @@ export function useOffsetPaginatedList<T>(
   );
 
   const { data, status, error, refresh } = useAsyncData(
-    () => stateKey.value,
+    key,
```
patzick, line 96: **Nit** — switching to the static `key` is the right call (it's what keeps `data` populated across page changes so `isInitialLoading` can distinguish the two), but `stateKey` is now purely a watch trigger while still prefixing `key` to itself. Dropping the prefix would make the intent obvious.

Separately, it stringifies the whole `route.query`, so any unrelated param change refetches the list — e.g. the `?redirect=` the login link appends. Narrowing it to `pageQueryKey`/`limitQueryKey` + `watchSources` would be tighter.

templates/vue-starter-template/app/components/shared/Pagination.vue
```diff
@@ -1,92 +1,98 @@
...
-      @click="$emit('changePage', total)"
-    >
-      {{ total }}
-    </button>
-    <button
-      v-if="total > current + 1"
-      class="relative inline-flex items-center px-2 py-2 rounded-r-md border border-outline-outline-variant bg-white text-sm"
+      type="button"
+      class="relative inline-flex items-center px-2 py-2 rounded-r-md border border-outline-outline-variant bg-white text-sm outline outline-1 outline-offset-[-1px] outline-outline-outline-variant disabled:opacity-40 disabled:cursor-not-allowed"
+      :disabled="current >= total"
```
patzick, line 94: **Nit** — prev/next still `$emit` directly while the numbered cells go through `handleChangePage`. Harmless today because `:disabled` covers the bounds, but routing all three through the guard keeps the no-op check in one place.

patzick (approved):
Re-review of `21fa4e0a`. The blocker is resolved: `app.vue` now hydrates unconditionally and `count` comes from `products.total`, so the header badge survives a direct load of `/wishlist`. The id-only `includes` keeps that fetch cheap. Pagination a11y, the `removing` disabled state, the `translated.name` mismatch and the formatting churn are all addressed, and CI is green.

Approving. Five minors/nits inline — none blocking, pick up what you agree with.

On the `scrollToTarget` upper bound: your analysis is right that it's unreachable with today's layout, and the comment you left documents exactly why it should stay. Good outcome.

Still open from the last round: the three-requests-per-removal point (inline below) — no reply on that one.

PR #2642, opened 2026-08-18
feat: update schema and types to 6.7.12

### LINKED ISSUE #2674: Update schema to 6.7.12

mdanilowicz:
closes #2674

Update schema and types to 6.7.12


.changeset/olive-quotes-decline.md
```diff
@@ -0,0 +1,21 @@
+---
+"@shopware/composables": minor
```
mkucmus, line 2: **Changeset should be `major`.**

The bump is `minor` but the body is marked `[BREAKING]`. On a `minor`, consumers with `^1.12.0` pick this up on a routine update and break.

`.changeset/tidy-carrots-shake.md` in the same release uses `major` for its breaking change.

packages/composables/src/useB2bQuoteManagement/useB2bQuoteManagement.ts
```diff
@@ -88,16 +97,18 @@ export function useB2bQuoteManagement(): UseB2bQuoteManagement {
    * Decline quote
    *
    * @param {string} quoteId
-   * @param {string} comment
+   * @param {DeclineQuoteParams} params
    * @returns {Promise<void>}
    */
-  async function declineQuote(quoteId: string, comment: string) {
+  async function declineQuote(quoteId: string, params: DeclineQuoteParams) {
```
mkucmus: **Docs still show the old signature.**

`apps/docs/src/guides/b2b/quote-management.md:114` and `apps/docs/src/integrations/commercial/b2b-quote-management.md:114` both still have:

```ts
declineQuote(quote.value.id, declineComment.value);
```

A string has none of the new fields, so the body serializes to `{}` and the request 400s. Could these be updated alongside?

packages/api-client/api-types/storeApiTypes.d.ts
```diff
@@ -14676,141 +16696,591 @@ export type operations = {
...
+    body: {
+      /** Id of the payment method */
+      paymentMethodId?: string;
+      /** Id of the shipping method */
+      shippingMethodId?: string;
+    };
+    response: never;
+    responseCode: 204;
+  };
+  "declineQuote post /quote/{id}/decline": {
```
mkucmus, line 16866: **`@shopware/api-client` needs a changeset too.**

Its `./store-api-types` export is public API, and this PR changes it: the three quote bodies become required, `Product.type` drops `"grouped_bundle"`, and `SaasStorefrontDemoToken` is removed.

Without a changeset the package never bumps, so consumers do not get the 6.7.12 types and these changes ship unannounced on the next release.

templates/vue-starter-template-extended/api-types/storeApiTypes.d.ts
```diff
@@ -14,6 +14,19 @@ type GenericRecord =
   | {
       [key: string]: GenericRecord;
     };
+type CustomFields = {
```
mkucmus, line 17: **These types look un-regenerated.**

The header at line 5 still reports `Shopware API version: unknown`, and `declineQuote` here keeps the old `body?: { comment?: string }`.

This template's `api-gen.config.json` matches the sibling template that regenerated fine, so it may just have been missed.

mkucmus (commented):
A few blocking points before merge, mostly release/metadata rather than the composable change itself. Details inline.

.changeset/olive-quotes-decline.md
patzick: Can we make it backward compatible?
Either:
- accept a string OR object
- add new method for declining quota and deprecate current one

we definetly should not have BC in that bump

patzick: i see these params are additive in API, so i'd say another method and deprecate old one, less confusion

patzick, line 2: 🚫 **Blocker — this bump should not break `declineQuote`.**

Per @patzick above: the three body fields are additive on the API side, so there is no reason for the composable to break. The `minor` bump can stay — the signature change is what has to go.

Preferred shape (matches the "less confusion" argument over a `string | object` union):

```ts
type UseB2bQuoteManagement = {
  /** @deprecated use `declineQuoteWithComment` instead — it sends the `lineItemId`
   *  and `versionId` that `POST /quote/{id}/decline` expects. Removed in v2. */
  declineQuote: (quoteId: string, comment: string) => Promise<void>;
  declineQuoteWithComment: (quoteId: string, params: DeclineQuoteParams) => Promise<void>;
};
```

The old method keeps working (`comment` only), the new one carries the full body, and nobody breaks on a routine `^1.12.0` update.

One thing to confirm before landing this: the generated types now say all three fields are **required** (`body: { comment: string; lineItemId: string; versionId: string }`, was `body?: { comment?: string }`). If the API really does still accept a `comment`-only body, the deprecated path is safe as-is — otherwise the deprecation notice should say the old method 400s against 6.7.12 and callers must migrate, rather than implying it still works.

packages/composables/src/useB2bQuoteManagement/useB2bQuoteManagement.ts
```diff
@@ -24,6 +24,15 @@ type UseB2bQuoteManagement = {
   ) => Promise<void>;
 };
 
+export type DeclineQuoteParams = {
+  /** Message content */
+  comment: string;
+  /** Identifier of the quote line item that owns the new comment */
+  lineItemId: string;
+  /** Draft version identifier of the quote currently shown in storefront */
+  versionId: string;
```
patzick: ⚠️ **Major — `versionId` is now required, but the composable offers no way to obtain it.**

The schema documents this as the *draft* version identifier, produced by `createDraftQuoteVersion post /quote/{id}/draft-version` (which returns `{ versionId }`). `useB2bQuoteManagement` doesn't wrap that operation, so consumers are forced to drop to `apiClient.invoke` directly to satisfy a required parameter of a composable method — and, as `Quote.vue` shows, the tempting shortcut (`quote.versionId`) is the entity's own version id, not a draft version.

Please add a `createDraftQuoteVersion` wrapper to this composable alongside the signature change, so the new required field has a supported source.

examples/b2b-quote-management/src/components/Quote.vue
```diff
@@ -25,7 +25,11 @@ const handleChangeRequest = async () => {
 };
 
 const handleDecline = async () => {
-  declineQuote(quote.value.id, declineComment.value);
+  declineQuote(quote.value.id, {
+    comment: declineComment.value,
+    lineItemId: quote.value.lineItems?.[0]?.id,
+    versionId: quote.value.versionId,
```
patzick: ⚠️ **Major — both values can be `undefined` at runtime, and neither is the right source.**

In the `Quote` schema `lineItems` and `versionId` are both optional, so `lineItems?.[0]?.id` and `quote.value.versionId` are `string | undefined` — sent as `undefined` for fields the API now requires, giving a 400. This typechecks only because `quote` is declared `const quote = ref()` (i.e. `any`) on line 10; typing it as `Ref<Schemas["Quote"] | undefined>` would surface the error immediately.

Beyond nullability, the sources look wrong:
- `lineItemId` is documented as *"the quote line item that owns the new comment"* — arbitrarily taking `lineItems[0]` is not that.
- `versionId` is documented as the *draft* version id from `createDraftQuoteVersion`, not the quote entity's `versionId`.

This example is the reference devs copy, so it should show the intended flow: create/track the draft version, pick the line item deliberately, and guard before calling.

packages/composables/src/useB2bQuoteManagement/useB2bQuoteManagement.ts
```diff
@@ -88,16 +97,18 @@ export function useB2bQuoteManagement(): UseB2bQuoteManagement {
...
+  async function declineQuote(quoteId: string, params: DeclineQuoteParams) {
     await apiClient.invoke("declineQuote post /quote/{id}/decline", {
       pathParams: {
         id: quoteId,
       },
       body: {
-        comment,
+        comment: params.comment,
+        lineItemId: params.lineItemId,
+        versionId: params.versionId,
```
patzick: 💡 Nit — `DeclineQuoteParams` is structurally identical to the request body, so `body: params` says the same thing and won't silently drop a field if the schema grows one.

patzick (changes requested):
## 🔍 Review

Generated 6.7.12 types look consistent (all files bumped to `6.7.12.1`), and the `declineQuote` signature change is correctly driven by the schema: `POST /quote/{id}/decline` went from `body?: { comment?: string }` to a **required** body with `comment` + `lineItemId` + `versionId`.

Requesting changes on 4 items — 2 are in inline comments, 2 have no diff line to attach to:

**🚫 Blocker — docs still show the removed signature.** Not in the diff, so please update:
- `apps/docs/src/guides/b2b/quote-management.md:114`
- `apps/docs/src/integrations/commercial/b2b-quote-management.md:114`

Both still say `declineQuote(quote.value.id, declineComment.value)`, which no longer compiles after this PR.

**⚠️ Major — `templates/vue-starter-template-extended/api-types/storeApiTypes.d.ts` was not regenerated against 6.7.12.** It changed by 158 lines while every other generated file changed by ~3400, its header still reads `Shopware API version: unknown`, and none of the 6.7.12 additions are present (`EmployeeContext`, `B2bEmployeeAccount`, `AdvancedSearchCompoundDictionary`, `descriptionTeaser`, the `swLanguageId` parameter — 0 occurrences vs. 1–241 in the other files). Its header descriptions also differ (`"Language ID for localized content"` vs. `"Instructs Shopware to return the response in the given language"`), so it appears to come from a different instance. Please regenerate it, or state in the PR why this template is intentionally pinned to an older schema — the title claims 6.7.12 for everything.

**CI:** all 10 required checks pass; head is `MERGEABLE`.

packages/api-client/api-types/storeApiTypes.d.ts
mdanilowicz: adeed

examples/b2b-quote-management/src/components/Quote.vue
mdanilowicz: done

packages/composables/src/useB2bQuoteManagement/useB2bQuoteManagement.ts
mdanilowicz: done

templates/vue-starter-template-extended/api-types/storeApiTypes.d.ts
mdanilowicz: fixed

.changeset/olive-quotes-decline.md
mdanilowicz: done

packages/composables/src/useB2bQuoteManagement/useB2bQuoteManagement.ts
mdanilowicz: done

.changeset/olive-quotes-decline.md
mkucmus: Here are the operations in packages/api-client/api-types/storeApiTypes.d.ts where body goes from optional to required (i.e., a param that used to be omittable is now mandatory for apiClient.invoke(...) callers):

| Operation | Path | Before | After |
|---|---|---|---|
| `declineQuote` | `POST /quote/{id}/decline` | `body?: { comment?: string }` | `body: { comment: string; lineItemId: string; versionId: string }` |
| `markMessagesAsReadInQuote` | `POST /quote/{id}/read-message` | `body?: GenericRecord` | `body: { versionId: string }` |
| `requestQuote` | `POST /quote/request` | `body?: { comment?: string }` | `body: { comment?: string }` |

however It would be cleaner to have b2b or other schema that comes from an extension - isloated from core-ish ones.

.changeset/heavy-schemas-update.md
```diff
@@ -0,0 +1,23 @@
+---
+"@shopware/api-client": minor
```
patzick, line 2: schema update should not cause major bump

the whole changeset should just be one-liner about version bump

.changeset/olive-quotes-decline.md
patzick: no more breaks here with additive api

patzick (changes requested):

.changeset/heavy-schemas-update.md
mdanilowicz: done

packages/composables/src/useB2bQuoteManagement/useB2bQuoteManagement.ts
```diff
@@ -7,7 +7,25 @@ import type { Schemas } from "#shopware";
...
+  /**
+   * @deprecated use `declineQuoteWithComment` instead — it sends the
+   * `lineItemId` and `versionId` that `POST /quote/{id}/decline` expects.
+   * Removed in v2.
+   *
+   * @todo Confirm against a 6.7.12 instance whether a `comment`-only body is
+   * still accepted, then state it here. If it is rejected, this note has to say
+   * that the method 400s on 6.7.12+ and callers must migrate.
+   */
   declineQuote: (quoteId: string, comment: string) => Promise<void>;
```
patzick: 🚫 **Blocker — this `@todo` cannot ship.**

This is my review question from the last round pasted verbatim into the public type declaration (and again at line 174 on the implementation). It will render in every consumer's IDE tooltip and in the generated API docs — "confirm against a 6.7.12 instance" is a note to the author, not documentation for users.

The question itself still needs answering, because it decides what the deprecation notice says:
- **If a `comment`-only body is still accepted:** drop both `@todo` blocks, the `@deprecated` text as written is correct.
- **If 6.7.12 rejects it:** the notice must say the method returns 400 on 6.7.12+ and callers *must* migrate — right now it implies the old method still works, which would be the opposite of the truth.

Could you check against a real 6.7.12 instance and fold the answer into the `@deprecated` text, removing both `@todo` blocks?

examples/b2b-quote-management/src/components/Quote.vue
```diff
@@ -1,43 +1,108 @@
...
+      lineItemId,
+      versionId,
+    });
+
+    declineComment.value = "";
+    declineLineItemId.value = undefined;
+    await refreshQuote();
+  } catch (error) {
+    declineError.value =
+      error instanceof Error ? error.message : "Could not decline the quote.";
```
patzick, line 109: ⚠️ **The draft version is created but never cleaned up.**

`createDraftQuoteVersion` runs on every click of *Decline quote*. If `declineQuoteWithComment` then fails, the `catch` only renders the message — the draft version stays on the server. Retry a few times against a validation error and you have accumulated several orphaned drafts for one quote.

The composable now exposes `deleteDraftQuoteVersion`, but nothing in the repo calls it — this example is the natural place to show the full lifecycle:

```ts
const versionId = await createDraftQuoteVersion(quote.value.id);
try {
  await declineQuoteWithComment(quote.value.id, { comment: declineComment.value, lineItemId, versionId });
} catch (error) {
  await deleteDraftQuoteVersion(quote.value.id, versionId);
  throw error;
}
```

Worth confirming with the B2B team whether the backend discards the draft itself on a successful decline. If it does not, the success path needs a `saveDraftQuoteVersion` (`POST /quote/{id}/draft-version/save`) or a delete too — that operation exists in the 6.7.12 schema but has no wrapper in the composable.

packages/composables/src/useB2bQuoteManagement/useB2bQuoteManagement.ts
```diff
@@ -85,7 +117,63 @@ export function useB2bQuoteManagement(): UseB2bQuoteManagement {
...
+   * Create a temporary draft version of the quote for storefront editing.
+   *
+   * The returned identifier is the `versionId` that quote write operations
+   * expect, for example {@link declineQuoteWithComment}. It is not the
+   * `versionId` property of the quote entity.
+   *
+   * @param {string} quoteId
+   * @returns {Promise<string>} draft version identifier
+   */
+  const createDraftQuoteVersion = async (quoteId: string) => {
```
patzick: 💡 Nit — `const createDraftQuoteVersion = async () => {}` is the only arrow function among the methods here; `deleteDraftQuoteVersion`, `declineQuote`, `declineQuoteWithComment` and the rest are all `async function`. Worth matching for consistency.

patzick (changes requested):
## 🔄 Re-review

Everything from the previous round is addressed — thanks:

- ✅ No BC break: `declineQuote(quoteId, comment)` kept and deprecated, `declineQuoteWithComment` added, changeset stays `minor`
- ✅ `createDraftQuoteVersion` / `deleteDraftQuoteVersion` wrappers added, so the required `versionId` now has a supported source
- ✅ `Quote.vue` typed as `Ref<Schemas["Quote"] | undefined>`, line item picked explicitly via `Select`, guards + error surface added, missing `Login` import fixed
- ✅ Both docs pages rewritten to the new flow
- ✅ All 13 generated type files now report `6.7.12.1`, incl. `vue-starter-template-extended`
- ✅ `@shopware/api-client` changeset added
- ✅ Test coverage for all four methods incl. the no-`versionId` throw

CI: all 11 checks pass, `MERGEABLE`.

**3 items left** (2 inline + 1 nit) — the first one is the only real blocker: my own review question got copied into the shipped JSDoc as a `@todo` and still isn't answered.

.changeset/lucky-quotes-draft.md
```diff
@@ -0,0 +1,19 @@
...
+const { createDraftQuoteVersion, declineQuote } = useB2bQuoteManagement();
+
+const versionId = await createDraftQuoteVersion(quoteId);
+
+await declineQuote(quoteId, {
+  comment: "Too expensive",
+  lineItemId,
+  versionId,
+});
+```
```
mkucmus: **Stale changeset - this sample does not compile against the shipped signature.**

`declineQuote` kept its `(quoteId, comment: string)` signature after the last round; the object form is `declineQuoteWithComment`. Line 7 and the code sample both still describe the old shape, and this text goes straight into the release notes.

````suggestion
Quote write operations such as `declineQuoteWithComment` expect the identifier of a temporary storefront draft version, not the `versionId` property of the quote entity. `createDraftQuoteVersion` wraps `POST /quote/{id}/draft-version` and returns that identifier, throwing when the API responds without one; `deleteDraftQuoteVersion` discards the draft again.

```ts
const { createDraftQuoteVersion, declineQuoteWithComment } =
  useB2bQuoteManagement();

const versionId = await createDraftQuoteVersion(quoteId);

await declineQuoteWithComment(quoteId, {
  comment: "Too expensive",
  lineItemId,
  versionId,
});
```
````

mkucmus (approved):

PR #2645, opened 2026-08-19
fix(docs): resolve cms-base-layer types for reference page

mkucmus:
## Problem

#2609 made `cms-base-layer` a plain Nuxt layer and deleted `packages/cms-base-layer/src/index.ts`.

The docs pipeline did not expect that. `ReadmeBasedReference` runs `ts-dox` on `<package>/src/index.ts` for every page under `packages/`, with no check that the file exists. For `cms-base-layer` it threw `ENOENT` and took down the whole Developer Portal build.

## Fix

- `plugin.ts` skips API extraction when a package has no TypeScript entry point. Such packages keep their README-only page instead of failing the build.
- `component-dirs.json` now holds the layer's component directories and auto-import exclusions. `nuxt.config.ts` reads it, so the docs can use the same source instead of duplicating the rules.
- `cms-base-plugin.ts` builds source links from `projectRootDir`. It used to split on `"frontends/"`, which matched the Developer Portal's `src/frontends/` mount and produced `/_source/...` links that 404.
- `component-dirs.json` added to `files`. Without it the published tarball omits the file and the layer fails on its own config.

## Verification

- Developer Portal healthcheck passes on this branch. That run started at 08:44Z, before shopware/developer-portal#638 merged at 09:19Z, so it passed on this fix alone.
- Generated `components.d.ts` is byte-identical before and after the config refactor, and `SwMedia3D` stays excluded.
- Installed the packed tarball in a scratch project: the layer resolves and 269 components register. Deleting `component-dirs.json` from the install reproduces `Cannot find module './component-dirs.json'`.
- Component permalinks: 70 links, none containing `_source`.

## Follow-up needed in developer-portal

@Isengo1989 FYI:

shopware/developer-portal#638 was merged and added this to `srcExclude` in `.vitepress/config.ts`:

```ts
// frontends packages without source files (e.g., cms-base-layer converted to pure Nuxt layer)
'frontends/packages/cms-base-layer.md',
```

That was a workaround for the same crash. Once this PR lands it has to be removed, otherwise the page stays out of the build and the sidebar link at `apps/docs/.vitepress/sidebar.ts:392` points at a 404.

### DIFF
--- a/packages/cms-base-layer/component-dirs.json
+++ b/packages/cms-base-layer/component-dirs.json
@@ -3,20 +3,17 @@
     {
       "path": "./app/components",
       "pattern": "Sw*",
-      "global": true,
-      "docsLabel": "Store components"
+      "global": true
     },
     {
       "path": "./app/components/public",
       "pathPrefix": false,
-      "global": true,
-      "docsLabel": "CMS components"
+      "global": true
     },
     {
       "path": "./app/components/ui",
       "prefix": "Sw",
-      "global": true,
-      "docsLabel": "UI components"
+      "global": true
     }
   ],
   "excluded": ["SwMedia3D"]
--- a/packages/cms-base-layer/component-dirs.json
+++ b/packages/cms-base-layer/component-dirs.json
@@ -1,0 +1,23 @@
+{
+  "dirs": [
+    {
+      "path": "./app/components",
+      "pattern": "Sw*",
+      "global": true,
+      "docsLabel": "Store components"
+    },
+    {
+      "path": "./app/components/public",
+      "pathPrefix": false,
+      "global": true,
+      "docsLabel": "CMS components"
+    },
+    {
+      "path": "./app/components/ui",
+      "prefix": "Sw",
+      "global": true,
+      "docsLabel": "UI components"
+    }
+  ],
+  "excluded": ["SwMedia3D"]
+}
--- a/packages/cms-base-layer/nuxt.config.ts
+++ b/packages/cms-base-layer/nuxt.config.ts
@@ -3,6 +3,8 @@ import type { NuxtConfig } from "@nuxt/schema";
 import { defineNuxtConfig } from "nuxt/config";
 import type { LoggingFunction, RollupLog } from "rollup";
 
+import componentDirs from "./component-dirs.json";
+
 const { resolve: resolveLayer } = createResolver(import.meta.url);
 
 export default defineNuxtConfig({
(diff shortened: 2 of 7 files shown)


packages/cms-base-layer/component-dirs.json
```diff
@@ -0,0 +1,23 @@
+{
+  "dirs": [
+    {
+      "path": "./app/components",
+      "pattern": "Sw*",
+      "global": true,
+      "docsLabel": "Store components"
```
patzick: `docsLabel` has no consumer on this branch — `git grep docsLabel` only hits this file and the destructure in `nuxt.config.ts` that throws it away. So the PR's "the docs can use the same source" isn't realized yet: `cms-base-plugin.ts` still walks `packages/cms-base-layer/app/components/public/cms` with `readdirSync` and never reads this JSON.

Either wire the plugin to it in this PR, or drop the three keys — as-is they ship in the published tarball as dead metadata and force the `docsLabel: _docsLabel` strip in the config.

packages/cms-base-layer/nuxt.config.ts
```diff
@@ -21,14 +23,9 @@ export default defineNuxtConfig({
...
-      const index = components.findIndex(
-        (c) =>
-          c.pascalName === "SwMedia3D" ||
-          c.kebabName === "sw-media3-d" ||
-          c.filePath?.includes("SwMedia3D.vue"),
-      );
-      if (index > -1) {
-        components.splice(index, 1);
+      for (const name of componentDirs.excluded) {
+        const index = components.findIndex((c) => c.pascalName === name);
```
patzick, line 27: The matching narrowed from three predicates (`pascalName`, `kebabName === "sw-media3-d"`, `filePath?.includes("SwMedia3D.vue")`) to `pascalName` alone. Correct today — `SwMedia3D.vue` sits at `app/components/` root, so its `pascalName` is exactly `SwMedia3D` — but an entry added to `excluded` for a nested component (whose `pascalName` picks up a path prefix) now silently no-ops, and the heavy import it was meant to keep out gets bundled with no warning.

Also `findIndex` + `splice` removes one occurrence per name. `components.splice(...)` in a loop is more fragile than filtering in place:

```ts
for (let i = components.length - 1; i >= 0; i--) {
  if (componentDirs.excluded.includes(components[i].pascalName)) {
    components.splice(i, 1);
  }
}
```

A `console.warn` when an `excluded` name matches nothing would make a future typo loud instead of silent.

patzick (approved):
# 🔍 Code Review

## ✅ Verdict

Approve. The two real fixes are correct: the `existsSync` guard in `plugin.ts` stops the ENOENT that takes down the portal build, and deriving the permalink from `projectRootDir` removes the `_source/` segment that `split("frontends/")` produced under the DevHub mount. `component-dirs.json` in `files` + the changeset are needed, since `nuxt.config.ts` imports it at load time.

## 🟡 Minor

Two minors inline: `docsLabel` has no consumer anywhere on this branch, and the `excluded` loop narrowed its matching. Neither blocks the merge.

## 🤖 Prior review

Copilot's inline note on `cms-base-plugin.ts` was written against `7a54658` and no longer applies: the head revision normalizes separators with `.replaceAll(sep, "/")` and strips a leading `_source/`.

## 🧪 Validation

CI only, per request: all 11 checks pass (Build, Test, lint, CodeQL, Developer Docs). Nothing run locally.

mkucmus:
@Isengo1989 may I ask you to revert this PR's changes: https://github.com/shopware/developer-portal/pull/638 ?

Isengo1989:
> Micha Hobert (@Isengo1989) may I ask you to revert this PR's changes: [shopware/developer-portal#638](https://github.com/shopware/developer-portal/pull/638) ?

Yep, will do. Thanks for fixing it source-side

Isengo1989:
@bojanrajh fyi

PR #2661, opened 2026-08-20
chore(deps): override nanoid to 3.3.18 to fix pnpm audit

patzick:
## Description

The scheduled `Audit check` workflow was failing on a **high** severity finding:

```
high  nanoid: custom generators can loop indefinitely when size is zero
Vulnerable versions  <3.3.18
Patched versions     >=3.3.18
```

[GHSA-2v37-7h3g-55p8](https://github.com/advisories/GHSA-2v37-7h3g-55p8) was widened to cover `nanoid <3.3.18`. The repo already carried an override from #2626, but it was pinned at the previous patch level (`nanoid@<3.3.17: ^3.3.17`), so the selector no longer matched the newly-affected `3.3.17` and 100 transitive paths (mostly `postcss > nanoid`) stayed vulnerable.

## Changes

Bumped the existing override to `nanoid@<3.3.18: ^3.3.18` and refreshed the lockfile. `nanoid@3.3.17` is now fully gone — every consumer (`postcss@8.5.25`, `@sanity/client@7.26.0`) resolves to `3.3.18`, which stays within the `^3.3.x` range they ask for, so this is a patch-level move with no API change.

## Verification

- `pnpm audit` → `No known vulnerabilities found` (was 1 high)
- `pnpm install --frozen-lockfile --prefer-offline` succeeds, all 6 packages build
- `pnpm lint` → 31/31 packages pass

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,7 +8,7 @@ overrides:
   editorconfig>minimatch: ^9.0.9
   '@vercel/blob>undici': ^6.28.0
   less: ^4.8.1
-  nanoid@<3.3.17: ^3.3.17
+  nanoid@<3.3.18: ^3.3.18
   esbuild: ^0.28.1
   vite: ^8.0.16
   '@vueuse/core': 14.4.0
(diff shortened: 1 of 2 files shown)


patzick:
Closing as a duplicate — this was authored from a stale base (`9274aa86`) and #2633 already landed the identical `nanoid@<3.3.18: ^3.3.18` override. `pnpm audit` on current main reports no known vulnerabilities, so there is nothing left to fix here.

Issue #2638, opened 2026-08-17
[CF] Platform & Adoption


mkucmus:
### 🎯 Goal

Guardrails that actually fire, a fast and honest first fifteen minutes, and readiness for the new Content API before it lands.

---

### 🧭 Why this epic matters

**The guards mostly do not run.** The board asked how to add guards so headless is not forgotten. Several already exist on paper and never fire.

- `test:e2e` appears in no workflow. It used to: `e2e-tests.yml` ran it on every preview deployment until it was deleted on 2025-11-07 in #2111, a commit titled "ci: dependency bump and cleanup". Of its 712 non-skipped runs, 506 failed and 206 passed. The suite is unwired, not untried, and a 71% failure rate is why it went.
- `lighthouse.yml` is `workflow_dispatch` only and has never run once. The cron in `accessibility-check.yml` is commented out; its last run was 2025-11-27, and it failed 51 of 147 scheduled runs. Both are watched by the failed-job Slack notifier, while `README.md:111` claims page speed and Lighthouse scores are constantly monitored.
- `Code examples` and `Stackblitz templates` are `disabled_manually` at the GitHub API level since 2025-02-22, 11 seconds apart, with nothing recording why. Re-enabling them needs `gh workflow enable`, not a YAML edit.
- `packages/cms-base-layer` is the only package with coverage enabled and no threshold. 145 tracked `.ts` and `.vue` files, 8 test files, none mounting a component. Any tests written there will not hold the line.
- `packages/composables` advertises a 100% floor but excludes `**/src/useListing/**`, its largest composable and the one driving every product listing.
- No template has a `test` script, and no template participates in `turbo run test`.
- No bundle-size budget exists anywhere, and the only `analyze` script lives in the template that is being deleted.
- The StackBlitz check has no ignore list and resolves against a hard-coded `tree/main`, so a template added in a PR fails until after merge, and any stray directory is auto-enrolled.

[...]

PR #2662, opened 2026-08-20
docs: README and AGENTS files update

mkucmus:
### Description

part of #2638


apps/e2e-tests/README.md
```diff
@@ -2,7 +2,7 @@
 
 ## Run tests
 
-Prepare a `.env` file with the `BASE_E2E_URL` of your local environment (please see the `.env.template` file).
+Prepare a `.env` file with the `BASE_E2E_URL` of your local environment (see `playwright.config.ts`).
```
mdanilowicz: Why not a template .env file?

mdanilowicz (changes requested):

apps/e2e-tests/README.md
mkucmus: that's a great idea. added!

AGENTS.md
```diff
@@ -118,7 +120,8 @@ Packages use workspace protocol for internal dependencies:
 
 Packages must be built in dependency order (handled by Turbo):
 
-- `api-client` → `helpers` → `composables` → `cms-base-layer` → `nuxt-module`
+- `api-client` + `helpers` (independent) → `composables` → `nuxt-module`; `api-gen` builds after `api-client`
+- `cms-base-layer` and `unocss-design-tokens-layer` are plain Nuxt layers with no build step
```
patzick, line 127: `cms-base-layer` does have a build step, so this is wrong and it should not have been dropped from the ordering line above.

`packages/cms-base-layer/package.json` → `"build": "nuxt prepare && unbuild"`, plus a `build.config.ts`, and it depends on `@shopware/api-client`, `@shopware/composables` and `@shopware/helpers` — so root `postinstall` (`pnpm run build --filter='./packages/*'`) does build it, after `composables`. Only `unocss-design-tokens-layer` genuinely has no `build` script (its `build.config.ts` is unused).

Suggestion:

```suggestion
- `cms-base-layer` also builds (`nuxt prepare && unbuild`), after `composables`; only `unocss-design-tokens-layer` has no build step
```

AGENTS.md
```diff
@@ -561,7 +564,7 @@ This creates a file in `.changeset/` - commit it with your changes.
 
 **Main Branch**: `main`
 
-**Current Status**: Clean working directory (as of e06b2f32)
+**Current Status**: Clean working directory (as of 73d6426b)
```
patzick: Minor: this line is stale the moment it merges — working-tree cleanliness is per-checkout, not a repo fact. Bumping the SHA just resets the clock. Worth deleting the line outright and keeping only `**Repository Version**: Based on commit …` below as the provenance marker.

patzick (changes requested):
Verified the doc claims against the code at this head. Almost everything checks out — `pnpm 11.5.2`, Node `20/22/24`, Vite `8.1.4`, Astro `7.1.3`, `build.config.ts`, `src/ApiError.ts`, `src/helpers/`, `src/tests/` (no `_tests/`), `hookable`/`defu`, `ApiClientHooks`/`AdminApiClientHooks`, `GlobalFetchOptions` as a `Pick`, `lcpImagePreload: false` default, `readProductGet get /product`, `readNavigation` wrapper args, `body: { limit: 2 }`, `encodeForQuery(criteria)`, api-gen `filename` = input schema, `/store-api/` endpoint, `unocss-design-tokens-layer` in the starter `extends`, `.lumora-separator` as the only remaining shortcut, coverage script only in `packages/composables`, `generate-types` in every Nuxt template but `vue-blank`, and the new `apps/e2e-tests/.env.template` matching `findEnv()`/`ENV_FILE` and the real defaults. Both earlier asks are addressed (`.env.template` added; README now documents the SSH prerequisite plus an HTTPS fallback).

One new claim is factually wrong — see the inline comment on `AGENTS.md:124`. One minor cleanup suggestion inline too.

AGENTS.md
mkucmus: cms-base-layer has no build step, #2609 dropped both the build script and build.config.ts when it became a plain Nuxt layer. turbo run build --filter='./packages/*' --dry shows Command = <NONEXISTENT> for both layers. Keeping the line as is.

packages/api-client/AGENTS.md
```diff
@@ -16,7 +16,7 @@ pnpm run dev      # Stub mode for development (hot reload)
 pnpm run test     # Run tests
 ```
 
-**Before commits**: `pnpm run lint:fix && pnpm format && pnpm run typecheck`
+**Before commits**: `pnpm run lint:fix && pnpm run typecheck`
```
patzick, line 19: Nit: `pnpm run typecheck` is redundant here — `lint:fix` in this package is `oxlint --fix . && oxfmt --write . && pnpm typecheck`, so this runs typecheck twice. `pnpm run lint:fix` alone covers it.

patzick (approved):
## Re-review — approved

Verified every changed doc claim against this head (`fe8a884`). All of it holds: `pnpm 11.5.2`, Node `20/22/24`, Vite 8, Astro 7.1.3, `build.config.ts` (no `unbuild.config.ts` left anywhere), `src/ApiError.ts` + `src/helpers/` + `src/tests/`, `ApiClientHooks`/`AdminApiClientHooks`, `readProduct post /product` with `body`, `readProductGet get /product` with `query._criteria`, api-gen `filename: storeApiSchema.json` (it is the *input* schema), `lcpImagePreload` default `false`, `.lumora-separator` shortcut, `createShopwareContext` in vue-vite-blank, `coverage` only in `packages/composables`, `generate-types` in every Nuxt template except vue-blank, and the `/store-api/` endpoint suffix (matches what the templates actually configure).

Build-order line is also correct: `helpers` and `api-client` have no workspace deps, `composables` depends on both, `nuxt-module` on all three, `api-gen` on `api-client`.

**My previous blocker was wrong** — I read a stale base. Neither `cms-base-layer` nor `unocss-design-tokens-layer` has a `build` script at this head, so the line as written is accurate. Thanks for pushing back. The stale "Clean working directory" line is gone as suggested.

CI is green on all 9 checks.

One minor out-of-diff follow-up, no need to block on it: `templates/vue-blank/.env.dist` still sets `NUXT_PUBLIC_SHOPWARE_ENDPOINT="https://demo-frontends.shopware.store"` without the `/store-api/` suffix, while its own `nuxt.config.ts` and every other template use the suffix. Since the env var wins over the config value in the module, a user following that template's README ("rename `.env.dist` to `.env`") ends up with a broken endpoint — the same bug this PR fixes in the nuxt-module README.

mdanilowicz (approved):

Issue #2653, opened 2026-08-19
[CF][Templates] Repoint the accessibility and Lighthouse checks to the starter


mkucmus:
Part of #2634

## Goal

The accessibility and Lighthouse workflows check the supported template.

## Current state

`.github/workflows/accessibility-check.yml`:

```
BASE_E2E_URL=https://frontends-demo.vercel.app/
```

`.github/workflows/lighthouse.yml` audits three URLs on the same host, including `/Products/` and `/Summer-BBQ/`, which are demo store category paths.

Both workflows are `workflow_dispatch` only. The accessibility schedule is commented out.

## Ownership question, answered

The deployments come from the shopware Vercel account through the Vercel GitHub integration on this repo. Three projects have Production and Preview environments:

- `frontends-vue-starter-template`
- `frontends-starter-template-extended`
- `old-frontends-demo`

The demo store project has already been renamed with an `old-` prefix. `https://frontends-starter-template.vercel.app/` returns 200. This is not blocked.

## Scope

- Point `BASE_E2E_URL` at `https://frontends-starter-template.vercel.app/`.
- Replace the three Lighthouse URLs. The demo store category paths do not exist on the starter, so pick equivalent routes on the starter's catalogue and record why in `lighthouserc.js`.
- Decide whether to re-enable the commented-out accessibility cron. If yes, uncomment it in this pull request.

## Done when

Neither workflow mentions `frontends-demo.vercel.app`, and a manual dispatch of each is green.

PR #2663, opened 2026-08-24
fix(cms-base-layer): give linked CMS images an accessible name

mkucmus:
### Description

A linked CMS image renders as an `<a>` wrapping only an image. With no alt text the link has no accessible name, which axe reports as a serious `link-name` violation.

The cause is not the missing alt text. Shopware ships `ariaLabel` and `isDecorative` on the CMS image element and `useCmsElementImage` read neither, so an author who filled in the aria label got nothing.

Found while repointing the accessibility check at `vue-starter-template` in #2653.

### Type of change

### What changed

- `useCmsElementImage` reads `ariaLabel` and `isDecorative` and returns them
- `imageAttrs.alt` is empty for a decorative image; `ariaLabel` stays on the link, as in the Storefront
- `CmsElementImage` names a linked image with `ariaLabel`, then the media title, then `cms.image.linkWithoutLabel`
- Both fields are optional on the config; `useCmsElementConfig` accepts optional members
- `SliderElementConfig` gained the `"none"` navigation value

### Please note

"Open linked page" is what makes the check pass today, and it is not useful wording. Every CMS image on the demo homepage has `alt`, `title` and `ariaLabel` empty, so there is nothing better to fall back to. The real fix is to set alt text in the Administration. This PR makes that possible.

### ToDo's

- [x] Changeset file provided
- [ ] Related Issue updated

### Additional context

Verified against a local starter dev server, 3 passed, homepage failed on `link-name` before. Composables 367 passed with 100% coverage held, `cms-base-layer` 73 passed, typecheck and lint clean.

Needs `@shopware/composables` released first. The changeset handles the order.

Pairs with #2653, which stays red until this merges.


.changeset/gentle-moons-shave.md
```diff
@@ -0,0 +1,8 @@
+---
+"@shopware/composables": minor
```
mdanilowicz, line 2: This should be split into two files.

packages/cms-base-layer/app/components/public/cms/element/CmsElementImage.vue
```diff
@@ -27,6 +51,24 @@ const {
...
+ * inside a link still needs the link named. The element's own aria label
+ * survives that, so prefer it over the generic fallback.
+ *
+ * Set alt text or the aria label in the Administration to get something more
+ * useful than the fallback.
+ */
+const linkAriaLabel = computed(() => {
+  if (!imageLink.value.url) return undefined;
+  if (imageAttrs.value.alt) return undefined;
+  return ariaLabel.value || translations.cms.image.linkWithoutLabel;
```
mdanilowicz:
```suggestion
if (imageLink.value.url && !imageAttrs.value.alt) {
  return ariaLabel.value || translations.cms.image.linkWithoutLabel;
}

return undefined;
```

packages/cms-base-layer/app/components/public/cms/element/CmsElementImage.vue
```diff
@@ -27,6 +51,24 @@ const {
...
+ * image's alt text. When the alt text is empty the link ends up nameless,
+ * which axe reports as a serious `link-name` violation.
+ *
+ * A decorative image has no alt text by definition, so a decorative image
+ * inside a link still needs the link named. The element's own aria label
+ * survives that, so prefer it over the generic fallback.
+ *
+ * Set alt text or the aria label in the Administration to get something more
+ * useful than the fallback.
+ */
```
mdanilowicz: Do we need such a long description of how such a simple function works?

mdanilowicz (changes requested):

packages/cms-base-layer/app/components/public/cms/element/CmsElementImage.vue
mkucmus: you are right, no one would read through it ;) I cut it down

.changeset/gentle-moons-shave.md
mkucmus: right, done

packages/composables/src/types/cmsElementTypes.ts
```diff
@@ -62,9 +62,11 @@ type ImageElementConfig = {
   media: ElementConfig<string>;
   newTab: ElementConfig<boolean>;
   product: ElementConfig<string>;
+  ariaLabel: ElementConfig<string | null>;
```
patzick: **Minor** — both new keys are added as *required* members of `ImageElementConfig`, which is also spread into `ImageSliderElementConfig` and reused by `CmsElementManufacturerLogo`.

Two consequences: an older Shopware instance that doesn't return `ariaLabel`/`isDecorative` in the slot config is now mistyped, and every downstream consumer that constructs this object (test fixtures, mocks, custom resolvers) stops compiling — a type-level break shipped under a `minor` bump.

```suggestion
  ariaLabel?: ElementConfig<string | null>;
```
and the same for `isDecorative` on line 69.

packages/composables/src/cms/useCmsElementImage.ts
```diff
@@ -74,9 +76,25 @@ export function useCmsElementImage(
...
+   */
+  const ariaLabel = computed(
+    () => element.data?.ariaLabel || getConfigValue("ariaLabel") || "",
+  );
+
+  /**
+   * The CMS author marked the image as purely decorative, so it must not
+   * expose any alternative text.
+   */
+  const isDecorative = computed(() => !!getConfigValue("isDecorative"));
```
patzick, line 89: **Nit** — asymmetric with `ariaLabel` right above it, which reads `element.data` first and falls back to config. If the resolver surfaces `isDecorative` on `data` the same way it does `ariaLabel` (the new `data.ariaLabel?: string | null` type suggests it does), a decorative image delivered via `data` won't be honoured here. Either add the `data` lookup for symmetry, or drop a one-liner saying it's config-only on purpose.

packages/composables/src/cms/useCmsElementImage.ts
```diff
@@ -74,9 +76,25 @@ export function useCmsElementImage(
...
+
   const imageAttrs: ComputedRef<ImgHTMLAttributes> = computed(() => ({
     src: element.data?.media?.url,
-    alt: element.data?.media?.alt || "",
+    alt: isDecorative.value
+      ? ""
+      : element.data?.media?.alt || ariaLabel.value || "",
```
patzick: **Minor** — worth a second look at the semantics: in the Storefront the element's `ariaLabel` names the *link*, not the image. Copying it into `` means an **unlinked** image whose author wrote "Go to the summer collection" now renders that as its alternative text, and any shop that already filled `ariaLabel` sees its rendered markup change on upgrade.

The changeset documents it, so I'll treat it as deliberate — but scoping the fallback to the linked case would avoid the surprise entirely.

packages/cms-base-layer/app/components/public/cms/element/CmsElementImage.vue
```diff
@@ -27,6 +51,16 @@ const {
   mimeType,
 } = useCmsElementImage(props.content);
 
+// A link wrapping only an image is nameless without alt text, which axe
+// reports as a `link-name` violation.
+const linkAriaLabel = computed(() => {
```
patzick, line 57: **Nit** — `media.title` is missing from the fallback chain. It's a real field authors fill in the media manager, and it would produce a meaningful link name in cases that currently land on the generic string. Ordering `media.alt → ariaLabel → media.title → linkWithoutLabel` costs one `||` and shrinks how often every linked image on a page shares the identical accessible name "Open linked page".

patzick (approved):
## 🔍 Code Review — approve

The fix is correct and minimal: `ariaLabel`/`isDecorative` are read, `alt` is suppressed for decorative images, and a linked image can no longer be nameless. CI is green on all 10 checks. No blockers.

Four minor points left inline — the type-optionality one is the only one I'd like addressed before merge; the rest are judgement calls.

Prior feedback from @mdanilowicz (split changeset, shortened comment, `ariaLabel` before the generic fallback) is all present on the head commit — that review can be dismissed.

mdanilowicz (approved):

Issue #2649, opened 2026-08-19
[CF][Templates] Run the e2e suite in CI against vue-starter-template


mkucmus:
Part of #2634

## Goal

The e2e suite runs in CI against `vue-starter-template`.

## Why

`test:e2e` is declared in `turbo.json` and in `apps/e2e-tests/package.json`, and appears in no workflow. The suite has never run in CI. Everything else in this milestone is unverifiable until it does.

## The tag problem

`apps/e2e-tests/package.json`:

```json
"test:e2e": "playwright test --grep @vue-demo-store"
```

The grep tag is the demo store. This is not only missing CI wiring; the selector itself has to change. Retag the specs that should survive the demo store deletion, or move to a tag that names the behaviour rather than the template.

## Scope

- Retag the suite away from `@vue-demo-store`.
- Add a workflow that builds and starts `vue-starter-template`, then runs the suite against it.
- Decide preview deployment versus local `nuxt build` + `nuxt preview` in the job. Local keeps the run hermetic and does not depend on Vercel.
- Run on pull requests that touch `templates/vue-starter-template/**` or `apps/e2e-tests/**`, plus nightly.

## Note

The suite will fail until the missing `data-testid` attributes land and the login and register page objects are rebuilt. Land this workflow first, allow it to be red, and treat the other two issues as the fix.

## Done when

A workflow file runs `test:e2e` against the starter and reports a status on pull requests.

PR #2664, opened 2026-08-24
ci: point the accessibility and Lighthouse checks at vue-starter-template

### LINKED ISSUE #2653: [CF][Templates] Repoint the accessibility and Lighthouse checks to the starter

mkucmus:
### Description

closes #2653

The accessibility and Lighthouse workflows measured `frontends-demo.vercel.app`, the deprecated demo store. They now measure `vue-starter-template`.

`/Products/` and `/Summer-BBQ/` are not a rename away. The starter runs a different sales channel and both return 404 on it, so Lighthouse audits `/Furniture/` and `/Sale/` instead. The reason is recorded in `lighthouserc.js`, which held a second copy of the URL list that the issue missed.

Two pre-existing breaks had to go first, or the repointed workflow could not run at all:

- `apps/e2e-tests` imported `find-up`, declared in no package.json. It resolved to a hoisted CommonJS copy, so the named import was undefined and `playwright.config.ts` threw before any test started.
- The accessibility specs navigated via demo store content, a "Products" menu item and a "YORK 3" product. Neither exists on the starter.

Also dropped a `test.describe.only` that would suppress every other spec in a full run. That one would have bitten #2649.

The nightly cron stays commented out. This job has never been green, and a schedule that is red every morning just gets ignored. The reason is written into the workflow.

### Type of change

### Added after the first review

**The workflow name was misspelled "Accessibiliy".** Copilot flagged it and named the coupling: `failed-job-check.yml` matches the workflow by exact name in its `workflow_run` list, so a rename without it would silently stop the Slack failure notification. Both files change together, plus the one docs line that repeated the typo.

**The check can now target any storefront.** `BASE_E2E_URL` was hardcoded, so dispatching on a branch ran that branch's code against production and told you nothing about the branch. A `base_url` dispatch input fixes that:

```sh
gh workflow run accessibility-check.yml --ref <branch> -f base_url=<vercel-preview-url>
```

It defaults to the deployed starter, so a plain dispatch behaves as before. The value is passed through `env:` rather than interpolated into the run script, so a crafted input cannot inject shell.

### ToDo's

- [ ] Related Issue updated

No changeset. Nothing here is a published package.

### Additional context

Verified by running it, not by reading it.

Against the deployed starter, 2 of 3 pass. The homepage fails on a real serious `link-name` violation, which is the check doing its job. #2663 fixes it.

Against #2663's Vercel preview, which carries that fix, all 3 pass:

```sh
cd apps/e2e-tests
BASE_E2E_URL=<2663-preview-url> \
  pnpm exec playwright test --grep @accessibility --project=chromium

3 passed
```

So this workflow stays red until #2663 merges, and goes green immediately after.

`apps/e2e-tests/README.md` documents all three ways to run it: the deployed starter, a pull request preview, and a local build.

### DIFF
--- a/.github/workflows/accessibility-check.yml
+++ b/.github/workflows/accessibility-check.yml
@@ -3,6 +3,9 @@ name: Accessibiliy Check
 run-name: Playwright accessibility tests 🚀
 on:
   workflow_dispatch:
+  # Kept on manual dispatch for now. This job has never run green in CI, so a
+  # nightly schedule would only produce a red run every morning. Re-enable the
+  # cron once a manual dispatch passes against the starter template.
   # schedule:
   #   - cron: "30 09 * * *"
 jobs:
@@ -31,4 +34,4 @@ jobs:
       - name: Run tests
         run: |
           cd apps/e2e-tests
-          BASE_E2E_URL=https://frontends-demo.vercel.app/ npx playwright test --grep @accessibility --project=chromium
+          BASE_E2E_URL=https://frontends-starter-template.vercel.app/ npx playwright test --grep @accessibility --project=chromium
--- a/.github/workflows/accessibility-check.yml
+++ b/.github/workflows/accessibility-check.yml
@@ -39,8 +39,12 @@ jobs:
       - name: Run tests
         # Passed as an env var, not interpolated into the script, so a crafted
         # input cannot inject shell.
+        #
+        # The fallback is not redundant with the input default: `inputs` is empty
+        # on a schedule and on a dispatch with a cleared field, and an empty
+        # BASE_E2E_URL points playwright.config.ts at localhost.
         env:
-          BASE_E2E_URL: ${{ inputs.base_url }}
+          BASE_E2E_URL: ${{ inputs.base_url || 'https://frontends-starter-template.vercel.app/' }}
         run: |
           cd apps/e2e-tests
           npx playwright test --grep @accessibility --project=chromium
--- a/.github/workflows/accessibility-check.yml
+++ b/.github/workflows/accessibility-check.yml
@@ -1,4 +1,4 @@
-name: Accessibiliy Check
+name: Accessibility Check
 
 run-name: Playwright accessibility tests 🚀
 on:
--- a/.github/workflows/accessibility-check.yml
+++ b/.github/workflows/accessibility-check.yml
@@ -3,6 +3,11 @@ name: Accessibility Check
 run-name: Playwright accessibility tests 🚀
 on:
   workflow_dispatch:
+    inputs:
+      base_url:
+        description: "Storefront to test. Pass a pull request's Vercel preview URL to check that branch's deployment."
+        required: false
+        default: "https://frontends-starter-template.vercel.app/"
   # Kept on manual dispatch for now. This job has never run green in CI, so a
   # nightly schedule would only produce a red run every morning. Re-enable the
   # cron once a manual dispatch passes against the starter template.
@@ -32,6 +37,10 @@ jobs:
           pnpm playwright install --with-deps
 
       - name: Run tests
+        # Passed as an env var, not interpolated into the script, so a crafted
+        # input cannot inject shell.
+        env:
+          BASE_E2E_URL: ${{ inputs.base_url }}
         run: |
           cd apps/e2e-tests
-          BASE_E2E_URL=https://frontends-starter-template.vercel.app/ npx playwright test --grep @accessibility --project=chromium
+          npx playwright test --grep @accessibility --project=chromium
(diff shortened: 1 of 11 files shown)


.github/workflows/accessibility-check.yml
```diff
@@ -29,6 +37,10 @@ jobs:
           pnpm playwright install --with-deps
 
       - name: Run tests
+        # Passed as an env var, not interpolated into the script, so a crafted
+        # input cannot inject shell.
+        env:
+          BASE_E2E_URL: ${{ inputs.base_url }}
```
patzick: `inputs.base_url` only exists on `workflow_dispatch`. The comment eight lines up invites re-enabling the cron — the moment that happens this resolves to `""`, and `playwright.config.ts` does `process.env.BASE_E2E_URL || "http://127.0.0.1:8080"`, so the scheduled run would silently audit nothing on localhost instead of failing loudly.

```yaml
          BASE_E2E_URL: ${{ inputs.base_url || 'https://frontends-starter-template.vercel.app/' }}
```

apps/e2e-tests/page-objects/HomePage.ts
```diff
@@ -65,6 +65,34 @@ export class HomePage extends AbstractPage {
...
+   * openCategoryPage and openCartPage above select on vue-demo-store content
+   * ("Products", "YORK 3") that does not exist in vue-starter-template.
+   */
+  async openFirstCategoryPage() {
+    const firstCategory = this.page
+      .getByRole("menubar")
+      .getByRole("menuitem")
+      .first();
+    await firstCategory.waitFor({ state: "visible" });
+    await firstCategory.click();
```
patzick: This binds the a11y run to whatever the demo sales channel happens to put first in the nav (today `New Arrivals`). Two content-driven ways it breaks with an opaque 30s timeout rather than a useful message:

- the first entry becomes a landing page with no product cards;
- `TopNavigation.vue` sets `target="_blank"` for `externalLink`/`linkNewTab` entries, so a click opens a new tab and this `page` never navigates.

Not worth blocking, but scoping the locator to same-tab entries, or falling back to the next menuitem when no product card shows up, would keep the failure honest.

lighthouserc.js
```diff
@@ -2,10 +2,18 @@ module.exports = {
...
       numberOfRuns: 5,
+      // vue-starter-template is the supported template. vue-demo-store is
+      // deprecated and gets deleted by 2026-12-31, so its deployment is not
+      // worth measuring any more.
+      //
+      // The old paths /Products/ and /Summer-BBQ/ are not a rename away: the
+      // starter runs a different sales channel with a different catalogue, and
+      // both paths return 404 on it. These three are the closest equivalents,
+      // one home page and two category listings, and all return 200 today.
       url: [
```
patzick, line 13: `treosh/lighthouse-ci-action`'s `urls:` input wins over `collect.url`, so this list is dead config while `lighthouse.yml` also carries it. The README tells the next person to change both together — dropping `urls:` from the workflow and leaving this file as the single source removes the chance to get that wrong.

patzick (approved):
Approving. Repointing is correct and verified: all three Lighthouse URLs return 200, `product-box-product-name-link` / `product-quantity` come from `@shopware/cms-base-layer` which the starter extends, the live menubar exposes exactly one `role="menubar"`, and the `Accessibiliy` → `Accessibility` rename is mirrored in `failed-job-check.yml`, so the Slack notification keeps firing. `find-up` is declared in no `package.json` in the repo — the replacement is a real fix, not a style change.

Three non-blocking notes inline: an empty `BASE_E2E_URL` the moment the cron is re-enabled, the first-menu-item assumption in `openFirstCategoryPage`, and the duplicated Lighthouse URL list.

Issue #2652, opened 2026-08-19
[CF][Templates] Port the remaining demo-store features to vue-starter-template


mkucmus:
Part of #2634

## Goal

`vue-starter-template` covers the demo store behaviour that is still missing, so deleting the demo store loses nothing.

## The gap, measured

Present in `vue-demo-store`, absent from `vue-starter-template`:

| What | Where in the demo store |
| --- | --- |
| Global error page | `app/error.vue` |
| Maintenance mode screen | `app/components/errors/MaintainMode.vue` |
| Routing error components | `app/components/errors/` |
| Password recovery | `app/pages/account/recover.vue` |
| Password reset | `app/pages/reset-password.vue` |
| Sitemap routes | `server/routes/sitemap.xml.ts`, `server/routes/sitemap-local.xml.ts`, `server/sitemap.ts` |

The starter's `server/` holds only `routes/account/login/imitate-customer.post.ts` and an empty `middleware/`.

## Not a straight copy

The starter is Nuxt 4, uses `@regle/core` for validation where the demo store uses `@vuelidate/core`, and routes everything through one `[...all].vue` catch-all. Recovery and reset need real routes added, and the forms need porting to Regle rather than lifted as-is.

## Scope

Port each row above, add the `data-testid` attributes the e2e suite expects on the new components, and add a changeset.

## Done when

Every row has an equivalent in the starter and the related e2e specs pass.

Issue #2651, opened 2026-08-19
[CF][Templates] Rebuild login and register page objects for the starter


mkucmus:
Part of #2634

## Goal

Login and register page objects work against the starter's navigation model.

## Why this is not a URL rename

`vue-demo-store` pages:

```
app/pages/login.vue
app/pages/register.vue
app/pages/reset-password.vue
app/pages/account/recover.vue
```

`vue-starter-template` pages:

```
app/pages/account/login.vue
```

The starter has no `/login`, no `/register`, and no `/reset-password` route. Registration is rendered inline on `/account/login`. A test that navigates to `/register`, fills a form and submits gets a 404 today.

## Scope

- Rewrite the register page object to reach registration through `/account/login` rather than a route of its own.
- Rewrite the login page object for the new path and markup.
- Any test that asserts a redirect to or from `/login` or `/register` needs its expected URL updated.

## Depends on

The password recovery and reset pages are part of the feature port. Page objects covering them cannot be finished until that lands. Split them out rather than blocking this issue.

## Done when

The register and login specs pass against a locally built `vue-starter-template`.

Issue #2650, opened 2026-08-19
[CF][Templates] Add the 48 missing data-testid attributes to vue-starter-template


mkucmus:
Part of #2634

## Goal

`vue-starter-template` carries every `data-testid` the e2e suite selects on.

## Numbers, measured

- 64 unique test ids are used across `apps/e2e-tests`.
- 45 unique `data-testid` attributes exist in `templates/vue-starter-template/app`.
- **48** ids used by the suite are missing from the starter.
- For reference, `vue-demo-store` carries 159.

The epic said 38. The measured number is 48.

## Sample of what is missing

`add-to-cart-button`, `cart-button`, `cart-checkout-link`, `cart-close-button`, `cart-product-image`, `cart-product-options`, `account-menu-hello-button`, `account-personal-data-email-input`, `account-personal-data-firstname-input`, `account-personal-data-lastname-input`, `checkout-pi-city-input`, `checkout-pi-email-input`, `checkout-pi-first-name-input`, `checkout-pi-last-name-input`.

Reproduce the full list:

```bash
grep -rho 'getByTestId("[^"]*")' apps/e2e-tests | sed 's/getByTestId("//;s/")//' | sort -u > /tmp/e2e_ids.txt
grep -rho 'data-testid="[^"]*"' templates/vue-starter-template/app | sed 's/data-testid="//;s/"//' | sort -u > /tmp/starter_ids.txt
comm -23 /tmp/e2e_ids.txt /tmp/starter_ids.txt
```

## Watch out

One id in the suite is `#newsletter-checkbox`, selected as a test id but written as a CSS id. Fix the selector, do not add an attribute with a `#` in it.

Some ids will have no home until the ported features land. Those belong to the port issue, not here. List them in a comment when you hit them.

## Done when

`comm -23` over the two files returns only ids whose components do not exist yet.

mkucmus:
The `comm -23` check in the description undercounts what is done, so the number here will never reach zero as written. Two fixes to the measurement, then the real residue.

**It greps only `templates/vue-starter-template/app`.** Ten of the ids the suite selects on live in `packages/cms-base-layer/app`, which the starter extends as a layer. They render fine and the specs pass: `add-to-cart-button`, `listing-pagination-limit-select`, `product-box-img`, `product-box-product-name-link`, `product-box-toggle-wishlist-button`, `product-variant`, `review-empty-star`, `review-submit-button`, `review-text-input`, `review-title-input`.

[...]

PR #2666, opened 2026-08-24
fix(vue-starter-template): make test ids reach the right element

mkucmus:
### Description

Part of #2650.

Every `registration-*-input` test id in `vue-starter-template` selects a `<div>`. `FormInputField` and `FormDropdownField` have a root `<div>`, so a plain `data-testid` attribute falls through to that root and never reaches the `<input>` or `<select>` inside `BaseInput` / `BaseDropdown`.

`country-select` was set twice, on the wrapper in `CountryStateInput` and on the inner input by `CountrySearchSelect`'s own default, so Playwright strict mode throws on it.

Measured in a real browser against a local dev server, before and after.

Before:

```
registration-first-name-input      div
registration-account-type-select   div
country-select                     div, input   <-- 2 matches, strict mode throws
fill() on first-name FAILED: Element is not an <input>, <textarea>, <select> ...
```

After:

```
registration-first-name-input      input
registration-account-type-select   select
country-select                     input
fill() on first-name -> "Ada"
```

### Type of change

### What changed

- `BaseInput` and `BaseDropdown` take a `dataTestId` prop, bound on the `<input>` and the `<select>`
- `InputField` and `DropdownField` forward it
- `SalutationSelect` and `AccountTypeSelect` forward it as a prop instead of as a `data-testid` attribute, and no longer default it to `""`
- `CountryStateInput` uses the prop, which removes the duplicate `country-select`
- The 9 `registration-*-input` attributes in `RegistrationForm` become `dataTestId`

`registration-form` sits on a real `<form>` and `registration-submit-button` on `BaseButton`, whose root is a `<button>`. Both already worked and are untouched.

No new test ids are added here. That is the rest of #2650.

### Second commit: two duplicate test ids

Playwright runs `getByTestId` in strict mode, so a test id that matches twice throws instead of picking one. Two cases were mislabelled rather than merely duplicated:

- `SummaryBox.vue` used `cart-subtotal` on both the subtotal and the total. The total is now `cart-total`.
- The order success page used `order-subtotal` on `order.amountTotal` in the summary header, and again on the real subtotal further down. Both render under the same `v-if`. The header one is now `order-summary-total`.

Neither old id is selected by any spec, so nothing breaks.

Checked and deliberately left alone:

- `order-item-unitprice` and `order-item-totalprice` repeat per line item. They are meant to be multi-match and need `.nth()`.
- The success page skeleton reuses `order-subtotal` and `order-total`, but it lives in `<template #placeholder>` so it never renders beside the real content.
- `PriceSummary` shares those ids but only renders on the account order pages, never on checkout success.
- `CountrySearchSelect` defaults `dataTestId` to `country-select`, so every instance emits it. All three call sites render once per page today. A page with billing and shipping forms would produce two. The default is load-bearing: `CheckoutPage.ts` and `RegisterPage.ts` both select it and checkout's `CustomerAddress` relies on it, so it should not simply be removed.
- `CountryStateInput` hardcodes `id="country"` and `id="state"`. Two instances would give duplicate DOM ids and break `<label for>`. No page renders two today.

### ToDo's

- [ ] Related Issue updated

No changeset. `vue-starter-template` is private and unpublished.

### Additional context

`registration-vatid-input`, `registration-company-input` and `checkout-pi-state-input` do not render on load. They are conditional on a business account type and on a country that has states. They go through the same components, so the same fix covers them.

This gives #2650, #2651 and #2652 a one-line-per-field pattern. Without it every test id those issues add is dead on arrival.

### DIFF
--- a/templates/vue-starter-template/app/components/form/BaseInput.vue
+++ b/templates/vue-starter-template/app/components/form/BaseInput.vue
@@ -18,6 +18,8 @@ const {
   invalid?: boolean;
   autocomplete?: string;
   id?: string;
+  /** Lands on the <input>. A plain data-testid attribute would fall through to the root element. */
+  dataTestId?: string;
 }>();
 
 const model = defineModel<string>({
@@ -51,6 +53,7 @@ const emit = defineEmits<{
         :placeholder="placeholder"
         :type="type"
         :autocomplete="autocomplete"
+        :data-testid="dataTestId"
         @focus="emit('focus')"
       />
       <slot name="rightIcon" />
--- a/templates/vue-starter-template/app/components/form/BaseInput.vue
+++ b/templates/vue-starter-template/app/components/form/BaseInput.vue
@@ -8,6 +8,8 @@ type InputTypeAttribute =
   | "time"
   | "url";
 
+defineOptions({ inheritAttrs: false });
+
 const {
   placeholder,
   type = "text",
@@ -18,8 +20,6 @@ const {
   invalid?: boolean;
   autocomplete?: string;
   id?: string;
-  /** Lands on the <input>. A plain data-testid attribute would fall through to the root element. */
-  dataTestId?: string;
 }>();
 
 const model = defineModel<string>({
@@ -34,9 +34,12 @@ const slots = defineSlots<{
 const emit = defineEmits<{
   focus: [];
 }>();
+
+const { wrapperAttrs, controlAttrs } = useControlAttrs();
 </script>
 <template>
   <div
+    v-bind="wrapperAttrs"
     class="focus-within:outline-2 focus-within:outline-outline-outline-focus focus-within:outline focus-within:outline-offset-[2px] rounded-lg"
   >
     <div
@@ -53,7 +56,7 @@ const emit = defineEmits<{
         :placeholder="placeholder"
         :type="type"
         :autocomplete="autocomplete"
-        :data-testid="dataTestId"
+        v-bind="controlAttrs"
         @focus="emit('focus')"
       />
       <slot name="rightIcon" />
--- a/templates/vue-starter-template/app/components/shared/CountryStateInput.vue
+++ b/templates/vue-starter-template/app/components/shared/CountryStateInput.vue
@@ -63,7 +63,7 @@ watch(countryId, (newValue, oldValue) => {
       :placeholder="$t('form.chooseState')"
       :label="$t('form.state')"
       :options="stateOptions"
-      dataTestId="checkout-pi-state-input"
+      data-testid="checkout-pi-state-input"
       :errorMessage="stateIdValidation?.$errors[0]"
     />
   </div>
--- a/templates/vue-starter-template/app/components/shared/CountryStateInput.vue
+++ b/templates/vue-starter-template/app/components/shared/CountryStateInput.vue
@@ -49,7 +49,7 @@ watch(countryId, (newValue, oldValue) => {
       v-model="countryId"
       :label="$t('form.country')"
       :placeholder="$t('form.chooseCountry')"
-      data-testid="country-select"
+      dataTestId="country-select"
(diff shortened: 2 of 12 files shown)


templates/vue-starter-template/app/components/form/BaseInput.vue
```diff
@@ -18,6 +18,8 @@ const {
   invalid?: boolean;
   autocomplete?: string;
   id?: string;
+  /** Lands on the <input>. A plain data-testid attribute would fall through to the root element. */
+  dataTestId?: string;
```
patzick: Non-blocking, worth considering as a follow-up.

A dedicated `dataTestId` prop fixes exactly one attribute. `name`, `required`, `maxlength`, `aria-describedby` and every future `data-*` still fall through to the wrapper `<div>` and hit the same bug this PR is fixing — and they fail the same silent way, only at Playwright run time.

`defineOptions({ inheritAttrs: false })` plus `v-bind="$attrs"` on the `<input>` (same for the `<select>` in `BaseDropdown`) solves the whole class at once, lets call sites keep writing plain `data-testid`, and removes the manual forwarding you had to add in `InputField`, `DropdownField`, `SalutationSelect` and `AccountTypeSelect`.

As it stands nothing stops the next call site writing `data-testid=` again, and there is no lint rule or test to catch it.

templates/vue-starter-template/app/components/shared/CountryStateInput.vue
```diff
@@ -63,7 +63,7 @@ watch(countryId, (newValue, oldValue) => {
       :placeholder="$t('form.chooseState')"
       :label="$t('form.state')"
       :options="stateOptions"
-      data-testid="checkout-pi-state-input"
+      dataTestId="checkout-pi-state-input"
```
patzick: Out of scope for this PR, but flagging it for #2650/#2651 since it is adjacent.

Three page-object locators still resolve to nothing in `vue-starter-template`:

- `checkout-pi-state-input` (`CheckoutPage.ts:38`) only renders through this component, but the checkout page uses `CheckoutCustomerAddress`, not `SharedCountryStateInput` — so it never appears on checkout.
- `checkout-pi-salutation-select` (`CheckoutPage.ts:30`) and `registration-salutation-select` (`RegisterPage.ts:26`) do not exist anywhere in the template. `RegistrationForm` renders no `SalutationSelect` at all.

The `dataTestId` prop this PR adds makes each of those a one-line fix, so it is a good handover point.

patzick (approved):
Approving. The fix is correct and I verified the claims rather than taking them on trust.

What I checked:

- `dataTestId` is a declared prop on `BaseInput`/`BaseDropdown` and bound on the `<input>`/`<select>`, forwarded unbroken through `InputField`/`DropdownField` and `SalutationSelect`/`AccountTypeSelect`. Static camelCase attributes (`dataTestId="..."`) do resolve as props in SFC templates, so all 9 `registration-*` ids land on the real control.
- Dropping the `dataTestId = ""` defaults is what stops the empty `data-testid=""` appearing on the wrapper; the two call sites that never pass it (`account/address/Form.vue`, `account/personal/DataForm.vue`) now render no attribute at all.
- `country-select` really was doubled: the wrapper got it by fallthrough and `CountrySearchSelect`'s own default put it on the input. Only the input keeps it now. The default is indeed load-bearing — `CustomerAddress.vue` is a second call site that relies on it, and the three render sites (checkout, registration, address form) are one-per-page today.
- `cart-subtotal`/`cart-total` in `SummaryBox` both render together, so that was a genuine strict-mode duplicate. Confirmed no spec or page object selects the old id.
- The success-page `order-subtotal`/`order-total` skeleton pair sits inside `<template #placeholder>`, so it never renders beside the real summary. `createOrder.spec.ts` asserting `order-total` `toHaveCount(1)` stays valid.

CI is green across all 10 checks and the branch is mergeable. Two non-blocking notes inline.

PR #2667, opened 2026-08-25
docs: tell people not to scaffold from the demo store template

mkucmus:
### Description

Relates to #2654, with one change of direction: **`vue-demo-store` is not being deleted.** The portfolio grows by adding new templates instead, and the demo store stays in the repository as a reference implementation.

So this PR is the whole answer to the problem #2654 describes, not a first step towards a deletion.

The problem was that the fastest way to start a project on the demo store was two clicks from its own README. It carried an "Open in StackBlitz" badge and a "Deploy with Vercel" button, and the docs page carried an `npx tiged` command. Meanwhile the word "deprecated" appeared in exactly one place in the whole repository.

Every notice now leads with the instruction: do not use this template to start a new project, use `vue-starter-template`. No removal date and no removal promise, because there is no removal. The notices say what the template is for now, which is reading, not building on.

### Type of change

Bug fix (non-breaking change that fixes an issue)

### What changed

**Demo store, stop it being a starting point**

- `templates/vue-demo-store/README.md` leads with the instruction, and the StackBlitz badge and Vercel deploy button are gone
- `introduction/templates/demo-store-template.md` gains a `:::warning DEPRECATED` block, and loses the StackBlitz link and the `npx tiged` command
- `introduction/templates.md` already said DEPRECATED. It now says what to do instead of only what we recommend

Kept on purpose: the `Setup & run`, `Configure` and `Limitations` headings. Three places link to those anchors. `Setup & run` now points at the starter template instead of holding scaffolding commands. The `stackblitz.png` asset stays, `astro-template.md` and `blank-template.md` still import it.

**Starter template, make it worth pointing at**

If we send everyone to `vue-starter-template`, its README has to hold up. It did not.

- 71 of 166 lines, 42% of the file, were untouched `nuxt init` scaffolding, with every command repeated for npm, pnpm, yarn and bun
- the two values the template cannot start without, `NUXT_PUBLIC_SHOPWARE_ENDPOINT` and `NUXT_PUBLIC_SHOPWARE_ACCESS_TOKEN`, first appeared on line 86 as Docker flags. `packages/nuxt-module/plugin.ts:89` throws on startup without them
- it had no `What's inside`, `Requirements`, `Set up your Shopware 6 instance` or `Try it online` section, which the other four template READMEs all share
- it was the only template README with no link to developer.shopware.com, and it never said how to scaffold the template out of the monorepo

[...]

Issue #2648, opened 2026-08-19
[CF][Templates] Refactor the five template lists to read the manifest


mkucmus:
Part of #2634
Blocked by the manifest issue.

## Goal

Every consumer of the template list reads `templates/manifest.json`.

## The five consumers today

1. `apps/e2e-tests/tests/checkStackblitzTemplates.spec.ts` - `fs.readdirSync` over `templates/`, so it opens a StackBlitz test for every directory including the untracked `demo-from-figma` and the unscaffoldable `vue-starter-template-extended`.
2. `apps/docs/.vitepress/theme/components/TryItOutPicker.vue` - `templatesById` is hard-coded, 2 of 6.
3. `apps/docs/.vitepress/sidebar.ts` - 7 entries under "Setup Templates". Two of them (`custom-vue-project`, `custom-react-project`) have no directory; they are integration guides, not templates.
4. `.devcontainer/` - 2 directories, `vue-starter-template` and `vue-starter-template-extended`. Codespaces links in `TryItOutPicker` point at `.devcontainer/${t}/devcontainer.json`, so this list and the picker list must stay in sync.
5. `pnpm-workspace.yaml` - the inert `templates:` key, removed in the manifest issue.

## Scope

- The StackBlitz spec iterates the manifest, filtered to `scaffoldable: true`. This drops `vue-starter-template-extended` from the nightly run, which currently cannot pass.
- `TryItOutPicker` builds its template cards from the manifest. Keep the per-template AI prompts as separate content keyed by id.
- The sidebar generates its template children from the manifest and keeps the two guide entries as explicit non-template links.
- A devcontainer exists for every `scaffoldable: true` template, checked by the drift test.

## Done when

`grep -r "vue-starter-template"` returns no second hard-coded template list outside the manifest and its schema.

mkucmus:
From the review of #2670: the StackBlitz nightly now runs only `scaffoldable: true` templates, which drops vue-starter-template-extended - but TryItOutPicker still renders a live **Open in StackBlitz** button for it (`buildOpenUrl` has no per-template exclusion), and its Local-clone flow offers a plain `tiged` command that produces a broken install. Before #2670 the permanently-red nightly was the only signal; now nothing reports it. When the picker reads the manifest, hide StackBlitz and the plain scaffold command for non-scaffoldable templates.

mkucmus:
Done. Every consumer reads `templates/manifest.json`.

| consumer | now |
| --- | --- |
[...]

PR #2670, opened 2026-08-27
feat(templates): add a manifest and a drift check

### LINKED ISSUE #2647: [CF][Templates] Add templates/manifest.json with a schema and a drift test
Part of #2634

## Goal

One machine-readable file describes the template portfolio. A test fails when it drifts from disk.

## Why

The template list lives in five places today and they already disagree. Before any consumer can be refactored, there has to be one file to point them at.

## Scope

Add `templates/manifest.json` and `templates/manifest.schema.json`.

Per template: `id`, `framework`, `purpose`, `supportLevel`, `scaffoldable`, `deployable`, `requiredEnv`, `buildCommand`, `startCommand`.

Agreed support levels:

| id | supportLevel | scaffoldable | deployable |
| --- | --- | --- | --- |
| `vue-starter-template` | `supported` | yes | yes |
| `vue-starter-template-extended` | `supported` | **no** | yes |
| `astro` | `example` | yes | no |
| `vue-blank` | `example` | yes | no |
| `vue-vite-blank` | `example` | yes | no |
| `vue-demo-store` | `deprecated` | yes | yes |

`vue-starter-template-extended` is `scaffoldable: false` on purpose. Its `package.json` is named `lumora-demo-store` and depends on `"vue-starter-template": "workspace:*"`, so `npx tiged` on it produces a broken install outside this monorepo. The `TryItOutPicker` AI prompt already documents a manual workaround; the manifest has to state the constraint so consumers stop offering it as a plain scaffold.

`example` means: builds in CI, no feature work, no support promise.

## Also

- `templates/demo-from-figma` is untracked and holds one empty `shopware-assets/` folder. Delete it locally, add nothing to git.
- Remove the inert `templates:` key from `pnpm-workspace.yaml`. It is not a pnpm field and nothing reads it.

## Test

A vitest spec that reads `templates/` from disk and fails when the directory listing and the manifest ids disagree in either direction.

## Done when

- `templates/manifest.json` exists and validates against its schema.
- The drift test fails on an added or removed directory.
- `pnpm-workspace.yaml` has no `templates:` key.

mkucmus:
### Description

closes #2647

The template list lived in five places and they had already drifted apart:

- `apps/e2e-tests/tests/checkStackblitzTemplates.spec.ts` reads `templates/` off disk, so the nightly StackBlitz run opens a test for every directory, including `vue-starter-template-extended`, which cannot pass
- `apps/docs/.vitepress/theme/components/TryItOutPicker.vue` hardcodes its own set
- `apps/docs/.vitepress/sidebar.ts` has 7 entries under Setup Templates, two of which have no directory
- `.devcontainer/` has 2
- `pnpm-workspace.yaml` had an inert `templates:` key that nothing read

`templates/manifest.json` is now the one list. This PR adds it and the check that keeps it honest. Refactoring the five consumers to read it is #2648, which was blocked on this.

### What changed

- `templates/manifest.json` - six entries, one per directory on disk
- `templates/manifest.schema.json` - JSON schema, so editors autocomplete and validate the manifest
- `scripts/checkTemplatesManifest.ts` - the drift and schema check, run with `pnpm run check:templates`
- `.github/workflows/templates-manifest.yml` - the check runs as its own workflow. It cannot live in code-check.yml: that workflow ignores `apps/**`, and half of this check resolves docsUrls into `apps/docs/src/`. A PR that only deletes a docs page must fail here, not on the next unrelated PR.
- `apps/e2e-tests/tests/checkStackblitzTemplates.spec.ts` now reads the manifest and opens only `scaffoldable` templates. It used to enumerate every entry of `templates/`, so this PR's two new files would have failed the nightly StackBlitz cron; driving it from the manifest also drops vue-starter-template-extended, which cannot pass there by design.
- each template's `package.json` gets `pnpm.onlyBuiltDependencies` (esbuild, vue-demi and friends), so a fresh pnpm 11 scaffold runs unattended instead of blocking on `pnpm approve-builds`
- removed the inert `templates:` key from `pnpm-workspace.yaml`
- `AGENTS.md` documents the manifest at the head of Working with Templates. Also fixed the stale Template Configuration section, which named `SHOPWARE_ENDPOINT` and `SHOPWARE_ACCESS_TOKEN`, env vars nothing reads, and the pnpm/Node versions.

### The fields

The root carries `manifestVersion` (bumped when field semantics change, so out-of-repo consumers can detect a contract they do not understand) and a `notes` string with the scaffold recipe. Each entry: `id`, `displayName`, `packageName`, `framework`, `purpose`, `docsUrl`, `issueUrl`, `supportLevel`, `scaffoldable`, `scaffoldCommand`, `deployable`, `devcontainer`, `env`, `node`, `devPort`, `buildCommand`, `devCommand`, `typeGeneration`, optional `notes`.

The set was tuned by simulating an agent that gets only this file and has to produce a running project:

- `scaffoldCommand` says how to obtain the template (the tiged one-liner). Without it, an agent has to reverse-engineer the repo URL from `issueUrl`. Null exactly when `scaffoldable` is false, and the check enforces the pairing.
- `devCommand`, not `startCommand`. It mirrors the `dev` script, and "start" means the production serve on every deploy platform. An agent following the old name would build, discard the build, and ship a dev server.
- `env`, not `requiredEnv`. Every template boots against the public demo backend with zero configuration; the vars only re-point it at your own instance. "required" makes an agent stop and ask for credentials before first run. A root-level `notes` states the scaffold steps, this default, and that the vars belong in a `.env` file (vue-starter-template ships a `.env.template`).
- `env` is an object with a per-var purpose, not a bare name list, and the starter templates also carry the two optional vars their `.env.template` documents - `NUXT_PUBLIC_SHOPWARE_DEV_STOREFRONT_URL` (customer registration in local dev fails without it) and `NUXT_SHOPWARE_ENDPOINT`.
- `displayName`, because three of the five #2648 consumers need one and cannot derive it from `id` ("Blank Template", "Demo Store", "Vue Starter Template Extended (Lumora)").
- `devPort`, for the devcontainers #2648 generates. The two existing ones hardcode 3000, which only fits Nuxt; astro is 4321, vite is 5173.
- `devcontainer` mirrors `.devcontainer/<id>/` on disk, so a Codespaces link can be generated safely.
- `node` mirrors `engines.node`, because an agent should learn about an incompatible runtime before installing, not from the crash. The old `^20.x || ^22.x || ^24.x` ranges were stale twice over: Node 20 is EOL since April 2026, and a clean-box scaffold on Node 20/22.0 installs with `EBADENGINE` warnings and then dies on a missing rolldown binary. The real floor comes from the dependency tree - nuxt 4.5 needs `^22.19.0 || ^24.11.0 || >=26.0.0` - so engines now say that, the manifest mirrors it, and the check additionally verifies each template's range against its framework's own `engines.node`, so a faithful mirror of a lying value fails CI too.
- `devCommand`/`buildCommand` are the contents of the `dev` and `build` scripts. The descriptions now say explicitly: run the scripts through a package manager, not these strings, which fail before install. The contents stay in the manifest (rather than just the script names) so that editing a template's build script forces a manifest touch and CI review.
- `typeGeneration` says how to regenerate the Store API types against your own instance so they include its extensions: the env vars, the `shopware-api-gen loadSchema` fetch, and the package.json script. The fetch step matters: `generate-types` without a fetched schema silently falls back to the published default types, which is exactly the kind of trap the manifest exists to name. Null for the three templates that do not wire `@shopware/api-gen` (astro, vue-blank, vue-vite-blank).
- each `typeGeneration.env` var carries a description saying how to derive its value from the storefront configuration (`OPENAPI_JSON_URL` is the endpoint without its `/store-api` suffix, `OPENAPI_ACCESS_KEY` is the same value as the access token), so an agent does not ask the user for values it already has. Verified by simulation: an agent reading only the manifest, given a storefront URL and token, derived both values and produced the full scaffold-to-typed-dev-server workflow with zero extra questions.

`docsUrl` is required but nullable. Optional would let a new template quietly omit it; required-and-nullable forces a deliberate "no page yet". That surfaced a real gap: `vue-vite-blank` has no documentation page anywhere. It is now a visible `null`.

`packageName` is there because two ids do not match their package name: `astro` is `shopware-astro` and `vue-starter-template-extended` is `lumora-demo-store`. Any consumer filtering pnpm workspaces needs both.

### What the check verifies

- disk and manifest agree in both directions, no duplicate ids
- `packageName`, `buildCommand`, `devCommand` and `node` match each template's `package.json`. A template directory without a `package.json` is itself reported, not skipped.
- `docsUrl` resolves to a real file under `apps/docs/src/`, so it cannot point at a 404
- `issueUrl`'s decoded title param starts with `[id]` (percent-encoded or literal brackets both work). Prefix, not substring: `vue-starter-template` is a substring of `vue-starter-template-extended`, so the likeliest copy-paste error used to pass.
- `scaffoldable` and `scaffoldCommand` agree, and the command copies the template's own directory
- the `devcontainer` flag against `.devcontainer/` in both directions (only directories holding a `devcontainer.json` count, so shared config folders stay out of it)
- each template's `node` range is a subset of its framework's `engines.node`, read from the installed framework package; a missing install is reported, not silently skipped
- `typeGeneration` against `package.json` in both directions: a named script must exist, and null is a lie when a `generate-types` script does
- the manifest against the schema, with ajv in strict mode. `additionalProperties: false` catches a typo'd field; strict mode catches a typo'd keyword in the schema itself. Malformed JSON is reported with the file name instead of a stack trace.

Verified by breaking it on purpose, twenty-eight ways: the original ten, plus the prefix issueUrl, a scaffoldCommand pointing at the wrong directory, a null scaffoldCommand on a scaffoldable template, a devcontainer flag that lies, a typo'd field under strict mode, malformed JSON, a template directory missing its package.json, a node range that disagrees with engines.node, a missing manifestVersion, a wrong manifestVersion, a typeGeneration script that does not exist, a null typeGeneration next to a generate-types script, a typo'd nested typeGeneration field, a typo'd key inside a typeGeneration env var's metadata, a lowercase env var name, an empty env var description, a node range that admits versions the framework rejects, and a hidden framework package (the skip is reported, not silent). A literal-bracket issue title and a shared `.devcontainer/` config folder were verified to pass.

What the check deliberately does not verify: `supportLevel`, `deployable`, `devPort` and the `env` contents are human judgements with no authority on disk. The check confirms they are well formed, never that they are right. And the docsUrl check proves the page exists, not that it is the right page for that template.

### Two corrections to the issue

The issue's table lists `templates/demo-from-figma`. That directory does not exist on disk, so it is not in the manifest.

The issue says `vue-demo-store` is scheduled for deletion. Per #2667 it stays as a reference implementation. Copying it out of the monorepo works mechanically, but new projects must not start from it, so `scaffoldable` is `false` and its `notes` say why.

### ToDo's

- [ ] Related Issue updated

No changeset. Nothing here is a published package.

### Additional context

There is no root vitest config, so a vitest spec would have forced a lockfile change against `--frozen-lockfile` CI. The check is an `esno` script instead, matching the existing scripts. It validates through the schema with `ajv` (already in the tree transitively), so the schema is the definition rather than a copy of one. `semver` was added as a root devDependency for the framework-floor comparison - range subset math is not something to hand-roll.


templates/vue-starter-template/package.json
```diff
@@ -63,5 +63,8 @@
   "overrides": {
     "glob": "^13",
     "eslint-visitor-keys": "^4"
+  },
+  "engines": {
+    "node": "^20.x || ^22.x || ^24.x"
```
patzick: node 20 is EOL soon, we should not mention it

templates/manifest.json
```diff
@@ -0,0 +1,174 @@
+{
```
patzick, line 1: **Blocker** — this file and `manifest.schema.json` break the nightly StackBlitz run.

`apps/e2e-tests/tests/checkStackblitzTemplates.spec.ts:19` does `for (const template of fs.readdirSync(directoryPath))` — every *entry*, not every directory. After merge it generates two extra tests, `manifest.json` and `manifest.schema.json`, which navigate to `stackblitz.com/github/shopware/frontends/tree/main/templates/manifest.json` and fail on `waitForRequest(.../store-api/context)`.

It does not show up here because `check-stackblitz-templates.yml` is `schedule`/`workflow_dispatch` only, so the failure lands on the nightly cron after merge.

The PR description says the spec "opens a test for every directory" — that is the assumption the spec fails to enforce. Smallest fix in this PR:

```ts
for (const entry of fs.readdirSync(directoryPath, { withFileTypes: true })) {
  if (!entry.isDirectory()) continue;
  const template = entry.name;
```

Or, since the manifest now exists, drive it from `manifest.templates.filter((t) => t.scaffoldable)` — which also drops `vue-starter-template-extended`, the case you flagged as unable to pass. Either way it should not wait for #2648.

templates/manifest.json
```diff
@@ -0,0 +1,174 @@
...
+      "framework": "nuxt",
+      "purpose": "Supported starting point for a new storefront.",
+      "docsUrl": "https://developer.shopware.com/frontends/introduction/templates/vue-starter-template.html",
+      "issueUrl": "https://github.com/shopware/frontends/issues/new?labels=template&title=%5Bvue-starter-template%5D+",
+      "supportLevel": "supported",
+      "scaffoldable": true,
+      "scaffoldCommand": "npx tiged shopware/frontends/templates/vue-starter-template my-project",
+      "deployable": true,
+      "devcontainer": true,
+      "env": [
```
patzick: **Minor** — `env` is missing two vars that `templates/vue-starter-template/.env.template` documents: `NUXT_PUBLIC_SHOPWARE_DEV_STOREFRONT_URL` and `NUXT_SHOPWARE_ENDPOINT`.

The first one matters for the stated audience: its own comment says it is "required for customer registration to work in local development". An agent that treats this array as complete will hit that failure and have no way to know the var exists.

Since the check deliberately does not verify `env` contents, nothing will catch the omission later.

scripts/checkTemplatesManifest.ts
```diff
@@ -0,0 +1,289 @@
...
+    );
+  }
+}
+
+if (fs.existsSync(devcontainersDir)) {
+  for (const entry of fs.readdirSync(devcontainersDir, {
+    withFileTypes: true,
+  })) {
+    if (!entry.isDirectory()) continue;
+    if (!inManifest.includes(entry.name)) {
```
patzick: **Minor** — this asserts that *every* directory under `.devcontainer/` is a template id. Both current entries are, but a shared base config or a `library-scripts/` folder (both common in devcontainer setups) would fail the templates-manifest check with a message that reads like manifest drift.

Consider only flagging directories that actually contain a `devcontainer.json`, mirroring the forward check on line 263.

scripts/checkTemplatesManifest.ts
```diff
@@ -0,0 +1,289 @@
...
+  }
+}
+
+// Six near-identical URLs are easy to copy and forget to edit, and a wrong one
+// files the report against the wrong template. Match the full bracketed title
+// prefix: a bare substring match would let an id that is a prefix of another id
+// (vue-starter-template inside vue-starter-template-extended) slip through.
+for (const template of manifest.templates) {
+  const titlePrefix = `%5B${encodeURIComponent(template.id)}%5D`;
+  if (!template.issueUrl.includes(titlePrefix)) {
```
patzick: **Nit** — the prefix is matched only in percent-encoded form (`%5B…%5D`). `?title=[astro] ` with literal brackets is a valid, working GitHub URL, so a correct entry written that way fails with "title is not prefixed with [astro]" while visibly being prefixed with `[astro]`.

Decoding the `title` param before comparing (`new URL(...).searchParams.get("title")`) would accept both and simplify the check.

patzick (changes requested):
Solid piece of work — the manifest is well-shaped and the check is genuinely thorough (bidirectional disk/devcontainer checks, prefix-not-substring issueUrl matching, the traversal guard).

One blocker: adding two files into `templates/` breaks the nightly StackBlitz e2e run, which enumerates that directory without filtering. PR CI is green because that suite is cron-only. Details inline, plus three minor notes.

templates/manifest.json
mkucmus: ok, it's been changed. the spec now reads the manifest and opens only scaffoldable templates

apps/e2e-tests/tests/checkStackblitzTemplates.spec.ts
```diff
@@ -14,12 +14,17 @@ declare global {
...
+// Only scaffoldable templates can boot outside the monorepo, so only they can
+// pass on StackBlitz.
+const manifestPath = path.join(__dirname, "../../../templates/manifest.json");
+const { templates } = JSON.parse(fs.readFileSync(manifestPath, "utf8")) as {
+  templates: Array<{ id: string; scaffoldable: boolean }>;
+};
 
-for (const template of fs.readdirSync(directoryPath)) {
-  test(`Open, { tag: "@stackblitz" },  ${template}`, async ({ page }) => {
+for (const template of templates.filter((t) => t.scaffoldable)) {
```
patzick, line 24: Correct filter, but worth naming what it removes: `vue-starter-template-extended` is `scaffoldable: false`, so it drops out of the nightly run — and `apps/docs/.vitepress/theme/components/TryItOutPicker.vue` still renders a live **Open in StackBlitz** button for it (`buildOpenUrl: (t) => .../templates/${t}` with no per-template exclusion).

Before this PR that combination produced a permanently-red nightly test. Noise, but a signal. After it, the button stays broken and nothing reports it. The manifest now states the reason (`workspace:*` dep on `vue-starter-template`), so the picker should hide StackBlitz for non-scaffoldable templates — presumably #2648. Worth linking it there so it does not get lost.

Separately, `vue-demo-store` also drops out. That one is fine: it has no `workspace:*` deps and no StackBlitz link in the docs, so nothing user-facing is left uncovered. Just noting `scaffoldable` is carrying two meanings here — policy ("must not start new projects on it") and mechanics ("boots outside the monorepo") — and only the second is what this filter actually wants.

scripts/checkTemplatesManifest.ts
```diff
@@ -0,0 +1,308 @@
...
+    templateDir,
+    "node_modules",
+    template.framework,
+    "package.json",
+  );
+  if (!semver.validRange(template.node)) {
+    problems.push(
+      `"${template.id}" has node "${template.node}", which is not a valid semver range`,
+    );
+  } else if (fs.existsSync(frameworkPkgPath)) {
```
patzick, line 199: This is the one check that silently no-ops. If `templates/<id>/node_modules/<framework>/` is absent the subset assertion is skipped with no output, and the script still prints its success line.

CI always installs first so it holds there, but locally `pnpm run check:templates` on a fresh clone reports "matches templates/ and its schema" while having verified nothing about `node`. Given the point of the check is that a green result means something, the skip is worth reporting:

```ts
} else if (!fs.existsSync(frameworkPkgPath)) {
  problems.push(
    `"${template.id}" node range could not be checked against ${template.framework}: run pnpm install first`,
  );
}
```

Or at minimum a `console.warn`.

package.json
```diff
@@ -41,17 +42,19 @@
...
     "esno": "4.8.0",
     "husky": "9.1.7",
     "oxfmt": "0.62.0",
     "oxlint": "1.77.0",
+    "semver": "^7.8.5",
     "turbo": "2.10.10"
   },
   "engines": {
-    "node": "^20.x || ^22.x || ^24.x"
+    "node": "^22.19.0 || ^24.11.0 || >=26.0.0"
```
patzick, line 57: Heads-up rather than an objection: this drops Node 20 for the whole monorepo, not just templates, in a PR titled "add a manifest and a drift check".

The floor is accurate — `npm view nuxt@4.5.2 engines` returns exactly `^22.19.0 || ^24.11.0 || >=26.0.0` — and the per-template ranges check out too (`vue-vite-blank` `>=22.12.0` ⊂ vite's `^20.19.0 || >=22.12.0`, astro exact). No `.npmrc` here, so `engine-strict` is off and Node 20 contributors get a warning rather than a failed install.

One leftover: `apps/docs/package.json` still declares `"node": ">=22"`, which admits 22.0–22.18, below the new floor. Nothing checks app engines, so it will not fail anything — just inconsistent now.

patzick (approved):
All four points addressed, and the StackBlitz fix went further than suggested — driving the spec off the manifest makes it the first real consumer, which is the best possible proof the manifest is usable. CI green on 12/12.

- Blocker (StackBlitz spec enumerating files) — fixed, manifest-driven
- `env` gaps — fixed, and upgraded to keyed descriptions with schema backing
- `.devcontainer/` reverse check — fixed, now gated on `devcontainer.json`
- `issueUrl` encoded-only match — fixed, now decodes the `title` param

Approving. Three non-blocking notes inline; the second one is the only one I'd act on before merge.

Issue #2585, opened 2026-07-22
Bug: starter template registration uses wrong storefrontUrl and hides API errors


patzick:
## Description

The `vue-starter-template` registration flow can fail without giving the user visible feedback after they click submit.

In the deployed starter template at `https://frontends-starter-template.vercel.app/`, the `/account/register` payload sends `storefrontUrl: "https://frontends-demo.vercel.app"`. That value belongs to the demo store, not the starter template deployment. If Shopware rejects the registration request, the form remains on screen and the user does not get useful inline feedback explaining what happened.

## Affected Template

- `templates/vue-starter-template`
- Live template: https://frontends-starter-template.vercel.app/

## Steps to Reproduce

1. Open the starter template registration form, for example `https://frontends-starter-template.vercel.app/account/login#registration`.
2. Fill in the registration form.
3. Click submit.
4. Inspect the failed `/account/register` request in the browser devtools.

<img width="3372" height="1606" alt="Image" src="https://github.com/user-attachments/assets/e1813546-e699-4749-9a41-6b3416903c3e" />

## Actual Result

- The request payload contains `storefrontUrl: "https://frontends-demo.vercel.app"`.
- The request fails.
- The user remains on the registration form without clear visible feedback after clicking submit.

## Expected Result

- The starter template should use the correct storefront URL for registration. In production it should not be forced to the demo storefront URL.
- Registration API errors should be displayed to the user in the registration form, with a fallback message if the API error payload is empty or cannot be resolved.

## Suggested Fix

- Do not default `devStorefrontUrl` in `templates/vue-starter-template` to `https://frontends-demo.vercel.app`; leave it empty so production uses the current browser origin.
- Apply the same default to the starter template Docker build/runtime args.
- Capture registration API errors in `RegistrationForm.vue` and render them in an accessible inline alert.
- Ensure the shared API error resolver emits the default translated error message when Shopware returns an empty error list.

PR #2677, opened 2026-08-31
feat(e2e): run the suite against vue-starter-template, and fix what it exposed

### LINKED ISSUE #2649: [CF][Templates] Run the e2e suite in CI against vue-starter-template

### LINKED ISSUE #2651: [CF][Templates] Rebuild login and register page objects for the starter

mkucmus:
### Description

Adds a workflow that builds `vue-starter-template`, serves it with `nuxt preview`, and runs `test:e2e` against it. Runs nightly and on manual dispatch.

Retags the suite from `@vue-demo-store` to `@frontends`, so the tag names the behaviour rather than one template.

Along the way it fixes three defects the suite exposed. Two are in `@shopware/cms-base-layer`, a published package, despite the commit subjects reading as test work:

- Listing filters wrote the URL only after the listing request resolved, so a slow or failed request left the filter out of the address bar and lost it on refresh or share. Reordered to push first, matching `changePage`, `changeLimit` and sorting. Applied to both the sidebar and horizontal filter components.
- Checkout always sent `guest: true`, so a customer who chose to create an account got a guest account and could never sign in.

closes #2649
closes #2651

### Type of change

Bugfix and new feature (non-breaking)

### ToDo's

- [x] E2E-Tests added/updated
- [x] Changeset file provided - patch for `@shopware/cms-base-layer`

### Result

21 passed, 3 flaky, 2 skipped, 0 failed on CI in 7.4m.

It does not run on pull requests. The path filter it started with matched 19 of the last 25 merged PRs, at a median of 12 minutes a run, which is the same trade #2706 made for the example builds.

Because nothing is gated on it any more, a failure now notifies Slack through `failed-job-check.yml`, and the test step no longer uses `continue-on-error` (with it, the run concludes `success` and the notifier never fires).

Also fixes the branch link in that Slack message, which pointed at `github.com/<repo>/<branch>` and 404'd. It affects all seven workflows listed in `failed-job-check.yml`, not just this one.

Every flaky scenario traces to one cause, now filed as #2682: Store API calls that fail without a response. Nitro logs none of it, so this branch adds a collector (`apps/e2e-tests/fixtures.ts`) that records every failed call, how long it ran before dying, and any error page the browser was shown. A summary lands in the job page and the raw JSONL in the artifact.

That instrumentation corrected two assumptions worth stating. Requests are not hanging: on CI they die at a median of 171ms and none exceeded 8s, so a request timeout would never fire. And about a quarter of what looked like backend failures are `net::ERR_ABORTED`, which is the client cancelling in flight, mostly the country combobox firing per keystroke.

The classifier itself had two bugs, fixed in `13a1d4ea`: it logged a success even on a 4xx or 5xx, and counted 429s, 5xx and error pages as no-response failures. They pulled the rate in opposite directions and the corrected figure is 4.8% rather than 9.5%. The latency numbers above never went through that path and are unaffected. Every CI run on this branch predates the fix, so treat its reported rate as approximate until the first nightly run.

Built and served locally rather than pointed at a Vercel preview, so a run tests the branch it runs on. Sets `NUXT_PUBLIC_SHOPWARE_DEV_STOREFRONT_URL` because the template default is the demo store domain, which the starter's sales channel rejects (#2585).

### Review follow-up

All four findings addressed.

1. **Account creation.** `createAccount` is now a model on `CustomerBaseInfo`, bound from the checkout page. Guest validates email and billing address, account validates the whole base-info form, and `guest` is derived rather than hardcoded. One spec per mode; the account one signs in over the Store API, which is what proves the customer is not a guest, since a guest cannot sign in at all.
2. **Path filter.** All seven `@shopware/*` dependencies are symlinked into `packages/`, so a change in one rebuilds the starter without touching the template. Added the six runtime packages and root dependency metadata. `api-gen` is left out, it is codegen only.
3. **Listing synchronisation.** The wait is armed before the click and resolves only on a successful listing response whose body carries the filter, then waits for the render. Bounded at 30s with specific failure messages. Applied to manufacturer and property filters on both listings, and to sorting.
4. **Variant identity.** `data-product-id` on the add-to-cart button and the cart line item; the test asserts exactly one line item carries the id the button submitted, rather than checking an image is visible.

A second round of nine findings is answered in [this comment](https://github.com/shopware/frontends/pull/2677#issuecomment-5601527330). All are fixed on the branch, including one real regression it caught in `SwProductListingFiltersHorizontal.vue`, where property-group filters stopped rendering as active.

### What this closes

- #2649, the workflow itself.
- #2651, login and register page objects rebuilt against the starter.

### What is still queued, and why

- #2650 is not auto-closed. 31 `data-testid` attributes were added across 27 files. Its `comm -23` check still lists 22 ids, but the check undercounts: it greps only the template directory and only the literal `data-testid=`, so it misses 10 ids that live in `cms-base-layer`, one prop default and one dynamic binding. Corrected, the residue is 11, of which only three or four are attributes on components that already exist; the rest are blocked on #2679, #2680 and #2652. Measurement and full breakdown in [this comment](https://github.com/shopware/frontends/issues/2650#issuecomment-5614923392).
- #2682, Store API calls that hang with no timeout or retry. The single cause of the flaky count, and the blocker on this job ever becoming a gate.
- #2683, checkout derives guest or account mode from a 600ms animation timer rather than from the click. Low severity, nobody fills a checkout form that fast, but presentation timing should not own form semantics. Found while adding the coverage above.
- #2690, `apiClientConfig` is a documented module option that nothing reads, so a consumer following the published types gets silence. A timeout was tried here and reverted: measured on CI it never fired once.
- #2691, `useListing` ignores `cacheableReads`, so category listings and search stay on POST. They lose CDN caching and, less obviously, ofetch's automatic retry, which only applies to GET.
- #2692, `cacheableReads` is applied by hand at each call site with nothing enforcing completeness. #2691 is the symptom.
- #2679 and #2680, two scenarios that stay skipped because the starter has no clear-wishlist action and renders no product reviews.
- #2585, registration `storefrontUrl`. The workflow works around it with an env var, the bug remains.

### Worth a reviewer's attention

Nothing in CI runs e2e against `vue-demo-store`, and retagging removed the last demo-store-specific suite. Three published-package files change here (`SwProductListingFilters.vue`, `SwProductListingFiltersHorizontal.vue`, `SwProductAddToCart.vue`) with no automated coverage of the app most affected by them. The reordering matches the convention already used elsewhere in the same package, and no template overrides those components, but that is analysis rather than a test.

Also fixes a `--` in the e2e README that silently drops arguments under pnpm 11.


templates/vue-starter-template/app/components/layout/MiniCart.vue
```diff
@@ -30,7 +30,11 @@ onClickOutside(miniCartContainer, () => {
 });
 </script>
 <template>
-  <div ref="miniCartContainer" class="z-20 max-w-[500px] w-full">
+  <div
+    ref="miniCartContainer"
+    data-testid="sidebar-right"
```
mdanilowicz: `testid` should not be based on the appearance or position.  You can sue `mini-cart-container`

templates/vue-starter-template/app/components/layout/MiniCart.vue
```diff
@@ -39,7 +43,11 @@ onClickOutside(miniCartContainer, () => {
       >
         {{ $t("cart.miniCart.title") }}
       </div>
-      <FormIconButton type="ghost" @click.stop="handleCloseMiniCart">
+      <FormIconButton
+        type="ghost"
+        data-testid="cart-close-button"
```
mdanilowicz:
```suggestion
        data-testid="mini-cart-close-button"
```

templates/vue-starter-template/app/components/layout/MiniCart.vue
```diff
@@ -68,6 +76,7 @@ onClickOutside(miniCartContainer, () => {
         {{ $t("cart.miniCart.taxEstimation") }}
       </div>
       <NuxtLink
+        data-testid="cart-checkout-link"
```
mdanilowicz:
```suggestion
        data-testid="checkout-cart-link"
```

Because checkout (in this case, the checkout process) refers to the page where the cart is processed, it’s a more semantically meaningful name.

templates/vue-starter-template/app/components/layout/Header.vue
```diff
@@ -72,6 +72,11 @@ watch(
             <div class="relative">
               <FormIconButton
                 type="ghost"
+                :data-testid="
```
mdanilowicz: This is an anti-pattern. A testId should identify an element, not represent its state

templates/vue-starter-template/app/components/layout/Header.vue
```diff
@@ -96,6 +101,7 @@ watch(
             </div>
             <ClientOnly>
               <NuxtLink
+                data-testid="wishlist-button"
```
mdanilowicz:
```suggestion
                data-testid="header-wishlist-button"
```

templates/vue-starter-template/app/components/layout/Header.vue
```diff
@@ -108,6 +114,7 @@ watch(
 
             <FormIconButton
               type="ghost"
+              data-testid="cart-button"
```
mdanilowicz:
```suggestion
              data-testid="header-mini-cart-button"
```

templates/vue-starter-template/app/pages/wishlist.vue
```diff
@@ -70,6 +70,7 @@ const fetchWishlistProducts: OffsetPaginationFetcher<
 
         <template #empty>
           <div
+            data-testid="wishlist-empty"
```
mdanilowicz:
```suggestion
            data-testid="wishlist-empty-container"
```

apps/e2e-tests/page-objects/RegisterPage.ts
```diff
@@ -45,44 +48,60 @@ export class RegisterForm {
...
+    await this.firstName.fill(`e2e ${faker.person.firstName()}`);
+    await this.lastName.fill(`e2e ${faker.person.lastName()}`);
+    await this.emailAdrdress.fill(uniqueEmail());
+    await this.password.fill(faker.internet.password());
+    await this.street.fill(faker.location.street());
+    await this.zipcode.fill(faker.location.zipCode());
+    await this.city.fill(faker.location.city());
+    await selectCountry(this.page, this.country, "Germany");
+    await selectFirstOptionIfPresent(this.countryState);
+    await this.submitRegistraionForm();
```
mdanilowicz:
```suggestion
    await this.submitRegistrationForm();
```

Please fix the function name with that change

apps/e2e-tests/page-objects/MyAccountPage.ts
mdanilowicz:
```suggestion
  async subscribeNewsletter() {
```

I know this isn’t from you, but I think it’s worth fixing :D

templates/vue-starter-template/app/components/layout/header/Search.vue
```diff
@@ -88,6 +88,7 @@ const handleEnterKey = () => {
     <div @click="suggestIsActive = true">
       <FormBaseInput
         id="search-input"
+        data-testid="layout-search-input"
```
mdanilowicz:
```suggestion
        data-testid="header-search-input"
```

templates/vue-starter-template/app/components/account/LoginForm.vue
```diff
@@ -42,6 +42,7 @@ function handleSignUp() {
     <LoginForm @submit="handleLogin" />
     <FormBaseButton
       v-if="!hideSignUp"
+      data-testid="login-sign-up-link"
```
mdanilowicz:
```suggestion
      data-testid="login-sign-up-button"
```

templates/vue-starter-template/app/components/wishlist/ProductTile.vue
```diff
@@ -63,10 +63,12 @@ async function handleRemoveFromWishlist() {
 </script>
 <template>
   <div
+    data-testid="product-box"
```
mdanilowicz:
```suggestion
    data-testid="wishlist-product-box"
```

templates/vue-starter-template/app/components/wishlist/ProductTile.vue
```diff
@@ -63,10 +63,12 @@ async function handleRemoveFromWishlist() {
...
     class="flex flex-col gap-4 relative"
     :style="`width: ${ELEMENT_WIDTH}px;`"
   >
     <ProductWishlistIcon
+      data-testid="product-box-toggle-wishlist-button"
```
mdanilowicz:
```suggestion
      data-testid="wishlist-product-box-toggle-button"
```

templates/vue-starter-template/app/components/wishlist/ProductTile.vue
```diff
@@ -98,6 +100,7 @@ async function handleRemoveFromWishlist() {
 
       <FormBaseButton
         class="mt-auto"
+        data-testid="add-to-cart-button"
```
mdanilowicz:
```suggestion
        data-testid="wishlist-add-to-cart-button"
```

templates/vue-starter-template/app/components/checkout/ProductTile.vue
```diff
@@ -48,6 +53,7 @@ const cover = computed(() =>
         <FormLinkButton
           class="text-sm border-b-1 border-b-solid border-b-brand-primary hover:border-none"
           label="Remove"
+          data-testid="product-remove-button"
```
mdanilowicz:
```suggestion
          data-testid="checkout-product-tile-remove-button"
```

templates/vue-starter-template/app/components/checkout/ProductTile.vue
```diff
@@ -19,9 +19,14 @@ const cover = computed(() =>
...
 <template>
-  <div class="flex gap-4">
+  <div
+    class="flex gap-4"
+    data-testid="cart-line-item"
+    :data-product-id="item.referencedId"
+  >
     <div class="w-37.5 h-37.5">
       <NuxtImg
+        data-testid="cart-product-image"
```
mdanilowicz:
```suggestion
        data-testid="checkout-product-tile-image"
```

templates/vue-starter-template/app/components/checkout/ProductTile.vue
```diff
@@ -19,9 +19,14 @@ const cover = computed(() =>
 );
 </script>
 <template>
-  <div class="flex gap-4">
+  <div
+    class="flex gap-4"
+    data-testid="cart-line-item"
```
mdanilowicz:
```suggestion
    data-testid="checkout-product-tile-item"
```

templates/vue-starter-template/app/components/checkout/CustomerBaseInfo.vue
```diff
@@ -18,15 +18,19 @@ const { errorMessages } = defineProps<{
...
 const switchAnimating = ref(false);
-const createAccountToggle = ref(false);
 
 function switchAnimation(e: Event) {
   e.preventDefault();
 
   switchAnimating.value = true;
   setTimeout(() => {
-    createAccountToggle.value = true;
+    createAccount.value = true;
```
mdanilowicz: ⚠️ Looks like a bigger problem. The animation is overriding the model data for account creation availability

packages/cms-base-layer/app/components/SwProductListingFilters.vue
```diff
@@ -180,9 +173,18 @@ const executeSearch = async () => {
     // otherwise drop `limit` and revert to the default on refresh/navigation.
     if (route.query.limit) query.limit = route.query.limit;
 
+    // URL first, the way changePage/changeLimit/sorting already do it. Gating
+    // it on the request left the selection out of the URL whenever the listing
+    // call was slow or failed, so a refresh or a shared link lost the filter.
     await router.push({
       query: query as LocationQueryRaw,
     });
+
```
mdanilowicz: [SwProductListingFiltersHorizontal.vue](https://github.com/shopware/frontends/pull/2677/changes#diff-3fae55eb406cedbe5cb7b31045c8849a0e04bf901e77ce660d55dd3e05d333e2) has the same changes. It should be moved to helper or composable (not only this part, shared code)

apps/e2e-tests/playwright.config.ts
```diff
@@ -23,13 +23,15 @@ const config: PlaywrightTestConfig = {
...
+  timeout: 60 * 1000,
   expect: {
     /**
      * Maximum time expect() should wait for the condition to be met.
      * For example in `await expect(locator).toHaveText();`
      */
-    timeout: process.env.CI ? 30000 : 5000,
+    // Well under the test timeout, so a failing assertion reports its own
+    // diff instead of the test dying first.
+    timeout: 10000,
```
mdanilowicz, line 34: Why was the variable removed?

apps/e2e-tests/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "1.0.0",
   "description": "",
   "scripts": {
-    "test:e2e": "playwright test --grep @vue-demo-store",
+    "test:e2e": "playwright test --grep @frontends",
```
mdanilowicz, line 7: Why frontends not vue-starter-template?

packages/cms-base-layer/app/components/SwProductListingFilters.vue
```diff
@@ -180,9 +173,18 @@ const executeSearch = async () => {
...
+    // Search context refetches from the URL (the page's useAsyncData watches
+    // route.query). Category listings have no such watcher, so fetch here.
+    if (!isProductSearch) {
+      await search(criteria);
+    }
```
mdanilowicz: Browser navigation is not updating the product list

.github/workflows/e2e-starter-template.yml
```diff
@@ -0,0 +1,154 @@
...
+
+      - name: Run tests
+        id: e2e
+        # Reports rather than gates. The suite passes against the starter, but
+        # the shared demo backend intermittently stalls store-api calls
+        # (context, cart, breadcrumb, product-listing) with no response at all,
+        # which strands the page mid-hydration and fails whichever spec was
+        # running. Retries absorb it. Kept out of failed-job-check.yml for the
+        # same reason.
+        continue-on-error: true
```
mdanilowicz, line 98: You’ll need to update `failed-job-check.yml` for this to make sense.

apps/e2e-tests/utils/listing.ts
```diff
@@ -0,0 +1,182 @@
...
+    .first();
+  const before = await firstProduct.getAttribute("href");
+
+  await page.getByRole("button", { name: `Page ${number}` }).click();
+  await page.waitForURL(new RegExp(`p=${number}`));
+  await listingRendered(page);
+  await expect(firstProduct).not.toHaveAttribute("href", before ?? "", {
+    timeout: LISTING_TIMEOUT,
+  });
+}
```
mdanilowicz: This should be a part of the POM, especially since these are UI interactions

apps/e2e-tests/page-objects/LoginPage.ts
```diff
@@ -17,10 +17,9 @@ export class LoginForm {
 
   // Define login page methods
   async login(username: string, password: string) {
-    await this.usernameInput.isVisible();
-    await this.usernameInput.type(username);
-    await this.passwordInput.type(password);
-    await this.page.waitForLoadState();
+    await this.usernameInput.waitFor({ state: "visible" });
```
mdanilowicz: We don't need this line. `fill` will wait until component is ready

mdanilowicz (changes requested):

apps/e2e-tests/playwright.config.ts
mkucmus: Good catch.

It was `process.env.CI ? 30000 : 5000`. The 30s equalled the test timeout, so the test died before the assertion could report its diff. Both CI and local hit the same backend anyway, so the split had no basis.

apps/e2e-tests/package.json
mkucmus: The tag names the behaviour, not the template. These scenarios should run against any Frontends template, and tying the tag to one means retagging every time another lands, which is what happened with `@vue-demo-store`.

.github/workflows/e2e-starter-template.yml
mkucmus: Fair point, and worth deciding explicitly. Two options:

- **Add it now** - the job blocks merges and pages on the nightly. With ~6 flaky per run it would be red often.
- **Leave it out** - the job reports, nobody gets paged, but a real break is easier to miss.

Going with the second for now. Every flaky run traces to the shared demo backend dropping Store API calls (#2682), which we do not control, and #2653 just disabled the accessibility nightly for exactly this. Happy to switch once #2682 is sorted.

packages/cms-base-layer/app/components/SwProductListingFiltersHorizontal.vue
```diff
@@ -252,10 +67,9 @@ const hasActiveFilter = (filter: { code: string }) => {
     return sidebarSelectedFilters.rating !== undefined;
   }
   if (filter.code === "shipping-free") {
-    return sidebarSelectedFilters["shipping-free"] === true;
+    return sidebarSelectedFilters["shipping-free"] !== undefined;
   }
-  // Properties filter - check if any property from this filter group is selected
-  return sidebarSelectedFilters.properties.size > 0;
+  return false;
```
patzick: **Regression.** Property-group filters never render as active any more.

`SwProductListingFilter` resolves a property group via the fallback `"options" in filter ? SwFilterPropertiesVue : undefined`, so `filter.code` stays the group's own code (`Colour`, `Selection`, …) — only the *emitted* code is normalised to `"properties"` (`SwFilterProperties.vue:53-56`). There is no filter whose `code === "properties"`.

So the branch added at line 54 is unreachable, and replacing the old fallback (`return sidebarSelectedFilters.properties.size > 0`) with `return false` means `SwFilterDropdown` gets `:is-active="false"` for every property group, however many values are ticked.

The old code handled this precisely *because* it was a fallback. Suggest keeping it as one:

```ts
  if (filter.code === "shipping-free") {
    return sidebarSelectedFilters["shipping-free"] !== undefined;
  }
  // Any other code is a property group; each emits code "properties".
  return sidebarSelectedFilters.properties.size > 0;
```

(That is still coarse — every group lights up when any group has a selection — but it is the pre-existing behaviour, and narrowing it properly needs the group's own option ids.)

packages/cms-base-layer/app/components/public/cms/element/CmsElementProductListing.vue
```diff
@@ -58,22 +58,24 @@ const limit = ref(
...
+    // A different path mounts its own component.
+    if (route.path !== initialPath) return;
+
+    const hasQuery = Object.keys(query).length > 0;
+    const criteria = hasQuery
+      ? query
+      : { limit: defaultLimit, p: defaultPage, order: defaultOrder };
+
+    changeCurrentPage(
+      hasQuery && query.p ? Number(query.p) : defaultPage,
```
patzick: The watcher fetches with the URL's `limit`, but `limit` (line 53) is only initialised at setup and otherwise updated by the pagination's `v-model`. On Back/Forward the two diverge: the products come back for `?limit=1` while the "Per Page" select and `ProductCardSkeleton v-for="index in limit"` still show the previous value.

Since this changeset promises "Browser back and forward now update the products", syncing the ref in the same watcher would close it:

```ts
if (query.limit) limit.value = Number(query.limit);
```

packages/cms-base-layer/app/components/public/cms/element/CmsElementProductListing.vue
```diff
@@ -58,22 +58,24 @@ const limit = ref(
...
-      order: defaultOrder,
-    } as unknown as operations["searchPage post /search"]["body"]);
+  () => route.query,
+  (query) => {
+    // A different path mounts its own component.
+    if (route.path !== initialPath) return;
+
+    const hasQuery = Object.keys(query).length > 0;
+    const criteria = hasQuery
+      ? query
```
patzick: `criteria` is now `route.query` verbatim, and this watcher is the only fetch path for a category listing. Before, a filter change went through `executeSearch` → `search(searchCriteriaForRequest.value)` with typed values; now the same request carries `min-price: "19"`, `rating: "4"`, `shipping-free: "true"` as strings, plus a stray `p`.

It evidently works against the demo backend, so not blocking — but the `as unknown as` cast is hiding it, and `useListingFilters.searchCriteriaForRequest` (with its `post-filter` construction for search-page categories) is now returned by the composable and consumed by nobody, i.e. dead. Worth either normalising here or deleting the unreachable branch so the next reader doesn't assume the post-filter is still applied.

packages/cms-base-layer/app/components/public/cms/element/CmsElementProductListing.vue
```diff
@@ -86,10 +88,6 @@ const changePage = async (page: number) => {
       limit: limit.value,
     },
   });
-  await changeCurrentPage(
-    page,
-    route.query as unknown as operations["searchPage post /search"]["body"],
-  );
   productListElement.value?.scrollIntoView({ behavior: "smooth" });
```
patzick: This used to run after `await changeCurrentPage(...)`. Now `router.push` resolves, the watcher flips `loading` to `true`, and `v-if="!loading"` unmounts `productListElement` — so the smooth scroll is started on a node that is about to be detached. Same at line 102.

Awaiting the listing (or scrolling after `loading` goes false) would keep the behaviour the pagination had.

apps/e2e-tests/page-objects/ListingPage.ts
```diff
@@ -0,0 +1,153 @@
...
+  ) {
+    const checkboxes = this.page.locator('input[type="checkbox"]');
+    const before = await checkboxes.count();
+
+    await panel.click();
+    await checkboxes.nth(before).waitFor({ state: "attached" });
+
+    // Armed first: the request leaves as soon as the box is ticked.
+    const listed = listingRequested(this.page, filterKey);
+    await checkboxes.nth(before).check({ force: true });
```
patzick: Picking the option by page-wide index delta is fragile: it assumes the panel's checkboxes are appended after every existing one and that nothing else changes the count between the two reads. If the filter rail is an accordion that closes the previously open panel, `before` is measured against a larger set and `nth(before)` lands on an unrelated box.

Scoping to the panel's own container is stable regardless of DOM order:

```ts
const option = panel.locator('xpath=following::*[@role="group" or @id][1]')
  .locator('input[type="checkbox"]').first();
```

or, more simply, give the filter panel body a `data-testid` in `cms-base-layer` and locate within it.

apps/e2e-tests/page-objects/ListingPage.ts
```diff
@@ -0,0 +1,153 @@
+import { expect, type Locator, type Page } from "@playwright/test";
+
+import { LISTING_TIMEOUT, listingRequested } from "../utils/listing";
+
+/** Property groups are sales channel content, so one is found by elimination. */
+const NON_PROPERTY_FILTERS = [
```
patzick: This list mixes two vocabularies: `"Sort"` and `"Categories"` are rendered labels, while `"shipping-free"`, `"rating"`, `"price"` and `"manufacturer"` are API filter codes that no button renders as text. Against the starter the buttons happen to be labelled with the codes, so it passes — but on any template that localises them (`Price`, `Rating`, `Free shipping`) the elimination picks one of those as "the property filter", ticks a price radio, and the run fails on the `properties` assertion instead of reporting "no property filter here".

Given the whole point of the retag is template independence, matching on a `data-testid`/`filter.code` attribute rather than button text would hold up.

apps/e2e-tests/tests/checkForSeoUrl.spec.ts
```diff
@@ -20,15 +19,12 @@ test.describe("Check for seo-url requests", { tag: "@vue-demo-store" }, () => {
...
-    await page.waitForLoadState("networkidle");
     await expect(SeoUrlRequest).toBe(false);
-    await page
-      .getByRole("menuitem", { name: "Summer Trends", exact: true })
-      .click();
-    await page.waitForLoadState("networkidle");
+    // Any nav entry will do. Clicked rather than navigated to, because this
+    // test is about client-side routing not issuing seo-url requests.
+    await page.locator('[role="menubar"] [role="menuitem"]').first().click();
     await expect(SeoUrlRequest).toBe(false);
```
patzick, line 34: With the `waitForLoadState("networkidle")` calls gone, this assertion runs immediately after the click resolves — before the client router can issue anything — so it is now vacuous: it would stay `false` even if the navigation did fire a `seo-url` request.

The earlier two assertions are still covered by the implicit selector waits inside `visitMainPage`/`openCartPage`, but this one needs its own signal. Waiting for the destination listing to render (or for the category request) would restore it.

apps/e2e-tests/page-objects/WishlistPage.ts
```diff
@@ -9,16 +9,40 @@ export class WishlistPage {
...
+    await this.page
+      .locator(
+        '[data-testid="wishlist-product-box"], [data-testid="wishlist-empty-container"]',
+      )
+      .first()
+      .waitFor({ state: "visible" });
+  }
+
+  /** Via the header link: /wishlist is ssr:false and gated on being signed in. */
+  async countEntries() {
```
patzick: Unused — no caller anywhere in the suite. Worth dropping unless a follow-up needs it.

apps/e2e-tests/instrument-ssr.mjs
```diff
@@ -0,0 +1,54 @@
...
+        write({
+          kind: `store-api-${response.status === 429 ? "throttled" : response.status}`,
+          url,
+          status: response.status,
+          traceId: response.headers.get("x-trace-id"),
+          ms: Date.now() - startedAt,
+        });
+      } else {
+        // A late render looks like a broken one, so measure latency too.
+        write({ kind: "store-api-ok", url, ms: Date.now() - startedAt });
```
patzick, line 55: `appendFileSync` on the SSR request path for *every* successful Store API call adds blocking I/O to the thing being measured — which matters because the PR description leans on these numbers ("median 171ms", "none exceeded 8s") to conclude a request timeout would never fire. Buffering the entries and flushing on exit, or recording only failures plus a sampled success, keeps the measurement clean.

patzick (changes requested):
## Review — changes requested

CI is fully green (13/13, e2e included). One real regression in a published package, the rest are minor/nit.

**Blocking**
1. `SwProductListingFiltersHorizontal.vue` — the rewritten `hasActiveFilter` never reports property-group filters as active. Property filters carry the group's own code, not `"properties"`, so the new branch is dead and the changed fallback (`return false`) drops the behaviour the old fallback provided. Inline comment has the detail.

**Non-blocking** (inline)
2. `CmsElementProductListing.vue` — the local `limit` ref no longer resyncs from the URL, so Back/Forward updates the products but not the "Per Page" select or the skeleton count.
3. `CmsElementProductListing.vue` — the watcher is now the only fetch path and passes `route.query` verbatim, so `min-price`/`rating`/`shipping-free` reach the API as strings where they used to be typed; the `post-filter` branch in `useListingFilters` became unreachable as a result.
4. `changePage`/`changeLimit` scroll while the element they target is being unmounted for the skeleton.
5. `ListingPage.checkFirstOptionIn` picks a checkbox by page-wide index delta.
6. `ListingPage.NON_PROPERTY_FILTERS` mixes rendered labels with API codes.
7. `checkForSeoUrl.spec.ts` — the last assertion lost its wait and is now vacuous.
8. `WishlistPage.countEntries()` is unused.
9. `instrument-ssr.mjs` writes synchronously on every successful SSR call, inflating the latency it measures.

**Open thread, your call:** @mdanilowicz's `failed-job-check.yml` point is still unresolved. With `continue-on-error: true` and no gate anywhere, a total e2e failure leaves the job green and notifies nobody — the reporting-only trade-off is reasonable, but it should be an explicit team decision rather than an unresolved thread.

Also noting the gap you flagged yourself: three published `cms-base-layer` components change with no automated coverage of the app most affected. Finding 1 is exactly what that gap lets through.

.github/workflows/e2e-starter-template.yml
mkucmus: Changed my mind on this, because the reason not to is gone.

The job no longer runs on pull requests at all, only on the nightly schedule and manual dispatch. So there is nothing left to block, and the only cost of notifying is Slack noise on a flaky night.

Added `E2E Starter Template` to `failed-job-check.yml`. That also meant dropping `continue-on-error` from the test step, otherwise the run concludes `success` and the notifier would never fire.

.github/workflows/e2e-starter-template.yml
```diff
@@ -0,0 +1,123 @@
+name: E2E Starter Template
+
+run-name: Playwright e2e suite against vue-starter-template 🚀
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 4 * * *"
```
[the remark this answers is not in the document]
mkucmus: Deliberate, and the description was stale - now updated.

Moving it off pull requests is the change. The path filter matched 19 of the last 25 merged PRs at a median of 12 minutes, the same trade #2706 made for the example builds. #2649 is still met: the suite runs nightly and a failure notifies Slack.

.github/workflows/failed-job-check.yml
```diff
@@ -9,6 +9,7 @@ on:
         Stackblitz templates,
         Audit check,
         Accessibility Check,
+        E2E Starter Template,
```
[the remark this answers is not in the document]
mkucmus: The description was stale and is now updated. Notifying is intentional.

The job no longer runs on pull requests, so there is nothing left to gate, and without the notifier a nightly failure would go unnoticed.

mkucmus:
Thanks, this was a useful review. All nine are fixed and on the branch, plus the open question at the end.

**1 (blocking).** Right, and it was my regression. The fallback is back as you wrote it, comment included.

**2.** `limit.value` now resyncs from the query in the same watcher.

**3.** Added a `buildCriteria()` helper that converts the query strings to numbers and booleans, so the API gets typed values again. Deleted the unreachable `post-filter` branch, its import, and the `searchCriteriaForRequest` export. The computed stays as an internal for `executeSearch`; `search.vue` builds its own post-filter independently.

**4.** `scrollToListing()` awaits `until(loading).toBe(false)` before scrolling.

**5.** `checkFirstOptionIn` scopes to the panel's own container instead of a page-wide index delta. Went with your second suggestion and added `data-testid="listing-filter-{code}"` in `SwProductListingFilter.vue`.

**6.** Now matches on `filter.code` from that same testid, so the elimination list is codes only and survives localisation.

**7.** Added a wait for the destination listing to render.

**8.** Dropped.

**9.** Buffered, flushed on a 2s interval and on exit.

While fixing 9 I found two bugs in my own measurement pointing opposite ways: the attempts log counted 4xx and 5xx twice, and `transport` counted 429s, 5xx and error pages as no-response failures. The number moved from 9.5% to 4.8%. Every percentage in the description before this should be read as approximate.

**On the open thread.** Settled by removing the pull_request trigger entirely: the suite is nightly plus manual dispatch. `continue-on-error` is gone, so a failure fails the run, and `failed-job-check.yml` notifies Slack the same way the audit check does. Nothing is gated because nothing runs on PRs any more.

apps/e2e-tests/README.md
```diff
@@ -22,7 +22,28 @@ The login tests need an account on that storefront. Put its `USER_EMAIL` and `PA
 pnpm run test:e2e
 ```
 
-`test:e2e` is `playwright test --grep @vue-demo-store`, so it only runs the specs tagged for the deprecated demo store template. Retagging it for `vue-starter-template` is tracked in [#2649](https://github.com/shopware/frontends/issues/2649).
+`test:e2e` is `playwright test --grep @frontends` and takes its target from `BASE_E2E_URL`. The specs navigate generically, so they run against any storefront rather than one template.
+
+CI runs them against `vue-starter-template`: on pull requests that touch `templates/vue-starter-template/**`, `apps/e2e-tests/**` or the workflow itself, nightly, and on manual dispatch. See [.github/workflows/e2e-starter-template.yml](../../.github/workflows/e2e-starter-template.yml).
```
patzick: Stale as of the latest push: the `pull_request` trigger was removed, so this is now `workflow_dispatch` and nightly only. Line 29's "still reports rather than gates" is also no longer true — `continue-on-error` is gone and the job fails, notifying Slack via `failed-job-check.yml`.

Suggested:

> CI runs them against `vue-starter-template` nightly and on manual dispatch. See [.github/workflows/e2e-starter-template.yml](../../.github/workflows/e2e-starter-template.yml).
>
> The run passes against the starter. Nothing is gated on it — it is scheduled only, and a failure notifies Slack through `failed-job-check.yml`. Two scenarios stay skipped …

AGENTS.md
```diff
@@ -124,9 +124,12 @@ written there:
...
 - Unit tests (Vitest) live next to their sources as `*.test.ts`. `pnpm run
 coverage` exists only in `packages/composables`.
-- `pnpm run test:e2e` is `playwright test --grep @vue-demo-store`, so it only
-  covers the deprecated demo store. The `@accessibility` specs are
-  template-agnostic and run against any storefront via `BASE_E2E_URL`:
+- `pnpm run test:e2e` is `playwright test --grep @frontends` and takes its
+  target from `BASE_E2E_URL`. The specs navigate generically, so they run
+  against any storefront. CI runs them against `vue-starter-template` in
+  `.github/workflows/e2e-starter-template.yml`; that job reports rather than
+  gates, because a few specs need a retry against the shared demo backend.
```
patzick: Same drift: "that job reports rather than gates" describes the `continue-on-error` version. It now fails and pages Slack, and it no longer runs on pull requests at all.

Suggested:

```md
  against any storefront. CI runs them against `vue-starter-template` nightly in
  `.github/workflows/e2e-starter-template.yml`; nothing is gated on that job,
  but a failure notifies Slack via `failed-job-check.yml`.
```

.github/workflows/e2e-starter-template.yml
```diff
@@ -0,0 +1,123 @@
+name: E2E Starter Template
+
+run-name: Playwright e2e suite against vue-starter-template 🚀
+on:
+  workflow_dispatch:
+  schedule:
```
patzick, line 6: Consequence worth flagging rather than a defect: with the `pull_request` trigger gone, this workflow no longer runs on this PR. The last `E2E Starter Template` run was on `9215075` — before the `ListingPage` panel-scoping rework, the `checkForSeoUrl` render wait and the `listing-filter-${filter.code}` testid this branch depends on. So the suite changes made in response to review have never actually been run.

A `workflow_dispatch` on `ci/e2e-suite-against-starter` before merge would confirm them, otherwise the first evidence arrives on the 04:00 nightly against `main`.

Also note `concurrency.group` lost its `github.ref`, which is correct now that only one branch schedules it, but means a manual dispatch queues behind an in-flight nightly instead of running alongside it.

patzick (changes requested):
## Re-review — all 9 previous findings addressed

Every item from [my last review](https://github.com/shopware/frontends/pull/2677#pullrequestreview-5151782733) is fixed, and correctly:

- **The blocking one** — `hasActiveFilter` fallback restored to `properties.size > 0` with a comment explaining why it must stay a fallback. ✅
- `limit` now resyncs from the URL inside the watcher. ✅
- `buildCriteria()` normalises query strings to numbers/booleans, drops unknown params, and the unreachable `post-filter` branch plus the unused `searchCriteriaForRequest` export are gone. Rejections are now caught rather than floating. ✅
- `scrollToListing()` waits on `until(loading).toBe(false)`. ✅ (Vue's pre-flush scheduler queues the route watcher before `router.push` resolves, so `loading` is already `true` by then — the wait is real, not a no-op.)
- `ListingPage` scopes the option to the panel via the new `listing-filter-${filter.code}` testid, and matches filter codes instead of localised button text. ✅ Nice generalisation.
- `checkForSeoUrl` waits for the destination to render. ✅
- `countEntries()` removed; SSR logging buffered with a timer + exit flush. ✅

And @mdanilowicz's `failed-job-check.yml` thread is resolved properly — `continue-on-error` dropped, workflow added to the notify list.

## Two things from the latest push

1. **The docs now contradict the workflow** (blocking, two lines). Dropping the `pull_request` trigger and `continue-on-error` left `apps/e2e-tests/README.md:27,29` and `AGENTS.md:131` describing behaviour that no longer exists. Inline.

2. **The harness fixes are unverified.** With the PR trigger gone, `E2E Starter Template` no longer runs here — the last run was on `9215075`, before the `ListingPage` rework, the `checkForSeoUrl` wait and the new `listing-filter-*` testid. Those are exactly the changes most likely to break the suite, and nothing has exercised them. A `workflow_dispatch` run on `ci/e2e-suite-against-starter` before merge would close it.

The scheduled-only trade-off itself reads as the right call given the Slack notification — just worth confirming the team is happy that the PR titled "run the e2e suite against vue-starter-template" runs nightly rather than per-PR, since #2649 closes with it.

mkucmus:
Both fixed, and thanks for catching the docs drift. That is exactly the kind of thing the change should not have left behind.

**1. Docs.** `apps/e2e-tests/README.md` no longer claims a pull request trigger, and both it and `AGENTS.md` now say the job runs nightly and on dispatch, that nothing is gated on it, and that a failure notifies Slack. I swept the rest of the repo for the same two claims: the only remaining mention of `continue-on-error` is the comment in the workflow explaining why it is absent, which is intentional.

While in there I found one more the review did not name. The README's "To reproduce the CI run locally" block built with `pnpm --filter vue-starter-template build`, while the workflow builds through `turbo run build --filter=vue-starter-template`. It works locally because postinstall builds the packages, but it is not what the workflow does, and it silently picks up stale linked packages if they are not built. Aligned. The second occurrence, under the accessibility section, is left alone since it makes no claim about reproducing CI.

**2. Verification.** Fair point, and I had the same worry. Dispatched a run on this branch: https://github.com/shopware/frontends/actions/runs/34515297522

I will report the result here either way rather than only if it is green.

**On the trade-off.** Agreed it should be an explicit team call rather than something that slips through with the PR. To be clear about what changed and why: the path filter matched 19 of the last 25 merged PRs at a median of 12 minutes a run, which is the same trade #2706 made for the example builds. #2649 asked for the suite to run in CI, and it does; nightly rather than per-PR, with a Slack notification on failure. If the team would rather have it per-PR, the trigger is three lines and I am happy to put it back.

mkucmus:
Dispatch run finished green: https://github.com/shopware/frontends/actions/runs/34515297522

**18 passed, 6 flaky, 2 skipped, 0 failed**, 13.5 minutes, on `507e32fc`. Every step succeeded including `Run tests`, which is the one that counts now: with `continue-on-error` gone, a real failure fails the job.

That covers what you flagged as unexercised. `507e32fc` is downstream of `da17a8a7`, so this run did exercise the `ListingPage` rework, the `checkForSeoUrl` wait and the new `listing-filter-*` testid.

Two caveats so the number is not read as more than it is. The 6 flaky are the retry-absorbed store-api drops tracked in #2682, not new; the count moves run to run. And the run predates the docs commit above, which is fine since docs cannot affect the suite, but it does mean the green tick is on the code as reviewed rather than on the branch tip.

patzick (approved):
## Approved

Both items from [the last round](https://github.com/shopware/frontends/pull/2677#pullrequestreview-5169730511) are closed, and my three threads are resolved:

- `288de64e` corrects `README.md` and `AGENTS.md` to "nightly and on manual dispatch, not on pull requests" and "nothing is gated, a failure notifies Slack". The local-repro snippet also switched to `turbo run build`, which matches the workflow. ✅
- The `workflow_dispatch` run on `507e32f` succeeded, so the `ListingPage` panel-scoping rework, the `checkForSeoUrl` wait and the `listing-filter-${filter.code}` testid have all actually executed. ✅

The merge from main shrank this PR usefully: #2700 landed the `createAccount` / `guest: !createAccount` rework independently, so `checkout/index.vue` is down to two `data-testid` lines here and `CustomerBaseInfo.vue` to three. The version in main is better than what I reviewed — `baseInfoRules` makes the password rule conditional on `createAccount`, so `handleRegister` can validate the whole form in one check instead of special-casing email. `hasActiveFilter`'s property-group fallback survived the merge intact.

### Two follow-ups, neither blocking

1. **Worth one more `workflow_dispatch` on the merged head before merging.** The last e2e run predates the merge, and the merge is exactly what the checkout and address specs drive: #2700 restructured `handleSaveAddress` into `handleRegister` with an `isRegistering` state and put the whole form behind `v-if="!isUserSession"`, while #2698 reworked `CountrySearchSelect` — which `utils/form.ts:selectCountry` probes by tag name. The testids all line up on inspection, so I expect it to pass; it is just cheap certainty on the one thing nothing has exercised.

2. **@copilot's `fixtures.ts` point is fair.** `NETWORK_LOG` and `ATTEMPT_LOG` are appended at fixed relative paths and never truncated, so locally the second run's `summarise-network.mjs` output mixes in the first run's calls and the failure rate stops meaning anything. Harmless on a fresh runner. Truncating once per run (or keying the filename on the run) would fix it — fine as a follow-up.

Still open and yours to close: @mdanilowicz's two threads on `playwright.config.ts:34` and the `@frontends` tag name, both already answered.

PR #2681, opened 2026-09-01
feat(docs): add Order History section and update related documentation

mdanilowicz:

### DIFF
--- a/apps/docs/src/frontends-recipes/account/order-history.md
+++ b/apps/docs/src/frontends-recipes/account/order-history.md
@@ -1,0 +1,354 @@
+---
+nav:
+  position: 30
+recipe:
+  area: account
+  status: stable
+  frameworks:
+    - vue
+  composables:
+    - useCustomerOrders
+    - useOrderDetails
+    - useDefaultOrderAssociations
+    - useUser
+  helpers: []
+  operations:
+    - readOrder post /order
+    - orderSetPayment post /order/payment
+    - handlePaymentMethod post /handle-payment
+    - cancelOrder post /order/state/cancel
+    - readPaymentMethod post /payment-method
+  schemas:
+    - Order
+    - OrderRouteResponse
+    - OrderLineItem
+    - StateMachineState
+    - Criteria
+---
+
+<script setup>
+import RecipeFlowDiagram from "../../components/RecipeFlowDiagram.vue";
+import SchemaTypeTooltip from "../../components/SchemaTypeTooltip.vue";
+
+const steps = [
+  {
+    title: "UI",
+    action: "Open the order history",
+    detail:
+      "The page passes a criteria to useCustomerOrders. Everything it wants to render on a row has to be requested here, because the composable adds no associations of its own.",
+    code: "loadOrders({ page, associations, sort })",
+    state: "local page state",
+    typeKeys: ['operations["readOrder post /order"]["body"]'],
+  },
+  {
+    title: "Composable",
+    action: "Force the limit and the exact count",
+    detail:
+      "useCustomerOrders spreads your parameters, then overwrites limit with its own limit ref and merges total-count-mode set to exact. The limit you pass in the criteria never reaches the Store API.",
+    code: 'body: { ...params, limit, "total-count-mode": "exact" }',
+    state: "limit ref",
+    typeKeys: ['Schemas["Criteria"]'],
+  },
+  {
+    title: "Store API",
+    action: "Read one page of orders",
+    detail:
+      "readOrder resolves the customer from the context token. No customer id is part of the request. The response carries one page of orders plus a paymentChangeable map keyed by order id.",
+    code: 'apiClient.invoke("readOrder post /order")',
+    state: "sw-context-token",
+    typeKeys: [
+      'operations["readOrder post /order"]["response"]',
+      'Schemas["OrderRouteResponse"]',
+    ],
+  },
+  {
+    title: "Composable state",
+    action: "Keep elements, total, page",
+    detail:
+      "useCustomerOrders stores orders.elements, orders.total, and orders.page in its own refs. It reads nothing else from the response, so paymentChangeable is dropped on the list.",
+    code: "orders + currentPage + totalPages",
+    state: "one page of orders",
+    typeKeys: ['Schemas["Order"]'],
+  },
+  {
+    title: "UI",
+    action: "Open one order",
+    detail:
+      "A row links to a detail route that calls useOrderDetails with the order id. The row data is not handed over; the detail view loads the order again.",
+    code: "useOrderDetails(orderId).loadOrderDetails()",
+    state: "provided swOrderDetails ref",
+    typeKeys: [],
+  },
+  {
+    title: "Store API",
+    action: "Read the full order",
+    detail:
+      "loadOrderDetails calls the same readOrder operation with ids set to one order, the associations from useDefaultOrderAssociations, and checkPromotion set to true.",
+    code: 'body: { ids: [orderId], associations, checkPromotion: true }',
+    state: "line items, deliveries, transactions",
+    typeKeys: [
+      'operations["readOrder post /order"]["body"]',
+      'Schemas["OrderLineItem"]',
+    ],
+  },
+  {
+    title: "UI",
+    action: "Render the order",
+    detail:
+      "The detail view reads order, status, total, paymentMethod, and paymentChangeable from useOrderDetails instead of keeping its own copy of the row it came from.",
+    code: "order + status + total + paymentChangeable",
+    state: "reactive UI",
+    typeKeys: ['Schemas["StateMachineState"]'],
+  },
+];
+</script>
+
+# Order History
+
+## Goal
+
+Build a paginated order history for a logged-in customer and a detail view for a single order. The important part is that both screens call the same `readOrder post /order` operation, and that what an order contains depends entirely on the associations you request, not on the generated `Order` type.
+
+## Shopware Flow
+
+`readOrder post /order` is a criteria route secured with the access key and the context token. There is no customer id in the request: the Store API returns the orders of whoever the `sw-context-token` header resolves to.
+
+`useCustomerOrders().loadOrders()` sends only your criteria plus `limit` from its own ref and `"total-count-mode": "exact"`. It adds no associations, so a list row has `orderNumber`, `orderDate`, and `amountTotal` but no state, line items, deliveries, or transactions until you ask for them.
+
+Hover a type chip to inspect fields generated from the current Store API schema.
+
+<RecipeFlowDiagram label="Order history flow diagram" :steps="steps" />
+
+Read the diagram from left to right:
+
+1. The customer opens the order history and the page calls `loadOrders` with a criteria.
+2. `useCustomerOrders` overwrites the `limit` in that criteria with its own `limit` ref and sets `"total-count-mode": "exact"`.
+3. The Store API returns one page of orders for the customer behind the context token, together with a `paymentChangeable` map.
+4. The composable keeps `orders.elements`, `orders.total`, and `orders.page`, and exposes `orders`, `currentPage`, and `totalPages`.
+5. The customer opens one order and the detail view calls `useOrderDetails(orderId).loadOrderDetails()`.
+6. `loadOrderDetails` calls `readOrder post /order` again with `ids`, the default order associations, and `checkPromotion: true`.
+7. The UI reads `order`, `status`, `total`, and `paymentChangeable` from composables instead of keeping its own copy.
+
+You do not assemble the detail criteria yourself. `useOrderDetails` calls `useDefaultOrderAssociations()` and merges it with the associations you pass as its second argument, with the defaults taking precedence on conflicting keys.
+
+## Request Flow
+
+| Step                      | Code                                   | Store API                  | Type                                                                                              |
+| ------------------------- | -------------------------------------- | -------------------------- | ------------------------------------------------------------------------------------------------- |
+| Load a page of orders     | `loadOrders(criteria)`                 | `POST /order`              | <SchemaTypeTooltip type-key='operations["readOrder post /order"]["body"]' />                      |
+| Change the page           | `changeCurrentPage(page)`              | `POST /order`              | <SchemaTypeTooltip type-key='operations["readOrder post /order"]["response"]' />                  |
+| Load one order            | `loadOrderDetails()`                   | `POST /order`              | <SchemaTypeTooltip type-key='Schemas["OrderRouteResponse"]' />                                    |
+| List selectable payments  | `getPaymentMethods()`                  | `POST /payment-method`     | <SchemaTypeTooltip type-key='operations["readPaymentMethod post /payment-method"]["response"]' /> |
+| Change the payment method | `changePaymentMethod(paymentMethodId)` | `POST /order/payment`      | <SchemaTypeTooltip type-key='operations["orderSetPayment post /order/payment"]["body"]' />        |
+| Start the payment again   | `handlePayment(finishUrl, errorUrl)`   | `POST /handle-payment`     | <SchemaTypeTooltip type-key='operations["handlePaymentMethod post /handle-payment"]["body"]' />   |
+| Cancel the order          | `cancel()`                             | `POST /order/state/cancel` | <SchemaTypeTooltip type-key='operations["cancelOrder post /order/state/cancel"]["response"]' />   |
+
+## Composables
+
+- `useCustomerOrders`: the order list. Exposes `loadOrders`, `changeCurrentPage`, `orders`, `currentPage`, `totalPages`, and `limit`. `limit` is a writable ref, and `loadOrders` uses it instead of any `limit` in your criteria.
+- `useOrderDetails`: one order, created with `useOrderDetails(orderId, associations?)`. Exposes `loadOrderDetails`, `order`, `status`, `statusTechnicalName`, `total`, `subtotal`, `shippingCosts`, `billingAddress`, `shippingAddress`, `personalDetails`, `shippingMethod`, `paymentMethod`, `paymentChangeable`, `getPaymentMethods`, `changePaymentMethod`, `handlePayment`, `paymentUrl`, `cancel`, `documents`, `hasDocuments`, `getMediaFile`, and `getDocumentFile`.
+- `useDefaultOrderAssociations`: the criteria associations `useOrderDetails` sends. It returns `stateMachineState`, `lineItems` with `cover` and `downloads.media`, `addresses`, `deliveries` with `shippingMethod`, `shippingOrderAddress` and `stateMachineState`, and `transactions` with `paymentMethod` and `stateMachineState`. Override it in your project when every order view needs a different shape.
+- `useUser`: provides `isLoggedIn`, which decides whether the context token resolves to a customer and the order routes can answer at all.
+
+## Types
+
+Use generated Store API types when you type the list criteria, the response, or the order you render:
+
+<div style="display: flex; flex-wrap: wrap; gap: 6px; margin: 12px 0 18px;">
+  <SchemaTypeTooltip type-key='operations["readOrder post /order"]["body"]' />
+  <SchemaTypeTooltip type-key='operations["readOrder post /order"]["response"]' />
+  <SchemaTypeTooltip type-key='Schemas["OrderRouteResponse"]' />
+  <SchemaTypeTooltip type-key='Schemas["Order"]' />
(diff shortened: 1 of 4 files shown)


apps/docs/src/frontends-recipes/account/order-history.md
```diff
@@ -0,0 +1,354 @@
...
+    </template>
+  </section>
+</template>
+```
+
+## State And Session
+
+The Store API resolves the customer of `readOrder post /order`, `orderSetPayment post /order/payment`, and `cancelOrder post /order/state/cancel` from the `sw-context-token` header. Nothing in the request names the customer, so the order history changes with the session and not with a route parameter.
+
+`handlePaymentMethod post /handle-payment` is the exception: it is secured with the access key alone and identifies the order by `orderId` in the body.
```
patzick, line 304: **Major — factual error.** `/handle-payment` is *not* secured with the access key alone. In `packages/api-client/api-types/storeApiSchema.json` (6.7.13.0):

```
/handle-payment  → security: [{ ApiKey: [], ContextToken: [] }]
/order           → security: [{ ApiKey: [], ContextToken: [] }]
/payment-method  → security: [{ ApiKey: [] }]
```

So it is the same security as the other order routes, not an exception. The "access key alone" claim only holds in the stale docs copy `apps/docs/src/.data/schema/storeApiSchema.json` (6.7.10.0).

Suggest dropping the sentence, or rewriting it as: `handlePaymentMethod` still identifies the order explicitly via `orderId` in the body rather than relying on the session alone — which is the part that is actually useful here.

apps/docs/src/frontends-recipes/account/order-history.md
```diff
@@ -0,0 +1,354 @@
...
+
+The two composables hold state differently. `useCustomerOrders` creates its refs per call, so two components calling it have two independent lists. `useOrderDetails` injects and provides a single `swOrderDetails` ref, so a component tree below the first caller shares one order object.
+
+Guest orders reach the same route with `email`, `zipcode`, and an `equals` filter on `deepLinkCode` in the body. With `login: true`, the response carries an `sw-context-token` header, and the API client adopts that token as its new default header when the response is not publicly cacheable. Neither order composable sends those fields, so a guest order page calls `apiClient.invoke("readOrder post /order")` directly.
+
+## Edge Cases
+
+- `loadOrders` builds its body as `{ ...parameters, limit: limit.value }`. A `limit` in your criteria is overwritten, so page sizes are changed through `limit.value` before the call.
+- The list request sends no associations. `Schemas["Order"]` declares `stateMachineState` and `documents` as required, but they are associations and stay undefined at runtime until you request them.
```
patzick, line 315: **Major — wrong for `documents`.** The claim holds for `stateMachineState`, but not for `documents`:

- `useDefaultOrderAssociations()` never requests `documents`.
- `useOrderDetails` computes `hasDocuments` as `!!_sharedOrder.value?.documents.length` — the optional chain stops at `_sharedOrder.value`, so an undefined `documents` would throw a `TypeError`.
- Both starter templates render `<AccountOrderDownloads v-if="hasDocuments" …>` on a default `loadOrderDetails()` (`templates/vue-demo-store/app/components/account/order/detail/View.vue:208`, `templates/vue-starter-template/app/components/account/order/Details.vue:156`), and `packages/composables/src/mocks/Order.ts:1124` carries `documents: []`.

The route returns `documents` regardless of the criteria. Please limit the bullet to `stateMachineState`.

apps/docs/src/frontends-recipes/account/order-history.md
```diff
@@ -0,0 +1,354 @@
...
+- The same applies to `lineItems`, `deliveries`, and `transactions`. A list row cannot show the ordered products, the shipping method, or the payment state unless the criteria asks for those associations.
```
patzick, line 316: **Minor.** "The same applies" loses the point of the previous bullet. `lineItems`, `deliveries` and `transactions` are declared **optional** (`lineItems?:`, `deliveries?:`, `transactions?:`) in `Schemas["Order"]`, so there is no required-vs-undefined mismatch for them — the type already tells you they may be absent. Worth saying so explicitly, since that contrast is what makes `stateMachineState` a trap.

apps/docs/src/frontends-recipes/account/order-history.md
```diff
@@ -0,0 +1,354 @@
...
+- `useCustomerOrders` reads only `data.orders` from the response and discards `paymentChangeable`. Sending `checkPromotion: true` on the list gains nothing through this composable; `paymentChangeable` is available on `useOrderDetails`, whose request always sets it.
+- `totalPages` is computed as `orders.total` divided by the current `limit` ref. Writing `limit.value` after a load changes `totalPages` without reloading anything.
```
patzick, line 318: **Minor.** It is `Math.ceil(totalOrderItemsCount.value / limit.value)` (`packages/composables/src/useCustomerOrders/useCustomerOrders.ts:79`). "Divided by" drops the ceiling, which is exactly what makes the last partial page reachable in the `v-if="currentPage < totalPages"` example above.

apps/docs/src/frontends-recipes/account/order-history.md
```diff
@@ -0,0 +1,354 @@
...
+      <p v-else-if="!orders.length">You have not placed an order yet.</p>
+
+      <template v-else>
+        <ul>
+          <li v-for="order in orders" :key="order.id">
+            <a :href="`/account/order/details/${order.id}`">
+              {{ order.orderNumber }}
+            </a>
+            <span>{{ formatDate(order.orderDate) }}</span>
+            <span>{{ order.amountTotal }}</span>
```
patzick, line 270: **Nit.** `amountTotal` is rendered raw, so the example prints e.g. `123.45` with no currency. Either format it with `usePrice().getFormattedPrice(order.amountTotal)` or add a short comment that formatting is left out on purpose — copy-paste recipes tend to be taken literally.

apps/docs/src/frontends-recipes/account/order-history.md
```diff
@@ -0,0 +1,354 @@
...
+// Immediate watcher instead of onMounted: the orders belong to the customer
+// behind the context token, so the list is reloaded when the session changes.
+watch(
+  isLoggedIn,
+  (loggedIn) => {
+    if (loggedIn) {
+      loadFirstPage();
+    }
+  },
+  { immediate: true }
```
patzick, line 248: **Nit.** Missing trailing comma: the sibling `wishlist.md` recipe ends the same watcher with `{ immediate: true },`. Cosmetic only — `apps/docs` has no `lint` script, so nothing enforces it.

patzick (changes requested):
## 🔍 Code Review — docs recipe

Good, well-grounded recipe: composable signatures, `defu` precedence, the `limit`-overwrite behaviour, the guest-auth criteria shape and all `SchemaTypeTooltip` type keys check out against the current schema. Two factual claims contradict the shipped code/schema, plus two smaller accuracy points — see inline comments.

**Blockers:** none · **Major:** 2 · **Minor:** 2 · **Nits:** 2

CI: all 9 checks green. No merge conflicts. PR is still a draft and the description is the unfilled template.

Side note (not this PR): `apps/docs/src/.data/schema/storeApiSchema.json` is pinned at **6.7.10.0** while `packages/api-client` ships **6.7.13.0** — the `schema:update:store` script copies to `.data/schema/…` relative to `apps/docs`, but the file lives in `apps/docs/src/.data/schema/`, so it never lands. That stale copy is what the tooltips on this page render.

mdanilowicz:
@patzick this is a draft, not ready for review

apps/docs/src/frontends-recipes/account/order-history.md
```diff
@@ -0,0 +1,440 @@
...
+- **Documents** — `documents`, `hasDocuments`, `getDocumentFile`, `getMediaFile`.
+
+Six things the generated reference will not tell you:
+
+- `limit` on `useCustomerOrders` is a writable ref that starts at `15`, and `loadOrders` overwrites whatever `limit` your criteria carries with it. Page sizes are changed through `limit.value`, never through the criteria.
+- `useOrderDetails(orderId, associations?)` does not let you replace the default criteria. The second argument is merged **under** `useDefaultOrderAssociations()`, so the defaults win on conflicting keys. To change the shape of every order view, override `useDefaultOrderAssociations` in your project instead.
+- `handlePayment` declares a third `paymentDetails` parameter that the implementation never sends. Only `orderId`, `finishUrl`, and `errorUrl` reach the Store API.
+- `handlePayment()` stores the returned `redirectUrl` in `paymentUrl` and does nothing else. No navigation happens until you watch `paymentUrl` and redirect yourself — and check its scheme first, because `new URL()` parses `javascript:` and `data:` without throwing. The [Payment recipe](../checkout/payment.html) owns that guard.
```
patzick, line 179: Wording is slightly off: `handlePayment`'s actual runtime implementation doesn't declare a third parameter at all — only the TS interface type declares `paymentDetails`. "the implementation never sends" implies it's received then dropped, but it's never accepted by the function signature in the first place.

apps/docs/src/frontends-recipes/account/order-history.md
```diff
@@ -0,0 +1,440 @@
...
+
+## State And Session
+
+The Store API resolves the customer of `readOrder post /order`, `orderSetPayment post /order/payment`, and `cancelOrder post /order/state/cancel` from the `sw-context-token` header. Nothing in the request names the customer, so the order history changes with the session and not with a route parameter.
+
+`handlePaymentMethod post /handle-payment` names its target explicitly instead: the order comes from `orderId` in the body, not from the session. `readPaymentMethod post /payment-method` is secured with the access key alone, because payment methods are sales-channel data rather than customer data. Before relying on that for any other route, read the `security` block of the schema your backend ships: `/handle-payment` declared the access key alone through 6.7.10 and gained `ContextToken` by 6.7.13.
+
```
patzick, line 388: Can we cite a source for "declared the access key alone through 6.7.10 and gained ContextToken by 6.7.13"? The schema snapshots in this repo only go up to 6.7.10.1, so this can't be verified from the codebase — a link to the core changelog/PR would help future readers trust the cutoff.

patzick (changes requested):
Docs are accurate against the composable source overall — a couple of precision nits inline.

patzick:
[Automation] PR looks good

apps/docs/src/frontends-recipes/account/order-history.md
```diff
@@ -0,0 +1,441 @@
...
+- The two composables hold state differently. `useCustomerOrders` creates its refs per call, so two components calling it keep two independent lists. `useOrderDetails` injects and provides one shared `swOrderDetails` ref, so everything below the first caller reads the same order object.
+- `getMediaFile(downloadId)` returns a `Blob` from `orderDownloadFile get /order/download/{orderId}/{downloadId}`. `getDocumentFile(documentId, deepLinkCode)` returns `Blob | string` from `download post /document/download/{documentId}/{deepLinkCode}`, because a document can be a PDF, HTML, or XML — narrow it before you hand it to `URL.createObjectURL`. Both are about attachments, not about the order body. `documents` is the one field that arrives without being asked for: `useDefaultOrderAssociations()` never requests it, but the route returns it anyway — which is why `hasDocuments` gets away with reading `order.documents.length` unguarded, while `documents` itself still falls back to an empty array.
```
patzick: **Minor — misleading.** `getDocumentFile` hardcodes `accept: "application/pdf"` (`packages/composables/src/useOrderDetails/useOrderDetails.ts:265-278`). The operation is a discriminated union on `accept` (`storeApiTypes.d.ts` — `application/pdf` → `Blob`, `application/xml` / `text/html` → `string`), so the `string` arms are unreachable through this composable. The `Blob | string` is just the hand-written annotation on `UseOrderDetailsReturn`.

As written a reader will expect HTML/XML documents back. Suggest: "…returns `Blob | string` because the declared type covers every `accept` variant of the operation, but `getDocumentFile` always sends `accept: \"application/pdf\"`, so you get the `Blob` arm — you still have to narrow it to satisfy TypeScript before `URL.createObjectURL`."

apps/docs/src/frontends-recipes/account/order-history.md
```diff
@@ -0,0 +1,441 @@
...
+
+Hover a type chip to inspect fields generated from the current Store API schema.
+
+<RecipeFlowDiagram label="Order history flow diagram" :steps="steps" />
+
+Read the diagram from left to right:
+
+1. The customer opens the order history and the page calls `loadOrders` with a criteria.
+2. `useCustomerOrders` overwrites the `limit` in that criteria with its own `limit` ref and sets `"total-count-mode": "exact"`.
+3. The Store API returns one page of orders for the customer behind the context token, together with a `paymentChangeable` map.
```
patzick: **Minor — contradicts the diagram above.** Step 3 says the list response comes "together with a `paymentChangeable` map", but the diagram at L59 says it arrives "only when the request sent `checkPromotion`, which the list never does", and L71 says it is dropped on the list. The numbered walkthrough is the part a reader trusts.

Suggest dropping the clause: "3. The Store API returns one page of orders for the customer behind the context token. The list request sends no `checkPromotion`, so no `paymentChangeable` map comes back."

apps/docs/src/frontends-recipes/account/order-history.md
```diff
@@ -0,0 +1,441 @@
...
+- Opening an order calls `readOrder post /order` with `ids` set to that order and `checkPromotion: true`, and fills `order`, `status`, and `paymentChangeable`.
+- `getPaymentMethods()` calls `readPaymentMethod post /payment-method` with `onlyAvailable` set to true.
+- `changePaymentMethod()` calls `orderSetPayment post /order/payment` and reloads the order details.
+- `cancel()` calls `cancelOrder post /order/state/cancel` and reloads the order so `statusTechnicalName` reflects the cancelled state.
+- `handlePayment()` calls `handlePaymentMethod post /handle-payment` and sets `paymentUrl` to the returned redirect URL.
+- A failing request shows a list-level error instead of an empty order history, keeps the previously loaded page on screen, and offers a retry.
+- Starting a page change announces the loading state and leaves focus on the pagination control that triggered it.
+- A `paymentUrl` whose scheme is not `https:` — including `javascript:` and `data:` — does not trigger a navigation.
+- Order dates and totals render in the storefront's locale and currency, not the host default.
```
patzick: **Minor — factually wrong for the locale half.** `browserLocale` is the *visitor's browser* locale, not the storefront's: `packages/nuxt-module/plugin.ts:184-192` sets it from `navigator.language` (client) or the `accept-language` header (SSR), and `useShopwareContext.ts` documents it as "Browser locale, working in SSR". An `en-US` browser on a `de-DE` storefront gets US dates.

The currency half is correct (`usePrice` watches `sessionContext.currency`). Suggest: "Order dates render in the visitor's browser locale (`browserLocale`) and totals in the session currency, not the host default." Same correction applies to the code comment at L230.

apps/docs/src/frontends-recipes/account/order-history.md
```diff
@@ -0,0 +1,441 @@
...
+
+Neither `useCustomerOrders` nor `useOrderDetails` refreshes the session context or the cart. Unlike login, reading orders does not change the session, so `orders` simply keeps describing the customer that was authenticated when the request was sent.
+
+Guest orders reach the same route with `email`, `zipcode`, and an `equals` filter on `deepLinkCode` in the body. With `login: true`, the response carries an `sw-context-token` header, and the API client adopts that token as its new default header when the response is not publicly cacheable. Neither order composable sends those fields, so a guest order page calls `apiClient.invoke("readOrder post /order")` directly.
```
patzick: **Nit.** "Neither order composable sends those fields" is only true of `useOrderDetails`. `useCustomerOrders.loadOrders(parameters)` takes the full `operations["readOrder post /order"]["body"]` and spreads it (`useCustomerOrders.ts:54-64`), so it *will* forward `email`, `zipcode`, `filter` and `login`.

Suggest: "Neither composable adds those fields for you, and `useOrderDetails` cannot send them at all, so a guest order page calls `apiClient.invoke(\"readOrder post /order\")` directly."

patzick (changes requested):
Re-review at `7a65c2b`. All 8 findings from the previous rounds are fixed — verified against source, not against the doc text. CI is green.

Four new precision issues from a fresh pass over the recipe, all inline. None are blocking; #1 and #3 are factually wrong as written.

apps/docs/src/frontends-recipes/account/order-history.md
```diff
@@ -0,0 +1,447 @@
...
+
+Hover a type chip to inspect fields generated from the current Store API schema.
+
+<RecipeFlowDiagram label="Order history flow diagram" :steps="steps" />
+
+Read the diagram from left to right:
+
+1. The customer opens the order history and the page calls `loadOrders` with a criteria.
+2. `useCustomerOrders` overwrites the `limit` in that criteria with its own `limit` ref and sets `"total-count-mode": "exact"`.
+3. The Store API returns one page of orders for the customer behind the context token. The list request sends no `checkPromotion`, so no `paymentChangeable` map comes back.
```
patzick: **Correctness.** This is wrong as an absolute — and it is the wording I suggested last round, so the correction is on me.

`loadOrders` spreads the caller's body verbatim (`packages/composables/src/useCustomerOrders/useCustomerOrders.ts:58-64`), and both reference templates *do* send `checkPromotion` on the list request:

- `templates/vue-demo-store/app/pages/account/order/index.vue:90` — `checkPromotion: true, // determines whether "paymentChangeable" list is available in the response`
- `templates/vue-starter-template/app/pages/account/order/index.vue:16` — `checkPromotion: true`

Step 3's own `detail` string at L59 has it right ("which `useCustomerOrders` does not add"). Suggest matching it:

> 3. The Store API returns one page of orders for the customer behind the context token. `useCustomerOrders` adds no `checkPromotion` of its own, so unless your criteria passes one no `paymentChangeable` map comes back.

apps/docs/src/frontends-recipes/account/order-history.md
```diff
@@ -0,0 +1,447 @@
...
+
+## State And Session
+
+The Store API resolves the customer of `readOrder post /order`, `orderSetPayment post /order/payment`, and `cancelOrder post /order/state/cancel` from the `sw-context-token` header. Nothing in the request names the customer, so the order history changes with the session and not with a route parameter.
+
+`handlePaymentMethod post /handle-payment` names its target explicitly instead: the order comes from `orderId` in the body, not from the session. `readPaymentMethod post /payment-method` is secured with the access key alone, because payment methods are sales-channel data rather than customer data. Before relying on that for any other route, read the `security` block of the schema your backend ships rather than assuming it: the declared requirement of a route can change between Shopware releases, and `/handle-payment` is one route where it has.
```
patzick: **Nit — one word.** Under a heading about session resolution, "not from the session" reads as "does not depend on the session". At the schema this branch ships (`packages/api-client/api-types/storeApiSchema.json`, 6.7.13.0), `/handle-payment` `security` is `[{ApiKey, ContextToken}]` — identical to the three routes in the sentence before. Only the *order identity* comes from `orderId`; the route still needs the token.

Suggest "…not from the session alone."

(The closing clause about the requirement changing between releases is correct — the ApiKey-only form is what the stale docs copy at 6.7.10.0 still declares.)

patzick (changes requested):
Re-review at `3f0a283b`. All four findings from the `7a65c2b` round are fixed, verified against source. CI green.

One factual error and one precision nit inline.

patzick (commented):
Re-review at `fa49ca87`. Both findings from the `3f0a283b` round are fixed, verified against source:

- L129 — `useCustomerOrders` adds no `checkPromotion` of its own: `loadOrders` spreads `...parameters` verbatim (`useCustomerOrders.ts:58-65`), so a caller-supplied one does reach the API. Correct as now worded.
- L387 — `/handle-payment` security is `[{ApiKey, ContextToken}]` in the schema this branch ships; `/payment-method` is `[{ApiKey}]`. Correct as now worded.

Fresh pass over the whole recipe found nothing factually wrong: all 7 operation ids/routes and their security blocks, request/response fields and all 11 `SchemaTypeTooltip` type keys resolve; every `useCustomerOrders` / `useOrderDetails` / `useDefaultOrderAssociations` claim matches source (defu precedence, shared `swOrderDetails`, `Math.ceil` pagination, type-only `paymentDetails`, hardcoded `accept: application/pdf`, unrequested `documents`); the Vue example type-checks against `Schemas["Order"]` and `browserLocale` is a plain string; guest error codes and the `vue-starter-template` comparison at L381 check out; sidebar/index registration and all cross-links resolve.

No change requests from my side. CI green (9/9).

patzick (approved):

PR #2689, opened 2026-09-02
feat(docs): add registration page and update sidebar and account documentation page

mdanilowicz:

### DIFF
--- a/apps/docs/src/frontends-recipes/account/register.md
+++ b/apps/docs/src/frontends-recipes/account/register.md
@@ -1,0 +1,493 @@
+---
+nav:
+  position: 20
+recipe:
+  area: account
+  status: stable
+  frameworks:
+    - vue
+  composables:
+    - useUser
+    - useSalutations
+    - useCountries
+    - useSessionContext
+    - useInternationalization
+    - useShopwareContext
+  helpers:
+    - getTranslatedProperty
+  operations:
+    - register post /account/register
+    - registerConfirm post /account/register-confirm
+    - readContext get /context
+    - readSalutation post /salutation
+    - readCountry post /country
+    - getCustomerGroupRegistrationInfo get /customer-group-registration/config/{customerGroupId}
+  schemas:
+    - Customer
+    - CustomerAddress
+    - Salutation
+    - Country
+    - CountryState
+    - CustomerGroup
+---
+
+<script setup>
+import RecipeFlowDiagram from "../../components/RecipeFlowDiagram.vue";
+import SchemaTypeTooltip from "../../components/SchemaTypeTooltip.vue";
+
+const steps = [
+  {
+    title: "UI",
+    action: "Collect the registration body",
+    detail:
+      "The form owns one object shaped like the register body. A billing address is part of that body, not a second request, and salutations and countries come from their own read routes before the customer can pick anything.",
+    code: "reactive<Omit<RegisterBody, 'storefrontUrl'>>({ ... })",
+    state: "local form state",
+    typeKeys: ['Schemas["Salutation"]', 'Schemas["Country"]'],
+  },
+  {
+    title: "Composable",
+    action: "Add the storefront URL",
+    detail:
+      "useUser().register() is the only place storefrontUrl is set. It calls useInternationalization().getStorefrontUrl(), which returns devStorefrontUrl or window.location.origin, and the Store API rejects a value that is not a configured sales channel domain.",
+    code: "register({ ...params, storefrontUrl: getStorefrontUrl() })",
+    state: "sw-context-token",
+    typeKeys: ['operations["register post /account/register"]["body"]'],
+  },
+  {
+    title: "Store API",
+    action: "Create the customer",
+    detail:
+      "The register route creates the customer in the sales channel session identified by the context token and responds with the full Customer, including active and doubleOptInRegistration.",
+    code: 'apiClient.invoke("register post /account/register")',
+    state: "customer record",
+    typeKeys: [
+      'operations["register post /account/register"]["response"]',
+      'Schemas["Customer"]',
+    ],
+  },
+  {
+    title: "Composable",
+    action: "Decide whether a session exists",
+    detail:
+      "register() assigns the returned customer to the shared customer context only when active is true and doubleOptInRegistration is false. A double opt-in registration therefore creates a customer without creating a customer session.",
+    code: "if (data.active && !data.doubleOptInRegistration) _user.value = data",
+    state: "shared customer context",
+    typeKeys: ['Schemas["Customer"]'],
+  },
+  {
+    title: "Context",
+    action: "Refresh the session context",
+    detail:
+      "register() awaits refreshSessionContext(), which reads GET /context and replaces the reactive session context. The customer on that response is what isLoggedIn is computed from.",
+    code: "await refreshSessionContext()",
+    state: "sessionContext",
+    typeKeys: ['operations["readContext get /context"]["response"]'],
+  },
+  {
+    title: "UI",
+    action: "Branch on the response",
+    detail:
+      "The component reads user, isLoggedIn, and isGuestSession from useUser, and uses the returned doubleOptInRegistration flag to show the confirmation notice instead of navigating to the account area.",
+    code: "user + isLoggedIn + isGuestSession + doubleOptInRegistration",
+    state: "reactive UI",
+    typeKeys: ['Schemas["Customer"]'],
+  },
+];
+</script>
+
+# Register
+
+## Goal
+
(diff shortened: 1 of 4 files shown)


apps/docs/src/frontends-recipes/account/register.md
```diff
@@ -0,0 +1,493 @@
...
+
+## Edge Cases
+
+- The generated `billingAddress` type is `Schemas["CustomerAddress"]`, which requires `id` and `customerId`. Send placeholder values as both starter templates do, because you cannot know the ids of an address that does not exist yet.
+- A registration with double opt-in enabled resolves successfully while leaving `isLoggedIn` false. Branch on the returned `doubleOptInRegistration` flag instead of assuming a session exists after the promise resolves.
+- `refreshSessionContext()` rethrows after logging, so a failing `readContext get /context` rejects the `register()` promise even though the customer was already created. A retry then hits `VIOLATION::CUSTOMER_EMAIL_NOT_UNIQUE`.
+- `getStorefrontUrl()` falls back to `window.location.origin`, which is empty during server-side rendering. Submit the form from the client, and set `devStorefrontUrl` when the Shopware domain differs from the origin your app runs on. An origin that is not a configured sales channel domain comes back as a constraint violation pointing at `/storefrontUrl`, which no field in your form owns.
```
patzick, line 460: **major — wrong SSR failure mode.** `getStorefrontUrl()` is `return devStorefrontUrl ?? window.location.origin ?? "";` (`useInternationalization.ts:92-94`). On the server `window` is not defined, so this **throws** a `ReferenceError` — it does not return an empty string. The advice that follows is right; only the stated mechanism is wrong.

Suggested: "…falls back to `window.location.origin`, which is not available during server-side rendering, so calling `register()` on the server throws."

apps/docs/src/frontends-recipes/account/register.md
```diff
@@ -0,0 +1,493 @@
...
+
+// A countryStateId left over from the previous country would be sent as is.
+watch(
+  () => form.billingAddress.countryId,
+  () => {
+    form.billingAddress.countryStateId = "";
+  }
+);
+
+// The generated body is a union, so accountType is a literal and vatIds can be
```
patzick, line 239: **minor — the payload type is not a union.** `Omit<T, K>` is `Pick<T, Exclude<keyof T, K>>`, which flattens `Body & (private | business)` into a single object type. On `RegisterPayload` the fields end up as `accountType: "private" | "business"`, `company: string | null | undefined`, `vatIds: [string, ...string[]] | null | undefined`, so `form.accountType = "business"` and `form.vatIds = [v]` compile directly (verified with tsc 5.9 --strict).

Two consequences worth stating instead:
- the computed models are a convenience for `v-model` + nullable fields, not a type requirement;
- the discrimination the *Types* section describes is **lost** on `RegisterPayload`, so nothing stops you sending `accountType: "business"` without a `company` — that is enforced by the Store API, not the compiler.

apps/docs/src/frontends-recipes/account/register.md
```diff
@@ -0,0 +1,493 @@
...
+| -------------------------------- | ---------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Load salutation options          | `fetchSalutations()`                                                                                             | `POST /salutation`                                          | <SchemaTypeTooltip type-key='operations["readSalutation post /salutation"]["response"]' />                                                            |
+| Load countries and their states  | `fetchCountries()`                                                                                               | `POST /country`                                             | <SchemaTypeTooltip type-key='operations["readCountry post /country"]["response"]' />                                                                  |
```
patzick, line 133: **minor — the route depends on `cacheableReads`.** Both composables switch to the GET variants when `cacheableReads` is set on the Shopware context:

- `useSalutations.ts:34-36` → `readSalutationGet get /salutation`
- `useCountries.ts:45-51` → `readCountryGet get /country`

So `POST /salutation` / `POST /country` in this table (and the two `operations` entries in the frontmatter) only describe the non-cacheable path. Worth listing both, as the rest of the page is precise about which route runs.

apps/docs/src/frontends-recipes/account/register.md
```diff
@@ -0,0 +1,493 @@
...
+
+`register()` does not call `refreshCart()`, and `login()` does. If the customer registers while a cart is on screen, call `refreshCart()` from `useCart` yourself, because line item prices, promotions, and rule matches are evaluated against the customer in the context.
+
+`storefrontUrl` is resolved inside `register()` by `getStorefrontUrl()`, which returns the configured `devStorefrontUrl` or `window.location.origin`. The Store API only accepts a value that matches a configured domain of the sales channel, and it is the base for the confirmation link in the double opt-in email.
+
+## Edge Cases
+
+- The generated `billingAddress` type is `Schemas["CustomerAddress"]`, which requires `id` and `customerId`. Send placeholder values as both starter templates do, because you cannot know the ids of an address that does not exist yet.
```
patzick, line 457: **minor — only one template does this.** `templates/vue-starter-template/app/components/account/RegistrationForm.vue:51-53` sends `id: ""` / `customerId: ""`, but `templates/vue-demo-store/app/components/account/RegistrationForm.vue:43-50` omits both (its state is untyped) and registers fine. So the placeholders are a way to satisfy the generated `CustomerAddress` type, not something the Store API needs. Please reword to say that, and drop "both starter templates".

apps/docs/src/frontends-recipes/account/register.md
```diff
@@ -0,0 +1,493 @@
...
+>;
+
+const form = reactive<RegisterPayload>({
+  accountType: "private",
+  salutationId: "",
+  firstName: "",
+  lastName: "",
+  email: "",
+  password: "",
+  company: "",
```
patzick, line 202: **minor — `company` placement diverges from both shipped forms.** The example puts `company` at the top level of the body (which matches the `business` branch of the generated type), while `vue-starter-template` binds `state.billingAddress.company` and `vue-demo-store` binds `state.billingAddress.company` too. These are two different fields (customer vs. address). Since the recipe's job is to explain what Shopware does with the body, one sentence on which one the register route actually consumes — and why the templates use the address field — would prevent readers from getting a different result than the templates they copy from.

apps/docs/src/frontends-recipes/account/register.md
```diff
@@ -0,0 +1,493 @@
...
+| Refresh session context          | `refreshSessionContext()`                                                                                        | `GET /context`                                              | <SchemaTypeTooltip type-key='operations["readContext get /context"]["response"]' />                                                                   |
+| Confirm a double opt-in link     | `apiClient.invoke("registerConfirm post /account/register-confirm")`                                             | `POST /account/register-confirm`                            | <SchemaTypeTooltip type-key='operations["registerConfirm post /account/register-confirm"]["body"]' />                                                 |
+
+## Composables
+
+- `useUser`: exposes `register`, which returns the created `Schemas["Customer"]`, plus `user`, `isLoggedIn`, `isCustomerSession`, `isGuestSession`, and `refreshUser` for the state you render afterwards.
+- `useSalutations`: exposes `getSalutations` and `fetchSalutations`. It fetches the list on mount when nothing has been provided yet and shares it through the `swSalutations` injection, so several forms on one page issue one request.
```
patzick, line 143: **nit — the option lists are client-only.** Both `useSalutations` and `useCountries` load via `onMounted` (`useSalutations.ts:52`, `useCountries.ts:81`), so `getSalutations` and `getCountriesOptions` are empty during SSR and the selects render with only the placeholder option on the server. Edge Cases covers SSR for `storefrontUrl` but not for these; a line here (or an explicit `fetchSalutations()`/`fetchCountries()` call in the example) would close the gap.

patzick (changes requested):
## 🔍 Code Review

Docs-only PR. Structure, frontmatter and sidebar match the existing `login`/`wishlist` recipes, all `SchemaTypeTooltip` type keys resolve against `storeApiSchema.json`, all `Related Links` targets exist, and the `useUser().register()` behaviour described (storefrontUrl injection, `active && !doubleOptInRegistration` gate, awaited `refreshSessionContext()`, no `refreshCart()`, `isLoggedIn` formula, rethrow on context failure) is accurate against `packages/composables`.

Six findings, all inline. One is a wrong technical claim (`window.location.origin` during SSR), the rest are accuracy/completeness gaps in a page whose value is being exactly right.

**Note on CI:** every check on this PR is green, but `code-check.yml` has `paths-ignore: ["apps/**", "**.md"]`, so Build, Typecheck, lint and tests did **not** run on this diff — the passing `lint` check belongs to the *Check PR title* workflow. The new 493-line Vue-in-markdown page has therefore never been compiled by CI; please run the docs build locally before marking ready.

patzick:
[Automation] PR looks good

patzick (approved):

PR #2693, opened 2026-09-04
feat(nuxt-module): implement apiClientConfig.timeout

### LINKED ISSUE #2690: Implement apiClientConfig.timeout, which is declared but read by nothing
Part of #2634

## State of play

There are two `apiClientConfig` surfaces in `@shopware/nuxt-module`. They overlap in name only: different location, different shape, different fate.

| surface | declared in | shape | read by | works |
| --- | --- | --- | --- | --- |
| `runtimeConfig.apiClientConfig` and `runtimeConfig.public.apiClientConfig` | `types/runtime-config.d.ts:9-13` | `{ headers }` | `plugin.ts:107-118` | yes |
| `shopware.apiClientConfig` (module option) | `src/index.ts:146-148` | `{ timeout }` | nothing | **no** |

So `headers` is declared where it is read, and works. `vue-demo-store/nuxt.config.ts:22` and `:33` show both variants in commented examples, and both are the working top-level location.

`timeout` is the opposite. It is declared as a module option on `ShopwareNuxtOptions`, `setup()` merges module options only into `runtimeConfig.shopware` and `runtimeConfig.public.shopware` (`src/index.ts:29-36`), and `plugin.ts` never looks there. Nothing in this repo sets it. Nothing anywhere reads it.

## The actual defect

`ShopwareNuxtOptions.apiClientConfig.timeout` is a published type that promises a feature no code implements. Someone setting it gets no timeout, no warning, and no way to tell short of instrumenting the network. That is the whole bug. It is smaller than this ticket originally claimed.

## What to build

Implement the option rather than retract it. The capability is missing, not wrong.

**1. Move `timeout` to the surface that already works.** `types/runtime-config.d.ts:9-13` declares `ApiClientConfig` with only `headers`. Add `timeout` next to it, so the type sits where `plugin.ts` actually reads.

**2. Pass it through.** `createAPIClient` accepts `fetchOptions` and spreads it into `ofetch.create`, so once the value is read there is nothing else to build.

**3. Decide what to do with the module-option declaration.** `ShopwareNuxtOptions.apiClientConfig` (`src/index.ts:146-148`) currently declares `timeout` in a location nothing reads. Either make `plugin.ts` honour the `shopware` scope as well, or drop that declaration now that the working surface carries the option. Leaving both as they are is the one outcome to avoid, because it keeps a public type that lies.

**4. Ship it unset.** Do not give it a default, and do not set it in any template.

That last point matters. Setting it is a separate decision from offering it, and the measurements below argue against the specific value we tried, not against the feature.

## What the measurements do and do not say

A 10s timeout was tried on this branch and reverted. Across a full CI run of 62 failed calls, none ran longer than 8s, so it never fired once. The two runs after reverting were about three minutes faster with under half the flaky count.

What that supports: **10s does nothing against connection failures that die in ~170ms**, and may cut off calls that would have completed.

What it does not support: that a timeout is worthless. Requests that hang were observed separately, and they are exactly what a timeout is for. Two runs against a backend this noisy is also thin evidence for a permanent decision.

So: build the option, document that it is a guard against hangs rather than a fix for connection failures, and leave the choice of value to whoever sets it.

## Retry is deliberately not included

`createAPIClient`'s `fetchOptions` also accepts `retry`, `retryDelay` and `retryStatusCodes`. Exposing those needs its own decision, because ofetch applies a numeric `retry` to every method. A blanket setting would also retry `checkout/order`, `customer/wishlist/add` and `account/register`. A duplicated order is a worse outcome than anything this ticket fixes.

Note ofetch already retries GETs once by default on a network error, and never retries POSTs. Related: #2691.

## Prior art, and what not to copy

The wiring attempt lives in `ff0d4207` and was reverted by `da97727b`:

```
git show ff0d4207 -- packages/nuxt-module/plugin.ts
```

It contains two separable things. The part that reads the config from both the top level and the `shopware` scope is sound and reusable if Option B is ever chosen. The part that sets a timeout is the part the measurements argue against. Do not lift it wholesale.

## Verification

```
pnpm --filter @shopware/nuxt-module typecheck
pnpm --filter @shopware/nuxt-module test
pnpm --filter @shopware/nuxt-module lint
```

`src/index.test.ts` covers module setup. There is no test for `plugin.ts` at all, so if Option B is chosen a small test asserting the option actually reaches `createAPIClient` would be the first one, and worth adding.

To confirm a value reaches the running app, build a template and read it back from the served payload:

```
pnpm --filter vue-starter-template build
pnpm --filter vue-starter-template preview
curl -s http://localhost:3000/ | grep -o 'apiClientConfig:{[^}]*}'
```

## Changeset

`@shopware/nuxt-module` is published, currently 1.5.1. This needs a changeset. `patch` for Option A, since it only removes a declaration that never functioned.

## Out of scope

#2682 is the underlying connection problem. Neither option here touches it.

mkucmus:
### Description

closes #2690

`apiClientConfig.timeout` was declared on `ShopwareNuxtOptions` but read by nothing. Anyone who set it got no timeout and no warning.

`timeout` now sits on `ApiClientConfig` next to `headers`, the surface `plugin.ts` already reads, and is forwarded to `createAPIClient` as `fetchOptions.timeout`. Resolution order, highest first:

```
runtimeConfig.apiClientConfig.timeout          (SSR only)
runtimeConfig.public.apiClientConfig.timeout
shopware.apiClientConfig.timeout               (SSR only, deprecated)
public shopware.apiClientConfig.timeout        (deprecated)
```

Each level is validated on its own before falling through, so an invalid high-priority value never blocks a valid lower one. A finite positive number is valid, and a numeric string is coerced. `0`, negatives, `Infinity`, `NaN`, non-numeric strings, booleans and objects are ignored and logged once as a warning naming the config path the value came from.

`apiClientConfig` under the `shopware` module options is kept, marked `@deprecated`, and read last. Removing it would break type-checking for anyone who set it, and this is not a major release. Move to `runtimeConfig.apiClientConfig`; the fallback goes away in the next major.

Shipped unset. No default, and no template sets it. Retry is not exposed: `fetchOptions` is built literally as `{ timeout }`, so `retry`, `retryDelay` and `retryStatusCodes` stay unreachable.

### Type of change

### ToDo's

- [x] Related Issue updated

### Additional context

The changeset is `minor`, not the `patch` the issue suggested. The issue's note assumed the change only drops a dead declaration, but it also adds an option that works at a path that did not exist before. New options ship as minor here: `cacheableReads` in 1.5.0, `browserLocale` in 1.4.0.

`plugin.ts` had no test at all. It has one now, which needed a vitest `define` shim so `import.meta.server` can be flipped per test. Behavioural tests were added in `packages/api-client` for the timeout error shape and for request-level `signal`.

Worth knowing before enabling it: a timed-out `GET` consumes its automatic retry without re-reaching the server, and the resulting error carries no HTTP status, so `instanceof ApiClientError` does not match it. Use `isTimeoutError()` from `@shopware/api-client` to detect one. Both are covered by tests.

#2702 has merged. This branch now uses `isTimeoutError()` in both checkout pages, drops the two api-client tests it superseded, and documents that a per-call `signal` combines with the timeout instead of replacing it.


.changeset/olive-pandas-wait.md
```diff
@@ -0,0 +1,7 @@
+---
+"@shopware/nuxt-module": minor
```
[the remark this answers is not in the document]
mkucmus: Thanks. Keeping this as minor.

The issue says patch for dropping the dead declaration, and that part on its
own would be a patch. But this change also adds an option that works.
`runtimeConfig.apiClientConfig.timeout` is a new path. Before this, `timeout`
was only documented under the `shopware` module option, where nothing read it.

packages/nuxt-module/README.md
```diff
@@ -114,6 +114,33 @@ More about Nuxt configuration can be found [HERE](https://nuxt.com/docs/getting-
...
+}
+```
+
+Only a positive number arms it. Anything else counts as unset.
+
+Once set:
+
+- The error carries no HTTP status. Read `error.cause.name`, which is `TimeoutError`.
+- A timed-out `GET` uses up its one automatic retry without reaching the server again.
+- Your own `signal` on a single call disables the timeout for that call.
```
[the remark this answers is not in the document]
mkucmus: Thanks, but this does not apply to the pinned version.

`AbortSignal.any` landed in ofetch 2.0.0-alpha.3. The pinned 1.5.1 only arms the timeout when there is no signal:

    if (!context.options.signal && context.options.timeout) {

So a per-request signal does disable the timeout. I checked against a hanging server through @shopware/api-client: with a per-call signal the request was still pending after 1200ms, without it it aborted after 252ms.

packages/nuxt-module/src/index.ts
```diff
@@ -143,6 +143,14 @@ export type ShopwareNuxtOptions = {
...
+  /**
+   * Read last, after `runtimeConfig.apiClientConfig` and
+   * `runtimeConfig.public.apiClientConfig`. Only a positive number is used.
+   *
+   * @deprecated Configure the API client through `runtimeConfig.apiClientConfig`
+   * or `runtimeConfig.public.apiClientConfig` instead. This compatibility
+   * fallback will be removed in the next major.
+   */
   apiClientConfig?: {
     timeout?: number | string;
```
mdanilowicz: `Only a positive number is used.` -> `number | string;` something is not right here

.changeset/olive-pandas-wait.md
```diff
@@ -0,0 +1,7 @@
+---
```
mdanilowicz, line 1: # Why `apiClientConfig.timeout` shouldn't ship as-is

The gap is real — there was no way to set a global Store API timeout, since the 90 `invoke()` calls in `packages/composables` never pass `fetchOptions`. The plumbing in `plugin.ts` is correct. The problem is what it arms.

**1. It makes checkout fail silently.** `errorInterceptor` only runs on `onResponseError`, which needs an HTTP response. A timeout has none, so callers get a bare `FetchError` (verified: `status=undefined`, `cause.name=TimeoutError`), and every consumer branches on `instanceof ApiClientError` with no `else` — `checkout/index.vue:238`, `SwProductCard.vue:110`, `SwContactForm.vue:152`, `SwNewsletterForm.vue:163`, `SwProductReviewsForm.vue:127`. No `vue:error`/`app:error` hook exists in either template. Today a slow order completes; with `timeout: 10000` it aborts, the spinner stops, nothing is shown, the button re-arms — and the server may already have committed it. That trades "slow" for "silently wrong" on the one transaction that matters. The starter template is worse: `handlePlaceOrder` has `try`/`finally` with no `catch` at all.

**2. The guard doesn't cover the hang it advertises.** ofetch clears the timer once headers arrive (`ofetch.CWycOUEr.mjs:287-291`), before the body is read. Reproduced: a server sending headers + partial body then stalling leaves a `timeout: 500` request **still pending after 5001ms**. A CDN holding a chunked response open — the sold scenario — is not covered. Our own templates also defeat it by passing their own `AbortSignal` (`FrontendNavigationPage.vue:69`), which ofetch treats as opting out.

**3. A dormant option goes live under `minor`.** The changeset says `shopware.apiClientConfig` "had never been read before". It is now, on both server and client. An app that set `timeout: 1000` years ago starts aborting every request after a same-major update, with no warning.

**4. The fix belongs one package down.** `api-client` has had `fetchOptions.timeout` since 2024-12; what it never grew is the error contract. `isTimeoutError()` belongs next to `ApiClientError`, and merging a caller signal with the configured timeout can only happen in `createAPIClient`. The README telling integrators to read `error.cause.name` by hand is a workaround for that missing export.

**Also:** the precedence order the changeset headlines has no test — swapping `plugin.ts:121`/`:122` leaves the suite green (54 passed, 0 failed).

## Recommendation

Land `isTimeoutError()` and signal-merging in `api-client` first, then this branch plus the missing `else` branches. If it must go now, two edits turn "silently broken" into "visibly failed": an `else` at `checkout/index.vue:238` and a `catch` at `vue-starter-template/.../checkout/index.vue:56`.

mdanilowicz: I think the problems mentioned above are worth considering.

mdanilowicz (changes requested):

packages/nuxt-module/src/index.ts
mkucmus: Done. The deprecated `timeout` is now typed as `number`, which is what the plugin reads.

.changeset/olive-pandas-wait.md
mkucmus: Thanks, this was a thorough look and most of it is now in the branch.

Done:
- Both checkout pages show a generic error for non-API failures. The demo store got an `else`, the starter a `catch` through `handleApiError`.
- Precedence is tested across all four sources, including an invalid higher value falling through to a valid lower one. Swapping any two fallback lines now fails a test.
- The README and changeset say the timeout covers the wait for response headers only. A new api-client test pins that a stalled body is not aborted.
- `shopware.apiClientConfig` logs a build-time deprecation warning, and the changeset says a value set there in the past becomes active.

Not in this PR:
- `isTimeoutError()` and merging a caller signal with the configured timeout. Both are api-client changes with their own changeset, so I would do them in a follow-up PR and link it here.
- The other `instanceof ApiClientError` sites. That pattern is template-wide, 35 places, and deserves its own change.

mkucmus:
@mdanilowicz following your recommendation, the api-client part goes first.

`isTimeoutError()` and combining a caller `signal` with the configured timeout are ready on `feat/api-client-timeout-signal` and will open as a separate PR. nuxt-module and the templates consume api-client, so that order keeps each change in one package.

Merge order:

1. The api-client PR merges.
2. This PR merges main, drops the test that asserts the old signal behaviour, fixes the two README bullets, and can use `isTimeoutError()` in the checkout branches.
3. Re-review here.

Until then this PR stays as is, so its CI does not go red before the api-client change is in. I will link the PR here once it is open.

.changeset/olive-pandas-wait.md
mkucmus: The follow-up is open: #2702 adds `isTimeoutError()` and combines a caller `signal` with the timeout in both clients. It should merge first, then this PR gets its three updates.

packages/nuxt-module/plugin.ts
```diff
@@ -55,6 +55,12 @@ function getApiErrors(data: unknown): ApiError[] {
   return Array.isArray(errors) ? errors.filter(isApiError) : [];
 }
 
+function toTimeout(value: unknown): number | undefined {
+  return typeof value === "number" && Number.isFinite(value) && value > 0
+    ? value
+    : undefined;
```
patzick: **Major — a rejected value produces no diagnostic, which is the bug this PR set out to fix.**

The PR description opens with "anyone who set it got no timeout and no warning". This keeps that property for every value `toTimeout` rejects: `"5000"`, `0`, `-1`, `NaN`, `Infinity`, `true`. The value is dropped and nothing anywhere says so — same silent no-op, new path.

It bites hardest on `"5000"`, which the previous `timeout?: number | string` type explicitly allowed.

Distinguish "unset" from "set but invalid":

```ts
function toTimeout(value: unknown, source: string): number | undefined {
  if (value === undefined || value === null) return undefined;
  if (typeof value === "number" && Number.isFinite(value) && value > 0) return value;
  console.warn(
    `[shopware] Ignoring ${source}.timeout: expected a positive number of milliseconds, got ${JSON.stringify(value)}.`,
  );
  return undefined;
}
```

packages/nuxt-module/src/index.ts
```diff
@@ -143,8 +151,16 @@ export type ShopwareNuxtOptions = {
...
+   * Read last, after `runtimeConfig.apiClientConfig` and
+   * `runtimeConfig.public.apiClientConfig`. Only a positive number is used.
+   *
+   * @deprecated Configure the API client through `runtimeConfig.apiClientConfig`
+   * or `runtimeConfig.public.apiClientConfig` instead. This compatibility
+   * fallback will be removed in the next major.
+   */
   apiClientConfig?: {
-    timeout?: number | string;
+    timeout?: number;
```
patzick: **Major — narrowing `number | string` to `number` is a type-level break in a `minor`.**

The changeset says "non-breaking change which adds functionality", but `timeout: "5000"` type-checked on the previous release and now fails. Combined with the silent runtime rejection in `plugin.ts`, someone who wrote a string gets both a new type error and a dead option.

Since this surface is deprecated and read-last anyway, keeping the wider type and coercing costs nothing:

```ts
apiClientConfig?: {
  /** @deprecated pass a number; strings are coerced for compatibility */
  timeout?: number | string;
};
```

Otherwise the changeset should call the narrowing out.

templates/vue-demo-store/app/pages/checkout/index.vue
```diff
@@ -236,12 +236,15 @@ const placeOrder = async () => {
...
   } catch (error) {
-    if (error instanceof ApiClientError)
+    if (error instanceof ApiClientError) {
       for (const errorItem of error.details.errors) {
         if (errorItem?.detail) {
           pushError(errorItem.detail);
         }
       }
+    } else {
+      pushError(t("errors.message-default"));
```
patzick, line 252: **Major — right direction, but this is the one error where "something went wrong" is actively misleading.**

Adding the `else` fixes the silent failure, so this is a real improvement. The residual problem is what the user does next: `errors.message-default` reads "Unfortunately, something went wrong.", `isLoading.placeOrder` resets on line 249, and the Place Order button re-arms. A timeout is exactly the case where the server may have committed the order — so the most natural reaction to this message is the one that risks a second order.

A timeout-specific branch would tell the truth:

```ts
} else if ((error as { cause?: Error })?.cause?.name === "TimeoutError") {
  pushError(t("errors.order-timeout")); // "we didn't get a confirmation in time — check your orders before retrying"
} else {
  pushError(t("errors.message-default"));
}
```

(#2702's `isTimeoutError()` is the clean version of that check, which is another reason to let it land first.)

templates/vue-starter-template/app/pages/checkout/index.vue
```diff
@@ -57,6 +58,8 @@ async function handlePlaceOrder() {
     const order = await createOrder();
     await push(formatLink(`/checkout/success/${order.id}`));
     refreshCart();
+  } catch (error) {
+    handleApiError(error);
```
patzick: Same as the demo-store page: `handleApiError` falls through to `errors.message-default` for a timeout, and `isPlacingOrder` resets on line 64, so the button re-arms with the order possibly already placed. Whatever wording you settle on there, please mirror it here.

packages/api-client/src/createApiClient.test.ts
```diff
@@ -706,6 +706,134 @@ describe("createAPIClient", () => {
...
+        // @ts-expect-error this endpoint does not exist
+        .invoke("testTimeoutShape get /slow-endpoint", {})
+        .catch((caught: unknown) => caught as Error & { status?: number });
+
+      expect(error.name).toBe("FetchError");
+      expect(error.status).toBeUndefined();
+      expect((error.cause as Error).name).toBe("TimeoutError");
+    });
+
+    it("lets a per-request signal replace the client timeout, so a slow request still resolves", async () => {
```
patzick: **Major — this test is written to fail once #2702 merges.**

It pins ofetch 1.5.1's "a per-request signal replaces the client timeout" behaviour, which #2702 exists to remove by combining the two. Nothing in either PR enforces the ordering you described in the thread, so whichever lands first breaks the other: #2702 first → this test goes red; this PR first → `main` documents behaviour that's about to be untrue.

Cleanest resolution is to merge #2702, then rebase this branch and drop this test with the README bullets. If this has to go first, at least make the coupling explicit in the test name/comment so the failure is self-explaining rather than mysterious.

packages/nuxt-module/README.md
```diff
@@ -114,6 +114,35 @@ More about Nuxt configuration can be found [HERE](https://nuxt.com/docs/getting-
...
+
+Only a positive number arms it. Anything else counts as unset.
+
+`shopware: { apiClientConfig: { timeout } }` still works as a deprecated fallback, read only when neither `runtimeConfig` path is set. Nuxt warns at build time when it is set. Move it to `runtimeConfig.apiClientConfig`; it goes away in the next major.
+
+Once set:
+
+- The error carries no HTTP status. Read `error.cause.name`, which is `TimeoutError`.
+- A timed-out `GET` uses up its one automatic retry without reaching the server again.
+- Your own `signal` on a single call disables the timeout for that call.
```
patzick: Tied to the test above — #2702 makes a caller `signal` combine with the timeout rather than disable it, so this bullet has a short shelf life. Worth a note that it applies to the currently pinned ofetch, or hold it until the ordering is settled.

packages/nuxt-module/plugin.ts
```diff
@@ -101,11 +107,20 @@ function setupShopwarePlugin(NuxtApp: ShopwarePluginNuxtApp): {
...
     ? (runtimeConfig.apiClientConfig as ApiClientConfig)
     : undefined;
+  const publicApiClientConfig = runtimeConfig.public
+    ?.apiClientConfig as ApiClientConfig;
+
+  const timeout =
+    toTimeout(privateApiClientConfig?.timeout) ??
+    toTimeout(publicApiClientConfig?.timeout) ??
+    toTimeout(shopwareRuntimeConfig?.apiClientConfig?.timeout) ??
+    toTimeout(shopwareRuntimeConfigPublic?.apiClientConfig?.timeout);
```
patzick: **Minor — the last two tiers can't differ in practice.**

`src/index.ts:29-35` `defu`s module `options` into both `runtimeConfig.shopware` and `runtimeConfig.public.shopware`, so `shopware: { apiClientConfig: { timeout } }` always produces the same value on both lines 122 and 123. They diverge only if someone hand-writes different values into `runtimeConfig.shopware` and `runtimeConfig.public.shopware` directly — which the `precedence` test does, but no real config would.

Not wrong, just one more level than the config can express. Collapsing to a single deprecated read would make the chain honest; if you keep both, a comment explaining when they differ would help.

apps/e2e-tests/.gitignore
```diff
@@ -1,3 +1,4 @@
 playwright-report
 test-results
-reports
\ No newline at end of file
+reports
+diagnostics/
```
patzick: Nit — nothing in the repo writes `diagnostics/`; this looks like leftover cleanup from the artifact that was removed earlier in the branch. Unrelated to the timeout API, so it'd read better as its own commit or dropped.

patzick (changes requested):
# 🔍 Code Review

The plumbing is right: `toTimeout` → `fetchOptions.timeout`, four-level precedence, tests that actually fail when the fallback order is swapped, and both checkout pages no longer swallow a non-`ApiClientError`. CI is fully green (11/11).

Four things I'd like resolved before merge.

## 💥 Blockers / Majors

1. **An invalid `timeout` is still silently ignored** — `plugin.ts:58`. The PR's own premise is "anyone who set it got no timeout and no warning". `toTimeout` reproduces exactly that for `"5000"`, `0`, `-1`, `NaN`, `Infinity`: rejected, no diagnostic anywhere. Warn when the key is present but rejected.

2. **The type narrowing is a breaking change under a `minor`** — `src/index.ts:163`. `timeout?: number | string` → `timeout?: number`. `"5000"` was type-legal before; those users get a type error on a same-major bump, and per (1) no runtime signal either.

3. **Checkout now shows a generic error where the order may already exist** — `vue-demo-store/.../checkout/index.vue:246`, `vue-starter-template/.../checkout/index.vue:62`. Visible beats silent, so this is the right direction — but "something went wrong" plus a re-armed button is an invitation to press it again, and a timeout is precisely the case where the server may have committed. A timeout-specific message would close the gap.

4. **Merge-order coupling isn't encoded anywhere** — `createApiClient.test.ts:737`, `README.md:143`. Both assert the ofetch 1.5.1 signal-replaces-timeout behaviour that #2702 deliberately removes. Whichever lands first, the other is wrong. Your plan in the thread is fine; I'd just rather see #2702 merged and this branch updated than the ordering living only in a comment.

## 🧩 Minors / Nits

- `plugin.ts:122-123` — the two deprecated tiers can never differ in practice: `src/index.ts:29-35` `defu`s module `options` into both `runtimeConfig.shopware` and `runtimeConfig.public.shopware`, so `shopware.apiClientConfig.timeout` always lands in both. The precedence test builds them independently, which no real `nuxt.config` can produce.
- `plugin.ts:109` — the local `ApiClientConfig` (`timeout?: unknown`) duplicates the augmented interface in `types/runtime-config.d.ts` (`timeout?: number`). Two declarations that can drift.
- `src/index.ts:50` — the warning fires on the mere presence of the object, so `apiClientConfig: {}` warns about moving a `timeout` that was never set.
- `apps/e2e-tests/.gitignore:4` — unrelated to the timeout API.

## ✅ Verified

- `errors.message-default` exists in all three locales of both templates.
- `useApiErrorsResolver().handleApiError` has the non-`ApiClientError` branch this relies on.
- The `vitest` `define` shim is behaviour-neutral for the existing suite — `import.meta.server` was already `undefined` under plain vitest.

**Verdict: changes requested** — (1) and (4) are the ones that matter; (2) and (3) are quick.

patzick (approved):
# ✅ Recheck on `f1e1829` — all eight findings addressed

CI green on the current head (10 check-runs + 3 statuses, all success). Superseding my earlier `CHANGES_REQUESTED`.

| # | Finding | Resolution |
|---|---|---|
| 1 | Rejected `timeout` silently dropped | `toTimeout` now takes a `source`, coerces numeric strings, and warns once per distinct message naming the config path — `plugin.ts:58-80` |
| 2 | `number \| string` → `number` type break | Reverted to `number \| string` on both tiers via the shared `ApiClientRuntimeConfig` — `src/index.ts:144-149` |
| 3 | Generic checkout error after a possibly-committed order | Dedicated `errors.order-timeout` ("your order may still have been placed…") in both templates, all three locales |
| 4 | Test/README coupled to unmerged #2702 | Test renamed `ofetch 1.5.1 only: … (deleted by #2702)`, README bullet now scopes the claim and links the PR |
| 5 | Dead precedence tier | Comment added; both deprecated tiers report as `shopware.apiClientConfig`, the name `nuxt.config` actually uses |
| 6 | Duplicated `ApiClientConfig` declaration | Collapsed — `runtime-config.d.ts` is now `interface ApiClientConfig extends ApiClientRuntimeConfig {}` |
| 7 | Warning fired on an empty object | Guarded on `?.timeout !== undefined` — `src/index.ts:50-51` |
| 8 | Out-of-scope `.gitignore` entry | Reverted cleanly, byte-for-byte |

## What I re-verified

- `undefined`/`null` short-circuit before the warn path, so "unset" stays silent — tested explicitly.
- The dedupe `Set` is module-scoped, and `vi.resetModules()` in `beforeEach` gives each test a fresh one, so "warns once for a repeated bad value" is testing the real thing rather than leaked state.
- The precedence fixtures were updated for string coercion — the old `"2000"` invalid-value cases became `"2000ms"`, which is still rejected. Easy thing to miss.
- `errors.order-timeout` is present in de-DE, en-GB and pl-PL for both templates.

## Non-blocking

The `isTimeoutError` stopgap is duplicated in both checkout pages. Both carry the `// Replace with isTimeoutError() from @shopware/api-client once #2702 lands.` comment, so it's traceable — just two places to remember.

Accepting string coercion on the *new* `runtimeConfig.apiClientConfig` path (not just the deprecated one) is looser than I'd have picked for a fresh option, since Nuxt's `destr` already turns `NUXT_*` env values into numbers. It's documented in both the README and the changeset, so it's a deliberate call and fine by me.

@mdanilowicz's earlier request for changes is still open and is theirs to clear — the two api-client items from their review are tracked in #2702.

PR #2700, opened 2026-09-07
feat(vue-starter-template): checkout flow

### LINKED ISSUE #2699: Checkout: "Create customer account" silently registers a guest instead
## Problem description and current behaviour

On the starter template's checkout page, step 1 offers a "Create customer account" link. Clicking it reveals a password field. Filling that field in and pressing "Next" completes the step — but the shopper is registered as a **guest**, and the password they chose is discarded. No account is created, no error is shown, and nothing tells them it did not happen. They only find out later, when they try to log in with credentials that do not exist.

There are two independent causes.

**1. The toggle never reaches the code that submits.** `createAccountToggle` is component-local state in `CustomerBaseInfo.vue` and is never emitted:

```ts
// templates/vue-starter-template/app/components/checkout/CustomerBaseInfo.vue:22
const createAccountToggle = ref(false);
```

The page therefore has no way to read it and hardcodes the guest flag:

```ts
// templates/vue-starter-template/app/pages/checkout/index.vue:122-123
      password: customerBaseInfo.value.password,
      guest: true,
```

Per the Store API contract, `guest: true` is exactly what discards the password — `packages/api-client/api-types/storeApiTypes.d.ts:9559`: *"If set, will create a guest customer. Guest customers can re-use an email address and don't need a password."*

**2. The toggle is a delayed latch, not a switch.** Even once it is wired up, the flag is not set by the click but by an animation timer 600 ms later, while the password field appears immediately:

```ts
// CustomerBaseInfo.vue:24-31
function switchAnimation(e: Event) {
  e.preventDefault();
  switchAnimating.value = true;
  setTimeout(() => {
    createAccountToggle.value = true;
    switchAnimating.value = false;
  }, 600);
}
```

The field's visibility is driven by `switchAnimating`, not by the flag (`CustomerBaseInfo.vue:69`), so for 600 ms the UI says "account mode" while the state still says "guest". A naive fix that only hoists the existing ref inherits this race: a shopper who types fast and submits inside that window still gets `guest: true`. There is also no `clearTimeout`/`onUnmounted`, so the timer fires after the component unmounts (the checkout content sits behind a real `v-if`/`v-else` on `isEmpty`, so emptying the cart destroys it).

## Expected Behavior

- Clicking "Create customer account" puts the form into account mode **synchronously**; the animation is presentation only.
- Submitting in account mode sends `guest: false` and the chosen password, and creates a real customer account.
- In account mode the password is required and validated before submit; in guest mode it is not required and no password error is shown.
- A shopper cannot end up with a guest record after explicitly asking for an account — either the account is created, or a visible error explains why not.

## Steps To Reproduce

1. Run `templates/vue-starter-template` against any Shopware instance.
2. Add a product to the cart and go to `/checkout`.
3. In step 1, fill in a valid e-mail and a complete billing address.
4. Click **Create customer account** and enter a valid password (8+ characters).
5. Press **Next**. The step completes.
6. Open `POST /store-api/account/register` in the network tab — the body contains `"guest": true`.
7. Log out and try to log in with the e-mail and password from step 4 — the credentials do not exist.

Second, narrower repro for cause 2 (only observable once cause 1 is fixed by hoisting the flag as-is): perform step 4 and press **Next** within 600 ms of clicking the toggle — the request still carries `"guest": true`.

mdanilowicz:
closes #2699

This pull request improves the registration and checkout experience by ensuring the cart is refreshed after user registration, aligning it with login and logout behaviors. It also enhances accessibility and error handling in the checkout flow, and updates button labeling for better user feedback during guest registration. Below are the most important changes:

**Cart Refresh and Registration Logic:**

* The `useUser().register()` method now calls `refreshCart()` after updating the session context, ensuring cart totals reflect the correct tax rates, shipping surcharges, and customer-group prices immediately after registration. [[1]](diffhunk://#diff-6eec8d66e0e5c3a74b2cbe7e500dbb207fbe6c5a218f689ad5ba11d7bbdfd88fR1-R7) [[2]](diffhunk://#diff-f6055843be7b7f31df7f2cd21efca6578d79ef68a42f71d128de3688761caabaR196)
* Corresponding unit tests updated to verify both `refreshSessionContext` and `refreshCart` are called after registration.

**Checkout Flow and Error Handling:**

* The guest registration step in the checkout process was refactored: form validation is more granular (separating email and billing address), error handling is improved with persistent error notifications, and focus management ensures accessibility by focusing the first invalid field or restoring focus after errors. [[1]](diffhunk://#diff-2980359c068008e2d662439c3957394f583df78c7a7854a417793f235ffd784eR26-R32) [[2]](diffhunk://#diff-2980359c068008e2d662439c3957394f583df78c7a7854a417793f235ffd784eR57-R87) [[3]](diffhunk://#diff-2980359c068008e2d662439c3957394f583df78c7a7854a417793f235ffd784eL73-R117) [[4]](diffhunk://#diff-2980359c068008e2d662439c3957394f583df78c7a7854a417793f235ffd784eR136-R141)
* The guest registration button now shows a loading state and dynamic label while registration is processing, providing better feedback to users. [[1]](diffhunk://#diff-2980359c068008e2d662439c3957394f583df78c7a7854a417793f235ffd784eR198) [[2]](diffhunk://#diff-2980359c068008e2d662439c3957394f583df78c7a7854a417793f235ffd784eL170-R221) [[3]](diffhunk://#diff-82012d4ff1e074fd28272afeb1a71c79f408e7cb6baae4bd67367df220caafabL31-R32) [[4]](diffhunk://#diff-5a553d21e1a3acebf5ca93b4ea7eae86540d774c69d00c2f151287c226036d35L31-R32) [[5]](diffhunk://#diff-8027c1db832867011c19ffafe1327ca827de24dc55a9b51cbcd71b94b0c13233L31-R32)

**Accessibility Improvements:**

* The `InputField.vue` component now associates error messages with inputs using `aria-describedby` and `aria-invalid`, improving screen reader support for form validation errors. [[1]](diffhunk://#diff-8b38d9b1ee9ec0d718baf6dec94a3205dc461417c624e64d79809cff0a7f49b2R29-R31) [[2]](diffhunk://#diff-8b38d9b1ee9ec0d718baf6dec94a3205dc461417c624e64d79809cff0a7f49b2R59-R68)

**UI and Notification Enhancements:**

* The checkout layout now includes a `LayoutNotifications` component to display notifications to users.

**Minor Fixes:**

* Conditional rendering in the `CustomerAddressChosen.vue` component ensures the country name is only displayed if available.
* Code cleanup in `useApiErrorsResolver.ts` removes an unused variable.


templates/vue-starter-template/app/pages/checkout/index.vue
```diff
@@ -70,35 +98,47 @@ function handleChangePaymentMethod(id: string) {
...
-      street: billingAddress.value.street,
-      zipcode: billingAddress.value.zipcode,
-      city: billingAddress.value.city,
-      countryId: billingAddress.value.countryId,
-    },
-    acceptedDataProtection: true,
-  });
+      email: customerBaseInfo.value.email,
+      password: customerBaseInfo.value.password,
+      guest: true,
```
patzick: This still submits every checkout registration as `guest: true`. The PR closes #2699, whose expected behavior is `guest: false` plus password validation when “Create customer account” is selected; because the child toggle is not exposed to this page, that path still creates a guest and discards the password.

templates/vue-starter-template/app/components/form/InputField.vue
```diff
@@ -53,10 +56,15 @@ defineExpose<FocusableInput>({
...
     />
-    <span v-if="errorText" class="text-states-error text-xs block mt-1">{{
-      errorText
-    }}</span>
+    <span
+      v-if="errorText"
+      :id="errorId"
+      class="text-states-error text-xs block mt-1"
+      >{{ errorText }}</span
+    >
```
patzick, line 68: This leaves a literal `>` text node after the error message, so validation errors render with an extra `>` in the form. Close the span normally instead of splitting the closing tag this way.

patzick (changes requested):
Changes requested. CI checks reported by GitHub are passing; I did not run local checks.

templates/vue-starter-template/app/components/form/InputField.vue
mdanilowicz: This is Prettier's whitespace-sensitive tag formatting, not a split closing tag — no > text node is produced.

mdanilowicz: We have 103 tags with that formatting. I will create new PR for that issue

patzick (approved):
Rechecked current head 2b737a57. The previous registration blocker is fixed: checkout now sends `guest: !createAccount.value` and validates the password only in account mode. I am not carrying forward the InputField tag-formatting note; CI accepts it and it is valid tag-closing whitespace. GitHub CI checks are passing; no local checks run.

PR #2702, opened 2026-09-07
feat(api-client): add isTimeoutError and merge signal with timeout

mkucmus:
### Description

Follow-up to the review on #2693.

`isTimeoutError()` is exported next to `ApiClientError`. A request that runs into `fetchOptions.timeout` rejects without an HTTP status, so `instanceof ApiClientError` never matches it. The request may already have reached the API and been processed, so the server-side outcome is unknown and a mutation must not be retried without checking. The helper reads the `TimeoutError` name from the error itself or its `cause`, which is the DOM convention and works across realms.

A per-request `signal` no longer switches off the timeout. ofetch 1.5.1 only arms its timer when no `signal` is set, so `mergeSignalWithTimeout` combines the caller's signal with the client-level or per-request timeout and hands back a disposer. Whichever fires first aborts the request. Both clients use the same module, and the admin client forwards the signal and timeout to its token refresh as well, so a stalled refresh is bounded too.

`AbortSignal.any` needs Chrome 116, Firefox 124, Safari 17.4 or Node 20.3. Older runtimes keep the previous behaviour, where a per-request `signal` wins and the timeout is dropped. A test pins that fallback.

One difference to ofetch's own timer: the combined timeout covers the whole request, including reading the body and a retry, while ofetch's stops once headers arrive. When it fires while the body is being read, ofetch does not wrap it, so the caller gets a `DOMException` named `TimeoutError` instead of a `FetchError`. `isTimeoutError()` recognises both.

image

### Changes after review

The merge block was duplicated in both clients. It now lives once in `mergeSignalWithTimeout.ts` with its own unit tests.

The timer is cleared. `AbortSignal.timeout` gives no handle to cancel it, so the module owns an `AbortController` and a `setTimeout`, and both clients release it in a `finally`. It is armed as the last step before the request, so nothing in between can leave it running.

The timeout is validated before use. It is coerced the way `setTimeout` coerces its delay, rounded up, and capped at 2147483647. A value that is not a finite positive number is ignored instead of throwing a `RangeError` out of `invoke()`.

The timeout is no longer forwarded to ofetch once it has been merged, so exactly one timer exists by construction rather than by relying on an ofetch 1.5.1 internal. It is assigned `undefined` rather than deleted, because `resolveFetchOptions` is a plain spread and `delete` would let the client-level default through.

The README now gives the Node floor, the validation rules, and the two places where behaviour differs by whether a `signal` was passed.

`mergeSignalWithTimeout.ts` carries a `TODO` pointing at unjs/ofetch#508, so the module can be deleted on the ofetch 2.0+ bump.

### Two points for the reviewers

Bounding the admin token refresh with the caller's signal means cancelling a request mid-refresh loses a rotation the server already committed. league/oauth2-server revokes the presented refresh token, so the client keeps a dead one and later calls fail with `ApiClientError` until re-authentication. Reproduced on both the timeout path and the pure-cancellation path. It is documented in the README rather than changed, because altering session state on abort needs a decision.

Without a `signal` the token refresh and the request each get the full `timeout`, so an expired session can take twice as long. Two tests pin both semantics. Making the deadline unconditional would remove the difference, but it changes the error shape on the no-signal path, so it is left out of this PR.

### Type of change

### ToDo's

- [x] Changeset file provided
- [x] Related Issue updated

### Additional context

No separate issue, this comes from the #2693 review. Coverage stays at the enforced 100%, now 187 tests.

Mutation testing found three assertions that could not fail, including on the timer release itself, so those tests were rewritten to actually discriminate.

Still to file: an issue to delete `mergeSignalWithTimeout.ts` on the ofetch 2.0+ bump.

Once this lands, #2693 needs three small updates: its test "lets a per-request signal replace the client timeout" asserts the old behaviour, and two README bullets there, reading `error.cause.name` and a per-call `signal` disabling the timeout, become wrong.


packages/api-client/src/createAdminAPIClient.ts
```diff
@@ -240,6 +241,18 @@ export function createAdminAPIClient<
...
+    const timeout = fetchOptions.timeout ?? clientTimeout;
+    if (
+      fetchOptions.signal &&
+      timeout &&
+      typeof AbortSignal.any === "function"
+    ) {
+      fetchOptions.signal = AbortSignal.any([
+        fetchOptions.signal,
+        AbortSignal.timeout(timeout),
+      ]);
```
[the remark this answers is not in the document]
mkucmus: Done. The refresh call now receives `options.signal` and `options.timeout`, and two tests cover an expired session with a stalled token endpoint, with and without a caller signal.

.changeset/api-client-timeout-signal.md
```diff
@@ -0,0 +1,7 @@
+---
+"@shopware/api-client": minor
+---
+
+Added `isTimeoutError()` to tell a request that ran into `fetchOptions.timeout` apart from other errors. Such a request never reaches the API, so it rejects with a plain `FetchError` without an HTTP status, and `instanceof ApiClientError` does not match it.
```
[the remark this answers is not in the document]
mkucmus: Done. The changeset now says the request may already have reached the API and been processed, and that a mutation must not be retried without checking.

packages/api-client/README.md
```diff
@@ -420,15 +435,19 @@ How the client handles this for you:
 
 Client is throwing `ApiClientError` with detailed information returned from the API. It will display clear message in the console or you can access `details` property to get raw information from the response.
 
+A request that runs into `fetchOptions.timeout` never reaches the API, so it is not an `ApiClientError` and has no HTTP status. Use `isTimeoutError` to tell it apart.
```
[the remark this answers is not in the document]
mkucmus: Done. The README says the server-side outcome is unknown and warns against retrying a mutation without checking.

packages/api-client/src/isTimeoutError.ts
```diff
@@ -0,0 +1,21 @@
...
+    typeof value === "object" &&
+    value !== null &&
+    (value as { name?: unknown }).name === "TimeoutError"
+  );
+}
+
+/**
+ * Whether a request failed because `fetchOptions.timeout` ran out.
+ * Such a request never reaches the API, so it rejects with a plain
+ * `FetchError` without an HTTP status and carries the timeout as its `cause`.
```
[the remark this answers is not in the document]
mkucmus: Done. The JSDoc no longer claims the request never reached the server. It says the rejected error has no HTTP status and names the timeout itself or in its `cause`.

packages/api-client/src/createAPIClient.ts
```diff
@@ -208,6 +209,18 @@ export function createAPIClient<
...
 
+    const timeout = fetchOptions.timeout ?? clientTimeout;
+    if (
+      fetchOptions.signal &&
+      timeout &&
+      typeof AbortSignal.any === "function"
+    ) {
+      fetchOptions.signal = AbortSignal.any([
+        fetchOptions.signal,
+        AbortSignal.timeout(Math.ceil(timeout)),
```
mdanilowicz: timeout value should be validated

>range-checks its argument as uint32. Math.ceil covers the fractional case but not sign or finiteness — verified on Node 24, both Math.ceil(-5) and Math.ceil(Infinity) throw RangeError ERR_OUT_OF_RANGE. A caller computing timeout: deadline - Date.now() gets a synchronous RangeError out of invoke() once the budget is spent; it is not a FetchError, not an ApiClientError, and isTimeoutError() returns false, so it escapes every branch of the error ladder this PR adds to the README. The same input on the no-signal path does not throw, so the error taxonomy depends on whether a signal was passed. Clamp before use: Math.min(Math.ceil(raw), 2_147_483_647) guarded on Number.isFinite(raw) && raw > 0.

mdanilowicz: There is no clear hook for AbortSignal

>AbortSignal.timeout schedules a timer with no handle to clear, and nothing cancels it after the request settles — ofetch's own implementation clears its timer in a finally. Measured on Node 24: 100k merged signals retain 153MB, and 88MB survives dropping every reference plus two forced GCs, because the pending timer pins the signal until it fires. On an SSR server with timeout: 30000 at ~500 req/s that is ~15k dead timers in steady state. Worth owning the controller and clearing it in a finally around the request — and this should land together with the unconditional-merge fix, since merging for every request makes the retention universal rather than signal-only.

packages/api-client/src/createAdminAPIClient.ts
```diff
@@ -240,6 +243,18 @@ export function createAdminAPIClient<
       ...(currentParams.fetchOptions || {}),
     };
 
+    const timeout = fetchOptions.timeout ?? clientTimeout;
```
mdanilowicz: https://github.com/unjs/ofetch/pull/508.  We need ToDo or task to remove code after upgrade to 2.0+

mdanilowicz (changes requested):

packages/api-client/src/createAPIClient.ts
patzick: **Blocker — the timeout timer is never cleared.**

`AbortSignal.timeout(ms)` returns a signal with no handle to cancel its underlying timer. Nothing here cancels it once the request settles, so every request with a caller `signal` leaves a live timer that pins the signal (and, through `AbortSignal.any`, the composite) until it fires. With SSR at `timeout: 30000` this is thousands of dead timers in steady state — @mdanilowicz measured 88MB surviving forced GC over 100k merged signals.

ofetch clears its own timer in a `finally`. Do the same by owning the controller:

```ts
let timeoutId: ReturnType<typeof setTimeout> | undefined;
if (fetchOptions.signal && timeout && typeof AbortSignal.any === "function") {
  const controller = new AbortController();
  timeoutId = setTimeout(
    () => controller.abort(new DOMException("The operation was aborted due to timeout", "TimeoutError")),
    normalizedTimeout,
  );
  fetchOptions.signal = AbortSignal.any([fetchOptions.signal, controller.signal]);
}
try {
  /* apiFetch.raw(...) */
} finally {
  clearTimeout(timeoutId);
}
```

The explicit `DOMException` named `TimeoutError` keeps `isTimeoutError()` matching, so no test changes are needed.

packages/api-client/src/createAPIClient.ts
```diff
@@ -208,6 +209,18 @@ export function createAPIClient<
       ...(currentParams.fetchOptions || {}),
     };
 
+    const timeout = fetchOptions.timeout ?? clientTimeout;
```
patzick: **Blocker — `timeout` is unvalidated before it reaches `AbortSignal.timeout()`.**

That API range-checks its argument as `uint32`, so `Math.ceil` handles the fractional case but not sign or finiteness: `AbortSignal.timeout(-5)` and `AbortSignal.timeout(Infinity)` both throw `RangeError`.

The realistic trigger is `timeout: deadline - Date.now()` — once the budget is spent, `invoke()` throws synchronously with an error that is not a `FetchError`, not an `ApiClientError`, and for which `isTimeoutError()` returns `false`. It escapes every branch of the error ladder this PR documents in the README. The same input on the no-signal path does not throw, so the error taxonomy depends on whether the caller passed a `signal` — the exact inconsistency this PR sets out to remove.

```ts
const rawTimeout = fetchOptions.timeout ?? clientTimeout;
const timeout =
  Number.isFinite(rawTimeout) && (rawTimeout as number) > 0
    ? Math.min(Math.ceil(rawTimeout as number), 2_147_483_647)
    : undefined;
```

Please add a test for a negative and a non-finite timeout on the signal path — line coverage is at 100% but this path is untested.

packages/api-client/src/createAPIClient.ts
```diff
@@ -208,6 +209,18 @@ export function createAPIClient<
...
+    if (
+      fetchOptions.signal &&
+      timeout &&
+      typeof AbortSignal.any === "function"
+    ) {
+      fetchOptions.signal = AbortSignal.any([
+        fetchOptions.signal,
+        AbortSignal.timeout(Math.ceil(timeout)),
+      ]);
+    }
```
patzick: **Major — `timeout` is still forwarded to ofetch alongside the merged signal.**

After the merge, the timeout is enforced entirely by `fetchOptions.signal`, yet `fetchOptions.timeout` is spread into `apiFetch.raw` unchanged. Today that is harmless only because ofetch 1.5.1 skips arming its timer when a `signal` is present — an internal this code depends on without stating it.

If that changes (ofetch 2.x reworks exactly this, per unjs/ofetch#508), two independent timers race and reintroduce the split error shape the PR is fixing: ofetch's timer produces a wrapped `FetchError`, the merged signal produces a bare `DOMException`.

`delete fetchOptions.timeout` after a successful merge, so exactly one timer exists by construction, and the behaviour no longer hinges on the ofetch version.

packages/api-client/src/createAdminAPIClient.ts
```diff
@@ -240,6 +243,18 @@ export function createAdminAPIClient<
...
+    if (
+      fetchOptions.signal &&
+      timeout &&
+      typeof AbortSignal.any === "function"
+    ) {
+      fetchOptions.signal = AbortSignal.any([
+        fetchOptions.signal,
+        AbortSignal.timeout(Math.ceil(timeout)),
```
patzick: **Major — the merge block is duplicated verbatim.**

These 11 lines are byte-identical to `createAPIClient.ts:212-222`, which means both blockers above (uncleared timer, unvalidated `timeout`) have to be fixed twice and can drift apart afterwards.

Extract it once — e.g. `mergeSignalWithTimeout(fetchOptions, clientTimeout)` in its own module, next to `resolveRequestHeaders` — returning the composite signal plus a disposer the caller invokes in a `finally`. One place to fix, one place to unit-test the clamping and the fallback, and one place to delete after the ofetch 2.0 upgrade.

packages/api-client/src/createAdminAPIClient.ts
patzick: **Minor — the removal path is still undocumented** (@mdanilowicz's open point).

unjs/ofetch#508 makes ofetch 2.x combine `signal` and `timeout` natively, at which point this whole block becomes dead weight. Please leave a `TODO` here and in `createAPIClient.ts` referencing that PR, and file a follow-up task so the shim is actually removed on the 2.0+ bump rather than quietly surviving as a second timer source.

packages/api-client/src/createAdminAPIClient.ts
```diff
@@ -159,6 +160,8 @@ export function createAdminAPIClient<
         await ofetch("/oauth/token", {
           baseURL: params.baseURL,
           method: "POST",
+          signal: options.signal,
+          timeout: options.timeout,
```
patzick, line 165: **Minor — worth confirming the budget is not doubled on the no-signal path.**

Forwarding `signal`/`timeout` to the refresh is the right fix, and the two stalled-refresh tests cover it. What is not covered: with **no** caller signal, `options.signal` is `undefined` here, so the refresh gets its own ofetch timer for the full `timeout`, and the outer request is then armed separately once `onRequest` returns. A caller asking for `timeout: 5000` on an expired session could wait ~10s.

On the merged path this cannot happen — one composite signal spans both fetches — so the guarantee again differs by whether a `signal` was passed.

A test with an expired session, a refresh endpoint stalling just under the timeout, and a slow `/order` would pin the intended semantics. If a shared deadline is wanted, the merged-signal helper from the other comment is the natural place to create it unconditionally, before `onRequest` runs.

packages/api-client/README.md
```diff
@@ -333,6 +333,21 @@ const request = client.invoke("readContext get /context", {
...
+
+const request = client.invoke("readContext get /context", {
+  fetchOptions: {
+    signal: controller.signal,
+    timeout: 5000,
+  },
+});
+```
+
+Combining the two needs `AbortSignal.any`, available since Chrome 116, Firefox 124 and Safari 17.4. Older browsers keep the previous behaviour, where a per-request `signal` switches the timeout off. A fractional timeout is rounded up to whole milliseconds when combined.
```
patzick: **Nit — the Node floor is missing.**

This package runs server-side in the starter templates, so `AbortSignal.any` requiring **Node 20.3+** matters at least as much as the browser versions. The PR description states it; the README should too, since that is what consumers read before relying on the combined behaviour.

patzick (changes requested):
## 🔍 Review — changes requested

CI is fully green (11/11) and the PR is mergeable. The blockers are behavioural, not signal-driven.

**Blockers**
1. `AbortSignal.timeout()` timer is never cleared — leaks a pinned signal per request until the timeout elapses (both clients).
2. `timeout` reaches `AbortSignal.timeout()` unvalidated — negative / non-finite values throw a synchronous `RangeError` out of `invoke()`, and only on the signal path.

Both were raised by @mdanilowicz and are still unaddressed — the last commit is a merge from `main`, no fix commits followed.

**Majors**
3. `timeout` is still forwarded to ofetch next to the merged signal, so the design silently depends on ofetch 1.5.1 not arming its own timer.
4. The merge block is duplicated verbatim in both clients; both blockers now need fixing twice.

**Minors** — missing TODO tying the shim to ofetch 2.0 (also @mdanilowicz's ask), possible refresh + request budget doubling on the no-signal path, README omits the Node floor.

Details inline. The `isTimeoutError` helper itself, the JSDoc/README wording after the Copilot round, the changeset, and the fallback test are all good.

packages/api-client/src/createAPIClient.ts
mkucmus: Done. `normalizeTimeout` now takes only finite positive values, rounds up and caps at 2147483647. Anything else is ignored instead of throwing. Covered by unit tests and by negative, `Infinity` and `NaN` cases through `invoke()`.

mkucmus: Done. `AbortSignal.timeout` is gone. The helper owns an `AbortController` and a `setTimeout`, and both clients clear it in a `finally`.

packages/api-client/src/createAdminAPIClient.ts
mkucmus: Done. `TODO` added in `mergeSignalWithTimeout.ts`, and the removal is tracked in #2710.

packages/api-client/src/createAPIClient.ts
mkucmus: Done, good catch. The helper owns the controller and the timer, and both call sites release it in a `finally`. A mutation check confirms it: dropping `releaseTimeout()` now fails three tests, where before it left the suite green.

packages/api-client/src/createAPIClient.ts
mkucmus: Done, clamped as you suggested. Added coercion on top, so a timeout arriving as a string still works the way `setTimeout` treats it. Negative and non-finite values are tested on the signal path, both as unit tests and through `invoke()`.

packages/api-client/src/createAPIClient.ts
mkucmus: Done. The timeout is cleared after a successful merge, so there is only ever one timer. One detail: it is assigned `undefined` rather than deleted, because `resolveFetchOptions` is a plain `{...defaults, ...input}` spread, so `delete` would let the client-level default through and leave the second timer in place.

packages/api-client/src/createAdminAPIClient.ts
mkucmus: Done. It now lives once in `mergeSignalWithTimeout.ts`, which returns a disposer the clients call in a `finally`. The clamping and the `AbortSignal.any` fallback are unit-tested there.

packages/api-client/src/createAdminAPIClient.ts
mkucmus: Done. `TODO` in the helper pointing at unjs/ofetch#508, and #2710 tracks the removal.

packages/api-client/README.md
mkucmus: Done. Node 20.3 is listed now, with a note that the package runs server-side so that floor matters as much as the browser ones.

packages/api-client/src/createAdminAPIClient.ts
mkucmus: Confirmed, the budget is doubled on the no-signal path. Two tests pin both halves now: with no signal the refresh and the request each get the full timeout, and the test asserts `onAuthChange` fired and that the whole call outlived the configured timeout; with a signal one deadline spans both. The README says so as well.

I left the deadline conditional rather than creating it unconditionally. Making it unconditional would also change the no-signal error shape, since ofetch would stop wrapping the timeout itself, and that felt like more than this PR should carry. Happy to do it here if you would rather have the asymmetry gone.

One thing that came out of looking at this, which I would value your view on: bounding the refresh at all means an aborted refresh can lose a token rotation the server already committed. Details in the PR description and in the comment below.

mkucmus:
@patzick @mdanilowicz one thing I want a second opinion on before this merges, since it is the only open risk left.

Forwarding `signal` and `timeout` to the token refresh (`createAdminAPIClient.ts:163-164`, added earlier in this PR after the Copilot review) means an aborted refresh can lose a token rotation the server has already committed. league/oauth2-server's `RefreshTokenGrant` revokes the presented refresh token and issues a new one, which is why `updateSessionData` re-stores `refresh_token` on every refresh. If the refresh is aborted after the server processed it but before we read the response, the client keeps the revoked token.

What makes it stick is the fallback in `onRequest`:

```ts
const body = params.credentials && !sessionData.refreshToken ? params.credentials : { grant_type: "refresh_token", ... };
```

`sessionData.refreshToken` is still set, just dead, so a client that holds `credentials` never falls back to them. Every later `invoke()` re-enters the refresh branch and gets 401 "The refresh token is invalid", and `isTimeoutError()` is false for those, so the caller cannot attribute them to the abort. The session does not recover on its own.

I reproduced it on the timeout path and on the pure cancellation path with no timeout configured anywhere. Both cases pass on a build with only those two lines removed, so it is new in this PR rather than upstream behaviour.

My preference is to keep bounding the refresh with the `timeout` but not with the caller's `signal`. A stalled refresh stays bounded, which is what the original review asked for, while a cancel-on-unmount can no longer corrupt session state. The cost is that a cancel does not abort an in-flight refresh, so cancel latency can reach the timeout, and the resolved timeout has to be carried to `onRequest` rather than nulled on the merged path.

The alternative, if you would rather keep the signal, is to clear `sessionData.refreshToken` when a refresh is aborted, so a client with `credentials` re-authenticates on the next call. That throws away a token that may still have been good, so I like it less.

Either way it is a change to session semantics, so I did not want to slip it in here unreviewed. Happy to do it in this PR or split it out, whichever you prefer. For now the consequence is documented in the README.

patzick (approved):
## ✅ Approved — all findings addressed

Re-checked at `54b6db1`. CI 12/12 green, including the Cursor security review.

| Finding | Fix | Verified |
| --- | --- | --- |
| Uncleared `AbortSignal.timeout` timer | Helper owns an `AbortController` + `setTimeout`; both clients call `releaseTimeout()` in a `finally` | `createAPIClient.ts:223-240`, `createAdminAPIClient.ts:257-273` |
| Unvalidated `timeout` → `RangeError` | `normalizeTimeout` coerces, rejects non-finite/≤0, rounds up, caps at `2_147_483_647` | `mergeSignalWithTimeout.ts:19-25` |
| Stale `timeout` forwarded next to the merged signal | `fetchOptions.timeout = undefined` after a successful merge | `mergeSignalWithTimeout.ts:72` |
| Duplicated merge block | Extracted to `mergeSignalWithTimeout.ts`, unit-tested on its own | new file + 17 tests |
| Missing ofetch 2.0 removal path | `TODO(#2710)` on the helper, tracked in #2710 | `mergeSignalWithTimeout.ts:43-44` |
| Possible doubled refresh budget | Confirmed, pinned by a test, and documented | README:355 |
| README missing Node floor | Node 20.3 listed, with why it matters server-side | README:349 |

@mdanilowicz — your three points (validation, timer ownership, ofetch#508 TODO) are all covered by the same commits, so this is unblocked from my side too.

The implementation ended up better than what I suggested: assigning `undefined` rather than `delete`-ing `timeout` is the correct call given `resolveFetchOptions` is a plain spread, `timeout` arriving as a string is now coerced the way `setTimeout` would, and the caveat about a rotated refresh token surviving an aborted refresh (README:357) is a real edge case I had not raised. `timeout: 0` reads as "no timeout" and deliberately does not fall back to the client value — consistent with the old ofetch behaviour and covered by a test.

No blockers left.

PR #2703, opened 2026-09-07
refactor: streamline AGENTS.md files 

### LINKED ISSUE #2701: AGENTS.md files are 4x over the recommended size limit and not loaded by Claude Code
Our `AGENTS.md` files have grown to a size where they cost more than they deliver. Three problems, in order of impact: they are not loaded by Claude Code at all, they are **4x over the recommended size limit**, and they duplicate content that already lives in READMEs, the docs site and `templates/manifest.json` — with measurable drift already present.

---

### 1. The 200-line recommendation — we are 4x over it

This is the core of the issue. The official Claude Code guidance on instruction files is explicit:

> **Size**: target under 200 lines per CLAUDE.md file. Longer files consume more context and reduce adherence.
> — [code.claude.com/docs/en/memory#write-effective-instructions](https://code.claude.com/docs/en/memory#write-effective-instructions)

Two things matter in that sentence. The obvious one is context cost. The important one is **"reduce adherence"** — an oversized instruction file does not merely waste tokens, it makes the agent follow *fewer* of the instructions in it. A 805-line file is not a more thorough 200-line file; it is a less effective one. `/doctor` in Claude Code now ships a dedicated trim check for exactly this reason, and it proposes cutting the same categories we are heaviest in: directory layouts, dependency lists and architecture overviews.

Where we stand:

| File | Lines | Over budget | ~Tokens per session |
| --- | --- | --- | --- |
| `AGENTS.md` (root) | **805** | **4.0x** | ~7,000 |
| `packages/cms-base-layer/AGENTS.md` | **532** | **2.7x** | ~4,700 |
| `packages/api-client/AGENTS.md` | 138 | within budget | ~1,700 |
| **Total** | **1,475** | | **~13,400** |

`packages/api-client/AGENTS.md` is the one file that already sits inside the limit, and it is also the most useful of the three. That is not a coincidence — it should be the template for the other two.

The 200-line target is the acceptance criterion for this issue, not a stylistic preference.

### 2. Claude Code does not read `AGENTS.md`

> Claude Code reads `CLAUDE.md`, not `AGENTS.md`.
> — [code.claude.com/docs/en/memory#agents-md](https://code.claude.com/docs/en/memory#agents-md)

There is no fallback. In this repository:

- there is no `CLAUDE.md` in the root, in `.claude/`, or as a symlink;
- `grep -rn "AGENTS.md"` across the repo returns **zero** hits — no script, workflow, `CONTRIBUTION.md` entry or PR template references these files;
- `.cursor/` is empty.

So the 1,475 lines currently reach Codex, Cursor, Copilot, Gemini CLI and Jules, and reach Claude Code not at all. The fix is one line:

```bash
printf '@AGENTS.md\n' > CLAUDE.md
```

A `CLAUDE.md` containing `@AGENTS.md` keeps a single source of truth and lets us append Claude-specific notes below the import. This should land together with the trim — wiring up a 805-line file first would make adherence worse, not better.

### 3. Duplication that has already drifted

`packages/cms-base-layer` carries a 521-line `README.md` and a 532-line `AGENTS.md` documenting the same subjects: responsive images, background image optimisation, `app.config.ts`, component overriding, image presets. The "Responsive Image Architecture" section is a near-verbatim copy, and the copies have already diverged:

```
README:  Fixed `width`/`height` (400px) — avoid hydration mismatches
AGENTS:  Fixed `width`/`height` (400) avoid hydration mismatches

README:  ⚠️ Avoid ... they trigger Vue hydration attribute mismatches
AGENTS:  Note: Avoid ... they've caused Vue hydration attribute mismatches
```

The README also has a `### Slider Components` section the AGENTS.md never got. The file is maintained by copy-paste, which guarantees this gap widens.

In the root file, **311 of 805 lines (39%)** are the "Working with Templates" section, which duplicates `templates/manifest.json`, `apps/docs/src/introduction/templates.md`, `.../templates/vue-starter-template.md`, `.../templates/vue-starter-template-extended.md` and `getting-started/templates/`. The section itself states *"This table is for humans. Code should read `templates/manifest.json`"* — and then describes the manifest in prose for another 40 lines, even though the manifest has `manifest.schema.json` and is validated in CI by `pnpm run check:templates`.

### 4. Internal contradiction: the root file recommends a deprecated template

`AGENTS.md` mentions `vue-demo-store` 11 times, including in the places an agent actually reads:

- TL;DR: `Templates: vue-demo-store (full featured), vue-starter-template (production)…` — deprecated template listed first
- Quick Start: `pnpm run dev --filter=vue-demo-store`
- `**Use Case**: Full reference for production applications`

`templates/manifest.json` — which the same file declares the single source of truth — says `supportLevel: "deprecated"` and *"new projects must not start from it"* (#2667). The deprecation is mentioned in `AGENTS.md` only at line **657**, parenthetically, next to the e2e script. An agent reading the TL;DR of an 805-line file will point users at the deprecated template. This is the most damaging consequence of the size problem: the correct information is present but unreachable.

mdanilowicz:
closes #2701 

This pull request introduces and documents a new convention for AI agent instruction files across the repository, clarifies the handling of agent instruction files for Claude and other agents, and improves documentation in both the root and package-level guides. It also refines the `cms-base-layer` documentation for image handling and component overrides.

**AI agent instruction files and documentation:**

* Added a section to `CONTRIBUTION.md` explaining the use of `AGENTS.md` for AI coding agents and the necessity of a corresponding one-line `CLAUDE.md` for Claude Code; clarified file placement, content guidelines, and update responsibilities.
* Added `@AGENTS.md` as the content of every `CLAUDE.md` (root and package-level) to ensure Claude Code loads the correct agent instructions.
* Rewrote `packages/api-client/AGENTS.md` to focus on non-obvious conventions, the type system, and extension points for contributors and AI agents, removing redundant or obvious information and emphasizing generated artifacts and test structure.

**Documentation improvements in `cms-base-layer`:**

* Clarified in `README.md` that the `productCard` image preset only affects URL modifiers, not rendered attributes, and explained the handling of `width`, `height`, `loading`, and `densities` in the context of `@nuxt/image` 2.1.0.
* Updated the component override instructions to specify that overrides must be placed in a globally registered `components` directory (e.g., `app/components/cms/` with `global: true` and `pathPrefix: false`), ensuring correct Nuxt component resolution.


AGENTS.md
```diff
@@ -1,805 +1,151 @@
...
+- For a stale Turbo cache, `rm -rf .turbo`. If dependencies themselves look
+  wrong, `rm -rf node_modules && pnpm i` — **keep `pnpm-lock.yaml`**. It is
+  committed and nearly every CI job installs with `--frozen-lockfile` (the
+  StackBlitz template check is the exception), so regenerating it turns a local
+  hunch into a repo-wide diff; what you lose are the pinned, reviewed
+  resolutions and their integrity hashes. The `overrides` and `minimumReleaseAge`
+  guard live in `pnpm-workspace.yaml` and are re-applied on every install, so
+  those survive either way.
+- Never hand-edit `packages/api-client/api-types/*.d.ts` — they are generated
+  by `@shopware/api-gen`. See that package's `AGENTS.md` for which script
```
mkucmus: P1: This points to packages/api-gen/AGENTS.md, but that file does not exist. Please link directly to packages/api-client/AGENTS.md, where the generation scripts are documented.

packages/cms-base-layer/AGENTS.md
```diff
@@ -1,532 +1,114 @@
...
+into the page, visibly, in production as well as locally. Nothing is logged, so
+there is no console warning to look for.
+
+Registration rules for the consuming template — which directories must be
+`global: true`, and the trap of registering one path twice — are commented at the
+`components` key of
+[vue-starter-template/nuxt.config.ts](../../templates/vue-starter-template/nuxt.config.ts).
+
+## Listing filters
+
```
mkucmus: P1: This filter guidance is incomplete. The fallback supports property-group filters, but not every backend filter with options. Adding an independent filter also requires event handling, handleFilterChange, URL serialization, reset/chip behavior, both sidebar and horizontal components, and search-page parsing.

AGENTS.md
```diff
@@ -1,805 +1,151 @@
...
+- `cms-base-layer` and `unocss-design-tokens-layer` are **plain Nuxt layers with
+  no build step**. They ship their sources as-is, so edits show up in templates
+  immediately.
+- `composables` needs no build for its main entries either: `.` resolves to
+  `src/index.ts` and `./nuxt-layer` to `nuxt.config.ts`, so source edits are live.
+  Only its `./lib` and `./dist` entries come from a build.
+- Every remaining package must be built, or run in stub mode, before a template
+  sees the change: `cd packages/<name> && pnpm run dev` (`unbuild --stub`).
+- The layer packages run `nuxt prepare` as the first half of their own
+  `typecheck`, so a type error there is never stale `.nuxt` output — re-running
```
mkucmus: P1: This troubleshooting rule is too strong. Type errors can have several causes, including stale preparation output, source errors, or configuration. Please guide agents to inspect the actual script, error location, preparation output, and dependency output.

AGENTS.md
```diff
@@ -1,805 +1,151 @@
 # AI Agent Guide for Shopware Frontends
 
-This document provides guidance for AI assistants working with the Shopware Frontends monorepo.
+Vue 3 / Nuxt 4 framework for building Shopware 6 storefronts. pnpm workspace
+monorepo, Turbo for build orchestration.
 
-## TL;DR
+Repository layout, dependency lists and available scripts are discoverable from
```
mkucmus, line 6: P2: Please add a compact task map here with verified starting files for API client work, API generation, composables, CMS, Nuxt integration, templates, and testing. Without it, agents still need broad repository searches.

AGENTS.md
```diff
@@ -1,805 +1,151 @@
...
-
-# Run tests in watch mode
-pnpm run test:watch
-
-# Check types continuously
-pnpm run typecheck
-```
-
-### 3. Before Committing
+## Quick start
```
mkucmus, line 10: P2: Please document focused validation commands as well, such as one test file, package tests, and package typechecking. The current quick start mainly points agents to broad workspace commands.

CONTRIBUTION.md
```diff
@@ -37,3 +37,32 @@ In order to keep track of the changes inside project we do use `changesets` pack
 
 We're using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) to keep our commit messages consistent. However to avoid having too much noise, we're squashing all commits into one when merging PRs.
 That's why the PR title is the commit message for the whole PR. Please make sure to follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format when creating a PR title. We do have a GitHub Action that will check if the PR title is valid.
+
+## Agent instruction files
```
mkucmus, line 41: P2: This section is too detailed for the shared contributor guide. Please keep a short note and link to AGENTS.md here; agent-specific loading, context, and maintenance rules belong in the agent guides.

packages/cms-base-layer/AGENTS.md
```diff
@@ -1,532 +1,114 @@
...
+`CmsBlockSpatialViewer` both create the async component unconditionally at setup
+and gate at render instead, on `isSpatial(currentImage)` and on `modelUrl`.
+Either shape keeps the chunk out of the initial bundle — just don't assume the
+gallery matches `CmsElementImage`. Consuming apps must add `@tresjs/nuxt` to
+their modules themselves.
+
+## Documenting components
+
+CMS **elements** and **sections** carry a sibling `.md` file
+(`CmsElementImage.vue` next to `CmsElementImage.md`): 18 of 20 elements and both
```
mkucmus: P3: Exact component counts will become outdated as the package changes and do not add much navigation value. Please describe the documentation convention with an example instead.

mkucmus (changes requested):
Thank you for the substantial reduction in duplicated guidance. This is a good direction and should be preserved.

I have requested a few small follow-up changes in the inline comments:

- Fix the broken API-generation reference.
- Complete and correct the listing-filter workflow.
- Make troubleshooting guidance less absolute.
- Add a compact task-to-entry-point map.
- Document focused validation commands.
- Keep detailed agent policies out of the shared contributor guide.
- Remove fragile component counts.

These changes will make the shorter documentation more reliable and easier to use for both coding agents and human contributors.

packages/cms-base-layer/AGENTS.md
mdanilowicz: I removed the listing filters section and added a comment to state property. 
That approach will make more sense.

AGENTS.md
mdanilowicz: That map used to be here: `Project Structure`, `Key Files to Know` and the
per-package `Key Files` sections, around 134 lines, removed earlier in this PR.
Nothing validated them and they had drifted from `templates/manifest.json` —
which is how a deprecated template ended up recommended as the starting point
for new projects.

In practice each area you list is one `ls` away, on a directory whose name is
already the answer (`packages/api-client/src` → `createAPIClient.ts`, and so
on). The per-package detail now lives in each package's `README.md`, plus an
`AGENTS.md` for `api-client` and `cms-base-layer`, so it loads when an agent is
actually working there and stays with whoever changes the code.

AGENTS.md
mkucmus: Thanks for the suggestion. We are intentionally keeping the root guide compact: the former task map duplicated repository structure and had drifted from templates/manifest.json. Package-specific entry points are now documented in the package README files and local AGENTS.md files, so the guidance stays next to the code it describes. I am resolving this as an intentional documentation decision.

AGENTS.md
mkucmus: Thanks for the suggestion. We are intentionally keeping focused commands in each package package.json, where they are maintained with the package scripts, while the root guide keeps workspace-level commands only. This avoids duplicating commands that can drift from the actual scripts. I am resolving this as an intentional documentation decision.

mkucmus (approved):
🤖 I'm so happy you keep it clean

Issue #2691, opened 2026-09-03
useListing ignores cacheableReads, so listings and search stay on POST


mkucmus:
Part of #2634

## Problem

`cacheableReads` opts reads into cacheable GET Store API routes instead of POST. Ten composables branch on it. `useListing` does not, and does not even destructure it from the context:

```ts
// packages/composables/src/useListing/useListing.ts
const { apiClient } = useShopwareContext();
```

Both search methods hardcode the POST operation with no condition:

```ts
apiClient.invoke("searchPage post /search", { ... })
apiClient.invoke("readProductListing post /product-listing/{categoryId}", { ... })
```

Compare `useCategorySearch`, which does it properly:

```ts
const { apiClient, cacheableReads } = useShopwareContext();
const result = cacheableReads
  ? await apiClient.invoke("readCategoryGet get /category/{navigationId}", { ... })
  : await apiClient.invoke("readCategory post /category/{navigationId}", { ... });
```

So a template that sets `cacheableReads: true`, as `vue-starter-template` does, gets it everywhere except the two heaviest read endpoints on the site.

## Why it matters

Two costs, one of them not obvious.

- No CDN caching for category listings and search, which is the whole point of the flag.
- No automatic retry. ofetch defaults to `retries = isPayloadMethod(method) ? 0 : 1`, so GETs are retried once on a network error and POSTs are not retried at all. Measurements in #2682 show Store API connections failing outright and quickly, which is exactly the case a single retry recovers. Listings currently get none.

## The GET variants exist and take full criteria

```
readProductListingGet get /product-listing/{categoryId}
searchPageGet         get /search
searchSuggestGet      get /search-suggest
```

They accept `_criteria`, a base64url(gzip(json)) encoded criteria object, so filters, aggregations and `post-filter` all fit. There is no technical blocker.

## Open question that must be answered first

A GET response is publicly cacheable. For the other ten endpoints (category, countries, salutations, navigation) the payload is the same for everyone, so the risk is negligible. A product listing is the first case where the response can depend on the customer: customer group pricing, or net versus gross.

Someone who knows the sales channel setup needs to confirm that listing payloads are not customer dependent, or that the GET variant respects context, before this is flipped. Getting this wrong trades flaky tests for wrong prices, which is a far worse outcome.

## Done when

`useListing` honours `cacheableReads` like every other read composable, or there is a written reason in the code why it cannot.

PR #2720, opened 2026-09-15
fix: use cacheable GET reads only for fresh default guest sessions

### LINKED ISSUE #2692: cacheableReads is applied per call site with nothing enforcing completeness
Part of #2634

## Problem

`cacheableReads` is a single global flag, but it is applied by hand in each composable. `createShopwareContext` only surfaces it:

> Opt in to reading data via cacheable GET Store API routes instead of POST. Surfaced on the Shopware context as `cacheableReads`.

Every read site then has to remember to branch on it. Ten do. #2691 is one that does not, and it happens to cover the two busiest read endpoints on a storefront.

Nothing catches a missed site. Not a type, not a test, not a lint rule. The flag reads as global, so "enabled" looks like "applied everywhere" while the truth is per call site. Whoever adds the next read composable has no signal that the branch is expected.

## Why this is the more important of the two

#2691 is a bug and can be fixed in one file. This is the reason it existed, and the reason the next one will.

## Options

- Centralise it in the API client: keep a map from read POST operations to their `...Get get /...` equivalents and swap at invoke time when the flag is on. Call sites stop caring, and the flag becomes genuinely global.
- Or keep the per-site branch, but enumerate the read operations that have a GET variant and add a test asserting each corresponding composable honours the flag. Cheaper, and it fails loudly when someone adds the eleventh.

The first removes the class of bug. The second only detects it, but is far less invasive.

## Worth checking while in here

Whether every operation with a `...Get get /...` variant in `storeApiTypes.d.ts` has a composable that uses it. The three listing ones are known missing via #2691; there may be others nobody has noticed for the same reason.

## Done when

A read composable that ignores `cacheableReads` is caught automatically rather than by someone reading the diff.

mkucmus:
## Problem

With `cacheableReads` on, every session read catalog data over GET. Shared caches key on the URL, not on `sw-context-token`. So a logged-in, cart or currency-switched session could get the cached guest response. On the starter, a €3,417.11 product showed €4,000.92 after a currency switch. Cached responses also carried the first visitor's token. Nothing checked that read composables honour the flag.

## Fix

- `@shopware/composables`: new `useCacheableRead().invokeRead(postOperation, params)`. It sends the GET twin only when the loaded session and cart look like a fresh default guest. That GET has no `sw-context-token`. Every other session keeps the POST. All flag branches use it. `useShopwareContext().cacheableReads` is deprecated for branching.
- `@shopware/api-client`: new `mergeRequestHeaders`. Header names match case-insensitively. A header set to `""` is removed. A request that drops the client token never adopts one from its response. `defu` dependency removed.
- `@shopware/nuxt-module`: app-wide `swCart` ref, plus `guestServerRender` for server renders without a visitor session.
- Coverage test: fails when a POST read with a GET twin is unclassified, or when code calls a cacheable read outside `invokeRead`.
- Docs: caching guide corrected, AGENTS.md, starter README.

Closes #2692

## Why `invokeRead` and not per-call branches

The decision needs the session, the cart and the token. Written by hand, that is 16 copies. They already drifted: `useListing` never got a branch (#2691), `useProductConfigurator` put `includes` where GET ignores it. One helper, tested once. The trade-off: the code names the POST operation while the network may show GET.

## Verification

- Unit tests: composables (100% coverage), api-client, nuxt-module. Lint and typecheck.
- Browser check against the demo backend. Guest reads go GET without a token. After a currency switch or add-to-cart they go POST with correct prices.

## Not in this PR

- `useListing` (#2691), `useLandingSearch`, `useProductAssociations`, and the starter's currency and search-suggest reads stay POST. Listed in the coverage test.
- Non-default languages stay POST. Shopware 6.7.6 to 6.7.14 do not key the built-in cache on `sw-language-id`. Can be relaxed after shopware#20200 ships.


packages/composables/src/useCacheableRead/useCacheableRead.ts
```diff
@@ -0,0 +1,203 @@
...
+  function invokeRead(
+    operation: CacheableReadOperation,
+    ...params: [ReadParams?]
+  ) {
+    const canUseGet =
+      cacheableReads &&
+      canUseCacheableGet({
+        headers: mergeRequestHeaders(
+          params[0]?.headers,
+          apiClient.defaultHeaders,
```
mdanilowicz, line 171: Should be 
```suggestion
          apiClient.defaultHeaders ?? {},
```

packages/composables/src/useCacheableRead/useCacheableRead.ts
```diff
@@ -0,0 +1,203 @@
...
+    "swCart",
+    null,
+  );
+
+  const invoke = apiClient.invoke as unknown as (
+    operation: string,
+    ...params: [ReadParams?]
+  ) => Promise<unknown>;
+
+  function invokeRead(
```
mdanilowicz, line 162: Why not just use POST requests if we have a GET alternative?  Aren't we overcomplicating the code? POST endpoints should be deprecated at some point.

mdanilowicz (changes requested):

packages/composables/src/useCacheableRead/useCacheableRead.ts
mkucmus: Thanks, fair point, and GET is where we want to end up too.

Today it does not work for every session. I checked on the demo backend: a currency-switched session gets USD prices over GET and EUR over POST. The backend keeps sessions apart with a hash header, and a browser on another domain cannot send it, CORS blocks it. That is #2692.

So the gate sends GET only when the session is a plain default guest, the same checks the backend makes, and POST otherwise. Once the browser is allowed to send the hash, most of the gate can go. That looks like a gap in core's CORS list, so I will raise it there.

packages/composables/src/useCacheableRead/useCacheableRead.ts
mkucmus: Good call. I put the guard in `mergeRequestHeaders` instead, so `defaultHeaders` may be undefined there, like `callerHeaders` already can. Every caller gets it. A missing header set now falls through to POST instead of throwing.

packages/composables/src/useCacheableRead/useCacheableRead.ts
```diff
@@ -0,0 +1,203 @@
...
+    operation: CacheableReadOperation,
+    ...params: [ReadParams?]
+  ) {
+    const canUseGet =
+      cacheableReads &&
+      canUseCacheableGet({
+        headers: mergeRequestHeaders(
```
patzick, line 169: mergeRequestHeaders is computed here to decide GET vs POST, then the request headers are recomputed again inside apiClient.invoke/resolveRequestHeaders for the actual call. Duplicate work per read, and the two merges can drift if one header set changes without the other.

packages/composables/src/useCacheableRead/useCacheableRead.ts
```diff
@@ -0,0 +1,203 @@
...
+}
+
+/**
+ * Uses the cacheable GET route for fresh default guests. Everything else stays
+ * POST.
+ *
+ * @public
+ * @category Context & Language
+ */
+export function useCacheableRead(): UseCacheableReadReturn {
```
patzick, line 145: useCacheableRead() isn't wrapped in createSharedComposable (unlike useCart/usePrice/useProductSearchListing), so each composable that calls it repeats useShopwareContext() + two injectLocal lookups and allocates its own invokeRead closure.

packages/composables/src/useCacheableRead/useCacheableRead.ts
```diff
@@ -0,0 +1,203 @@
...
+      : headers["sw-language-id"] || headers["sw-currency-id"]
+  ) {
+    return false;
+  }
+  // no token: the backend answers as a fresh guest
+  if (!token) return true;
+  // session unknown, or not refreshed after a token change
+  if (session?.token !== token) return false;
+  // guest server render: the cart is empty
+  if (guestServerRender) return true;
```
patzick, line 134: guestServerRender short-circuits before the cart-emptiness check on line 135. If guestServerRender is ever true while a real cart with line items already exists on the session, this returns true without checking cart.lineItems — bypassing the exact guarantee the non-SSR path relies on.

patzick (commented):
[Automation] Review notes — no correctness bugs found, a few minor efficiency/maintainability observations inline.

packages/composables/src/useCacheableRead/useCacheableRead.ts
mkucmus: Closing this one. Where it landed: GET stays limited to plain default guests until browsers may send `sw-cache-hash`. Core's CORS list still blocks that header on trunk. Once it is allowed, the gate shrinks to a flag check and every session can use GET. I will raise the CORS gap upstream as the next step. Happy to reopen if you see it differently.

packages/composables/src/useCacheableRead/useCacheableRead.ts
mkucmus: Thanks. Both places call the same `mergeRequestHeaders`, so they cannot drift. The cost is one small object per read. I'd keep it.

packages/composables/src/useCacheableRead/useCacheableRead.ts
mkucmus: Good thought. The lookups happen once per composable at setup, not per request. A shared instance would also freeze the injections of the first caller, which breaks component-local providers. `useShopwareContext` is not shared for the same reason.

packages/composables/src/useCacheableRead/useCacheableRead.ts
mkucmus: Fair point. It now skips the cart check only when no cart was loaded. A loaded cart with items falls back to POST. Test added in a62d8d51.

packages/composables/src/useCacheableRead/useCacheableRead.ts
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * POST reads with a cacheable GET twin. `query`: body keys sent as plain query
+ * params.
+ */
+export const cacheableReadRoutes = {
+  "readCategoryList post /category": {
+    get: "readCategoryListGet get /category",
+  },
+  "readCategory post /category/{navigationId}": {
+    get: "readCategoryGet get /category/{navigationId}",
```
patzick, line 25: **Blocker — the criteria is silently dropped on this route.**

`readCategoryGet` is the only one of the ten registered GET twins whose generated type has no `_criteria` query parameter (`packages/api-client/api-types/storeApiTypes.d.ts:10168-10231` — it declares `associations`, `slots`, `limit`, `p`, `page`, `order`, `properties`, `sort[]`, `filter[]`, … but no `_criteria`). Verified across all ten pairs; the other nine all declare it.

`invokeRead` sends `_criteria` unconditionally (line 194), so Shopware just ignores it here and the whole criteria object is lost.

Concrete failure: a fresh default guest with `cacheableReads` enabled opens a category page → `useCategorySearch.search(id, { withCmsAssociations: true })` builds `body: { associations: cmsAssociations }` → `GET /category/{id}?_criteria=<encoded>` → the `Category` comes back without `cmsPage`/`media` associations and without the listing params, so the page renders with no CMS content. The response *type* is identical to the POST twin, so nothing fails loudly — it just renders empty.

Either give this entry a `query` mapping for the params `readCategoryGet` does accept, or move `readCategory` to `postOnlyReads`.

packages/composables/src/useCacheableRead/cacheableReadCoverage.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  );
+  const bypasses = forbiddenKeys.filter((key) =>
+    quotes.some((quote) => rest.includes(quote + key + quote)),
+  );
+  if (/\bcacheableReads\b/.test(source)) bypasses.push("cacheableReads");
+  return bypasses;
+}
+
+describe("cacheable reads coverage", () => {
+  it("classifies every POST read that has a GET twin", () => {
```
patzick, line 82: This is why the `readCategoryGet` mismatch above passed CI: the test only pairs *operation names*, it never asserts the GET twin can actually carry what `invokeRead` sends it.

Worth extending to assert that each registered GET twin declares `_criteria`, or that every body key it cannot carry is listed in that route's `query` array — otherwise the next twin added with a divergent param surface fails the same silent way.

packages/composables/src/useUser/useUser.test.ts
```diff
@@ -412,26 +412,46 @@ describe("useUser", () => {
...
           _criteria: encodeForQuery({
             filter: [{ field: "id", type: "equals", value: "salutation-id" }],
           }),
         },
-      }),
+      },
+    );
+  });
+
+  it("loadCountry keeps POST for a logged-in session when cacheableReads is enabled", async () => {
```
patzick, line 440: Nit: this passes for the wrong reason. The test is named "keeps POST for a logged-in session", but the fixture has no `salesChannel`, so `isDefaultGuest` already returns `false` at the `!!salesChannel` check and the `!session.customer` branch is never reached. Deleting `!session.customer &&` from the gate leaves this test green.

Same pattern in `useProductSearch.test.ts`. A fully default-guest fixture varying only the field under test would pin the intended branch.

packages/composables/src/useProductSearch/useProductSearch.test.ts
```diff
@@ -46,15 +47,36 @@ describe("useProductSearch", () => {
...
           _criteria: encodeForQuery(cmsAssociations),
         },
         pathParams: {
           productId: "test",
         },
+      },
+    );
+  });
+
+  it("keeps POST for a non-default currency when cacheableReads is enabled", async () => {
```
patzick, line 63: Nit, same as the `useUser` one: the fixture omits `countryId`/`paymentMethodId`/`shippingMethodId`, so the `&&` chain short-circuits on a later `sameId()` regardless. Removing the currency `sameId()` check leaves this test green.

patzick (changes requested):
[Automation] Re-review at `a62d8d51`. The three earlier notes are resolved or correctly declined — the `guestServerRender` cart check now falls back to POST for a loaded non-empty cart, and the `mergeRequestHeaders` duplicate-call concern does not hold (both calls take the same `defaultHeaders` proxy and the same caller-headers ref, with no await between, so they cannot disagree).

One blocker found in the route registry, plus the coverage test that should have caught it. Details inline.

packages/composables/src/useCacheableRead/useCacheableRead.ts
mkucmus: Thanks for digging into this. I checked the backend: `CategoryRoute::load` has no `Criteria` parameter. It builds its own criteria, so the request criteria is ignored on GET and on POST alike. Verified live on the demo: `associations.seoUrls` is dropped both ways, and a wrong `id` filter still returns 200 both ways. So GET loses nothing that POST delivered, and `main` already made this call. The missing `_criteria` in the type is an OpenAPI gap, same as `searchPageGet`. The route does read `slots` from the request, so I added a `slots` query mapping.

packages/composables/src/useCacheableRead/cacheableReadCoverage.test.ts
mkucmus: Good idea in principle. The backend decides this, not the type. `RequestCriteriaBuilder` decodes `_criteria` on every GET route since 6.7.6, and the type lags for two routes. A type check here would flag the wrong thing, so I'd leave it.

packages/composables/src/useUser/useUser.test.ts
mkucmus: Good catch. Both tests now use a full default guest fixture and change only the field under test.

packages/composables/src/useProductSearch/useProductSearch.test.ts
mkucmus: Good catch. Both tests now use a full default guest fixture and change only the field under test.

packages/composables/src/useCacheableRead/useCacheableRead.ts
```diff
@@ -0,0 +1,204 @@
...
+    query: ["slots"],
```
patzick, line 26: This does not close the criteria-drop. Re-verified against `storeApiTypes.d.ts` at `d163edcf`: `readCategoryGet` (line 10168) still declares **no** `_criteria` — it is the only one of the ten twins without it — while it *does* declare `associations` and `slots` as plain query params.

With `query: ["slots"]`, `invokeRead` (lines 185-197) pulls only `slots` out; everything else still goes into `_criteria`, which this route ignores.

The actual failing call doesn't send `slots` at all — `useCategorySearch.search(id, { withCmsAssociations: true })` builds `body: { associations: cmsAssociations, ...query }`. So a fresh default guest opening a category page still gets a `Category` back with no `cmsPage`/`media` associations, and the page renders without CMS content.

`associations` (and the listing params `p`, `page`, `order`, `properties`, `limit`, `sort[]`, `filter[]`) need to be in the `query` list too, or `readCategory` should move to `postOnlyReads`. Note the query-param encoding for `associations` is not the same shape as the `_criteria` payload, so it needs checking against a real response rather than just the types.

The coverage-test point from the previous round still applies: `cacheableReadCoverage.test.ts` only pairs operation names, so neither the original mismatch nor this partial fix fails CI.

patzick (changes requested):
[Automation] Re-check at `d163edcf`. The two test-fixture nits are properly fixed by the shared `guestSession()`/`cartOf()` helpers — that reads well. The blocker is still open, though: `query: ["slots"]` only moves `slots` out of `_criteria`, and `slots` is not what the failing call sends. Detail inline.

PR #2722, opened 2026-09-15
feat(docs): add Cart recipe

### LINKED ISSUE #2721: Cart recipes

mdanilowicz:
closes #2721 

This pull request adds initial documentation and navigation for the Checkout section in the frontend recipes of the docs app. It introduces a new sidebar entry, a landing page for Checkout, and updates the main recipes index to reference this new section.

**Documentation and Navigation Additions:**

* Added a new "Checkout" section to the sidebar navigation, including a nested "Cart" page link (`apps/docs/.vitepress/sidebar.ts`).
* Created a new `checkout` index page introducing checkout-related recipes and referencing the "Cart" documentation (`apps/docs/src/frontends-recipes/checkout/index.md`).
* Updated the main frontend recipes index to include a summary and link to the new "Checkout" section (`apps/docs/src/frontends-recipes/index.md`).

**Component Addition:**

* Added a new `CodeExample.vue` component for displaying code examples with features like copy-to-clipboard, expandable/collapsible code blocks, and accessibility improvements (`apps/docs/src/components/CodeExample.vue`).


apps/docs/src/frontends-recipes/checkout/cart.md
```diff
@@ -0,0 +1,425 @@
...
+
+## State And Session
+
+The cart belongs to the sales channel session identified by the `sw-context-token` header, not to the customer. A guest has a cart, and logging in does not merge two carts in the frontend — the Store API resolves the cart for the token it receives.
+
+`useCart` is wrapped in `createSharedComposable`, so every call in the application returns the same instance. The cart itself lives in the `swCart` context value and the collected errors in `swCartErrors`, which is what makes a mini cart in the header and a cart page stay in sync without any prop passing or store of your own.
+
+That shared instance is module-scoped, and on the server nothing tears it down between requests — component scopes are never stopped after a render. **Never fetch the cart during SSR.** Load it from `onMounted` (or behind `import.meta.client`), as the example does; a `useAsyncData` wrapper around `refreshCart()` would write one customer's cart into process-global state, and under a template's `isr` route rules that HTML is then cached and served to everyone.
```
mkucmus: I think this part is not accurate. `createSharedComposable` in the pinned `@vueuse/shared` 14.4.0 returns the plain composable on the server (`if (!isClient) return composable`), and `useContext` provides state per app instance. So each request gets its own cart. The advice to load on the client is still right, but for the caching reason. Maybe something like:

```suggestion
`useCart` is not shared on the server. `createSharedComposable` returns the plain composable there, and `useContext` provides state per app instance, so each request gets its own cart. **Still, load the cart on the client.** Templates cache pages with `isr` route rules, so a cart rendered on the server would land in cached HTML and be served to everyone. Loading it from `onMounted` (or behind `import.meta.client`), as the example does, keeps personalized data out of the cache.
```

apps/docs/src/frontends-recipes/checkout/cart.md
```diff
@@ -0,0 +1,425 @@
...
+const pendingItemId = ref("");
+const writeError = ref("");
+const cartErrors = ref<Schemas["CartError"][]>([]);
+
+// Only one write may be in flight: every write returns the whole recalculated
+// cart, so two in parallel race and the slower response overwrites the faster.
+const isWriting = computed(() => pendingItemId.value !== "");
+
+// Client-side only. useCart is a shared composable whose state is module-scoped
+// on the server, so fetching the cart during SSR would leak it between requests.
```
mkucmus: Same point as below in the State And Session section. Could the comment name the caching reason instead?

```suggestion
// Client-side only. Templates cache pages with isr route rules, so a cart
// rendered on the server would end up in HTML served to everyone.
```

apps/docs/src/frontends-recipes/checkout/cart.md
```diff
@@ -0,0 +1,425 @@
...
+         used has just been re-enabled, so focus is nowhere near this message. -->
+    <p v-if="writeError" role="alert">{{ writeError }}</p>
+
+    <ul v-if="cartErrors.length" role="alert">
+      <li v-for="error in cartErrors" :key="error.key">{{ error.message }}</li>
+    </ul>
+
+    <p v-if="isLoading">Loading your cart…</p>
+
+    <p v-else-if="isEmpty">Your cart is empty.</p>
```
mkucmus, line 323: When the first `refreshCart()` rejects, `isLoading` turns false and `isEmpty` is still true. So the customer sees the error and "Your cart is empty." together. The testing checklist below asks for the opposite. Maybe a separate `loadError` ref, rendered as its own branch before `isEmpty`?

apps/docs/src/frontends-recipes/checkout/cart.md
```diff
@@ -0,0 +1,425 @@
...
+</template>
+```
+
+</CodeExample>
+
+`useCartItem` is not used here on purpose. It takes a `Ref<LineItem>` and must be called at the top level of a row component's setup, not inside a click handler in the list component. Extracting each `<li>` into its own row component is the natural next step, and it is what lets a row own its pending state.
+
+The errors are rendered inline rather than pushed through `codeErrorsNotification()`, so the example stands on its own. `codeErrorsNotification()` only writes into `useNotifications()` state — it renders nothing by itself, so it needs a notification outlet mounted somewhere above it, as `vue-starter-template` does with `<LayoutNotifications />` in its layouts.
+
+The example owns its initial load. `vue-starter-template` already calls `refreshCart()` once in `app.vue`, so inside that template you should drop the `onMounted` block here rather than fetching the cart twice on hydration.
```
mkucmus: Small trap here: `isLoading` starts as `true` and only the `onMounted` block sets it to `false`. If a reader drops the whole block, the page shows "Loading your cart…" forever. Maybe:

```suggestion
The example owns its initial load. `vue-starter-template` already calls `refreshCart()` once in `app.vue`. Inside that template, drop only the `refreshCart()` call and keep the rest of the `onMounted` block, or `isLoading` never turns false.
```

apps/docs/src/frontends-recipes/checkout/cart.md
```diff
@@ -0,0 +1,425 @@
...
+- `subtotal` reads `cart.price.positionPrice` and `totalPrice` reads `cart.price.totalPrice`. They differ once shipping costs or promotions apply — do not compute either from the line items yourself.
+- A line item with `stackable: false` must not render a quantity input, and one with `removable: false` must not render a remove button. Both flags come from the cart response.
+- Adding a product that is already in the cart increases the existing line item instead of creating a second one, so `addProduct` can change `count` by more than the quantity you sent.
+- `consumeCartErrors()` clears `swCartErrors`. If two components call it for the same response, only the first one sees the errors — and `codeErrorsNotification()` and `getErrorsCodes()` both call it, so calling one after the other for the same response leaves the second empty.
+- `Cart["errors"]` is a union of `CartError[]` and a keyed map. `codeErrorsNotification()` returns early on the array branch and pushes nothing, so do not rely on it as your only path — `getErrorsCodes()` normalises both.
```
mkucmus: Both consumers return early on the array branch, `getErrorsCodes()` gives `[]` too. In practice it does not matter, because `setCartErrors` runs `Object.assign` and turns an array into an index-keyed object. Maybe describe that instead?

```suggestion
- `Cart["errors"]` is a union of `CartError[]` and a keyed map, but `setCartErrors` merges either shape into one keyed object with `Object.assign`. Both `codeErrorsNotification()` and `getErrorsCodes()` read that object, so you never handle the array shape yourself.
```

apps/docs/src/frontends-recipes/checkout/cart.md
```diff
@@ -0,0 +1,425 @@
...
+
+Customer-specific prices, promotions and rules change with the customer context, so the cart has to be re-read when the session changes. `useUser().login()` and `logout()` call `refreshCart()` internally — but neither awaits it, and only `register()` does. Right after `await login()` resolves, the shared cart is still the pre-login guest cart for one more round trip, so `await refreshCart()` yourself if you render prices immediately after a session change.
+
+## Edge Cases
+
+- `count` only sums line items where `good` is `true`, so a promotion line item is visible in `cartItems` but does not raise the item count.
+- `subtotal` reads `cart.price.positionPrice` and `totalPrice` reads `cart.price.totalPrice`. They differ once shipping costs or promotions apply — do not compute either from the line items yourself.
+- A line item with `stackable: false` must not render a quantity input, and one with `removable: false` must not render a remove button. Both flags come from the cart response.
+- Adding a product that is already in the cart increases the existing line item instead of creating a second one, so `addProduct` can change `count` by more than the quantity you sent.
```
mkucmus: Stacking adds exactly the quantity you sent, so `count` cannot grow by more than that. Maybe:

```suggestion
- Adding a product that is already in the cart raises the quantity of the existing line item instead of creating a second one.
```

apps/docs/src/frontends-recipes/checkout/cart.md
```diff
@@ -0,0 +1,425 @@
...
+
+`useCartItem` is not used here on purpose. It takes a `Ref<LineItem>` and must be called at the top level of a row component's setup, not inside a click handler in the list component. Extracting each `<li>` into its own row component is the natural next step, and it is what lets a row own its pending state.
+
+The errors are rendered inline rather than pushed through `codeErrorsNotification()`, so the example stands on its own. `codeErrorsNotification()` only writes into `useNotifications()` state — it renders nothing by itself, so it needs a notification outlet mounted somewhere above it, as `vue-starter-template` does with `<LayoutNotifications />` in its layouts.
+
+The example owns its initial load. `vue-starter-template` already calls `refreshCart()` once in `app.vue`, so inside that template you should drop the `onMounted` block here rather than fetching the cart twice on hydration.
+
+## State And Session
+
+The cart belongs to the sales channel session identified by the `sw-context-token` header, not to the customer. A guest has a cart, and logging in does not merge two carts in the frontend — the Store API resolves the cart for the token it receives.
```
mkucmus: Shopware does merge the guest cart into the customer cart on login, on the server side. Could we say that, so readers do not think the guest cart is lost?

```suggestion
The cart belongs to the sales channel session identified by the `sw-context-token` header, not to the customer. A guest has a cart. On login the Store API merges it into the customer cart for the new context token, so the frontend only needs to refresh.
```

apps/docs/src/frontends-recipes/checkout/cart.md
```diff
@@ -0,0 +1,425 @@
...
+
+- **Read** — `cart`, `cartItems`, `count`, `isEmpty`, `subtotal`, `totalPrice`, `shippingCosts`, `appliedPromotionCodes`, `isVirtualCart`.
+- **Write** — `addProduct`, `addProducts`, `addPromotionCode`, `changeProductQuantity`, `removeItem`, `removeItemById`, `refreshCart`.
+- **Errors** — `consumeCartErrors()` returns the collected cart errors and clears them.
+
+Four things the generated reference will not tell you:
+
+- `useCartItem` takes a `Ref<LineItem>` and derives the whole row from it — `itemTotalPrice`, `itemStock`, `itemImageThumbnailUrl`, `isStackable`, `isRemovable` and the rest — so a row component needs no props beyond that one ref.
+- `useAddToCart` takes a `Ref<Product | undefined>`. The `undefined` is deliberate: it lets you call the composable at the top level of setup while the product is still loading.
+- `useCartNotification` gives you two ways to handle the same errors — `codeErrorsNotification()` pushes them as notifications, `getErrorsCodes()` returns them as `CartError[]`. Both call `consumeCartErrors()`, so the first one you call clears them for the other. Pick one per response.
```
mkucmus: One detail worth adding: `getErrorsCodes()` drops the `promotion-discount-added` entry, while `codeErrorsNotification()` shows it as a success.

```suggestion
- `useCartNotification` gives you two ways to handle the same errors. `codeErrorsNotification()` pushes them as notifications, `getErrorsCodes()` returns them as `CartError[]`. Both call `consumeCartErrors()`, so the first one you call clears them for the other. Pick one per response. Note that `getErrorsCodes()` drops the `promotion-discount-added` entry, while `codeErrorsNotification()` shows it as a success.
```

apps/docs/src/frontends-recipes/checkout/cart.md
```diff
@@ -0,0 +1,425 @@
...
+    // Consume on both paths: a 2xx response can carry errors, and a rejected
+    // write must not strand earlier ones in the shared state.
+    cartErrors.value = getErrorsCodes();
+    pendingItemId.value = "";
+  }
+};
+
+const changeLineItemQuantity = (item: Schemas["LineItem"], value: string) => {
+  // min="1" constrains the stepper and validation, not the value you read here:
+  // a cleared field yields "" and Number("") is 0.
```
mkucmus: The code uses `parseInt`, which gives `NaN` for `""`. Maybe align the comment:

```suggestion
  // a cleared field yields "", and parseInt("") is NaN.
```

apps/docs/src/components/CodeExample.vue
```diff
@@ -0,0 +1,211 @@
+<script setup>
```
mkucmus: The sibling components (`RecipeFlowDiagram.vue`, `SchemaTypeTooltip.vue`) use `lang="ts"`. Would you like to match them here?

apps/docs/src/components/CodeExample.vue
```diff
@@ -0,0 +1,211 @@
...
+  border-top: 1px solid var(--vp-c-divider);
+  background: var(--vp-c-bg-soft);
+}
+
+.code-example__body :deep(div[class*="language-"]) {
+  margin: 0;
+  border-radius: 0;
+}
+
+.code-example__body :deep(button.copy) {
```
mkucmus: VitePress already ships a copy button on every code block. Could we keep only the collapse part and drop the custom copy code? That would remove about half of the script.

apps/docs/src/components/CodeExample.vue
```diff
@@ -0,0 +1,211 @@
+<script setup>
+import { computed, onBeforeUnmount, onMounted, ref } from "vue";
+
+const props = defineProps({
+  title: { type: String, default: "Example" },
+  peek: { type: Number, default: 260 },
+  expanded: { type: Boolean, default: false },
+});
+
+const bodyId = computed(
```
mkucmus: Two examples with the same title on one page would get the same id, and `aria-controls` would point to the first one. Maybe add an instance counter or use `useId()`?

mkucmus (changes requested):
Nice recipe. The operation table, the type notes and the `useUser` await details all match the source.

A few suggestions inline. The main ones are the SSR explanation and the empty-cart state in the example.

One more small thing: `apps/docs` is outside the lint CI paths, so `pnpm format` was not enforced here. It would change trailing commas in `CodeExample.vue` and `cart.md`. Could you run it once before merge?

mkucmus (approved):
📓 🔍

patzick:
[Automation] PR looks good

apps/docs/src/frontends-recipes/checkout/cart.md
```diff
@@ -0,0 +1,443 @@
...
+Read the diagram from left to right:
+
+1. **UI** — the customer adds a product, changes a quantity or removes a line item.
+2. **Composable** — `useAddToCart` or `useCartItem` delegates to the matching `useCart` method, which calls one line item operation through `apiClient.invoke` with the current `sw-context-token`.
+3. **Store API** — the cart is recalculated in the current sales channel context and returned in full.
+4. **Shared state** — `useCart` replaces the shared `swCart` value with that response.
+5. **Errors** — any `errors` from the same response are merged into `swCartErrors`, for a notification layer to consume once.
+6. **UI** — components read `cartItems`, `count`, `subtotal` and `totalPrice` from composables instead of keeping their own copy.
+
+You do not need to call `refreshCart()` after a write. Use it on the initial page load, or after the customer session changes. `useUser().login()` and `logout()` do fire `refreshCart()` themselves, but they do not await it — only `register()` does — so code that renders prices straight after a session change should `await refreshCart()` itself.
```
patzick, line 121: `register()` does not call `refreshCart()` at all in `useUser.ts` — only `login()` and `logout()` do, and neither awaits it. This "only `register()` does" claim looks incorrect.

patzick (changes requested):
[Automation] PR looks good

PR #2753, opened 2026-09-20
feat(docs): recipe variants

### LINKED ISSUE #2752: Variants recipe

mdanilowicz:
closes #2752

This pull request adds a new "Catalog" section to the documentation, focusing on product catalog flows and, in particular, providing a comprehensive recipe for implementing product variant selection in Shopware-based frontends. The changes include new sidebar navigation, an overview page for the catalog section, and a detailed guide for building a variant selector, complete with a minimal Vue example, best practices, edge cases, and a testing checklist.

**Documentation Additions:**

* **Catalog Section Introduction and Navigation**
  - Added a new "Catalog" section to the sidebar and the main recipes index, including an overview page describing catalog-related flows [[1]](diffhunk://#diff-5f5e225b393b5ffe63e7a975023ec61058e4e44927a6ba22952f3e00dadf1b00R251-R261) [[2]](diffhunk://#diff-5845376f84de463f245b3ea76e14bed604bca8fca2d161d63a2d68d4b53d2413R17-R20) [[3]](diffhunk://#diff-1393ed99b04d96c88f791580447324ae5f5b3950b77acc6ec02c2bcb37598408R1-R11).

* **Product Variant Selection Recipe**
  - Introduced a comprehensive guide at `catalog/variants.md` detailing how to implement a product variant selector, including:
    - Step-by-step explanation of the variant resolution flow in Shopware Frontends.
    - Explanation of why variant resolution is a filtered product search, not a direct find-variant operation.
    - Details on composables, helpers, and Store API operations involved.
    - A minimal, annotated Vue example for building a variant selector.
    - Common pitfalls, edge cases, and a thorough testing checklist for robust implementations.

These updates significantly expand the documentation's coverage of catalog and variant selection patterns, providing frontend developers with practical guidance and best practices.


apps/docs/src/frontends-recipes/index.md
```diff
@@ -14,6 +14,10 @@ These pages are written for frontend developers who want to understand the Shopw
 
 <PageRef page="account/" title="Account" sub="Customer session and account flows such as login, logout, registration, account state, and the customer wishlist." />
 
+## Catalog
+
+<PageRef page="catalog/" title="Catalog" sub="Product catalog flows such as listings and filters, search and suggest, variant selection, price and tax state, product reviews, and cross-selling." />
```
patzick, line 19: **Overpromised scope in new Catalog index copy**

Both new index blurbs list six catalog topics — "listings and filters, search and suggest, variant selection, price and tax state, product reviews, and cross-selling" — but this PR only adds the `variants.md` recipe. Compare with the existing `Account`/`Checkout` index entries, where every topic named in the summary is actually covered by a page (or, for login/logout, within the single existing page's content). Here a reader following the Catalog link will find only variant selection; the other five topics have no page anywhere in this PR and there's no note that they're upcoming. Consider trimming the summary to what's actually shipped (variant selection) until the other recipes land.

apps/docs/src/frontends-recipes/catalog/variants.md
```diff
@@ -0,0 +1,354 @@
...
+  <SchemaTypeTooltip type-key='Schemas["ProductDetailResponse"]' />
+  <SchemaTypeTooltip type-key='Schemas["PropertyGroup"]' />
+  <SchemaTypeTooltip type-key='Schemas["PropertyGroupOption"]' />
+  <SchemaTypeTooltip type-key='Schemas["Product"]' />
+</div>
+
+```ts
+import type { Schemas, operations } from "#shopware";
+
+type ProductDetailResponse = Schemas["ProductDetailResponse"];
```
patzick, line 187: **Unverified claim about find-variant response shape**

This states the Store API returns the find-variant payload flat and instructs readers to read `variantId` from the response root instead of `response.data.foundCombination.variantId`. The generated type (`FindProductVariantRouteResponse` in `packages/api-client/api-types/storeApiTypes.d.ts`) nests it under `foundCombination`, and nothing in this repo (no composable, no test) exercises this operation to confirm the live API actually differs from its own generated schema. If the claim is wrong, a reader who follows this instruction gets `undefined` back for `variantId`. Worth double-checking against an actual Store API response before merging, since it contradicts the shipped type with no supporting evidence in-repo.

mdanilowicz: Double-checked against a live Store API, four calls: POST and GET on `/product/{productId}/find-variant`, with options as an array and as a groupId → optionId map. All four return the struct flat:

So `response.data.foundCombination.variantId` is the read that yields undefined — the page's instruction is the one that works. The apiAlias is the decisive bit: it is the alias of the FoundCombination struct itself, so the serialized response is that struct rather than an object wrapping it. The operation's own response description ("Returns an FoundCombination struct containing the ids matching the search") agrees; the foundCombination wrapper exists only in the hand-written schema annotation.

apps/docs/src/frontends-recipes/catalog/variants.md
```diff
@@ -0,0 +1,367 @@
...
+5. The response is one product carrying only the fields the `includes` clause allowed.
+6. The UI reads the selection from `getSelectedOptions` instead of keeping its own copy, and either navigates to the resolved variant or emits it into `changeVariant()`.
+
+You do not get a request from `handleChange` itself. The composable separates "the customer changed the selection" from "resolve what that selection means", and the second half is yours to trigger.
+
+`cms-base-layer` ships that second half: `SwVariantConfigurator` renders the groups and resolves the variant, with an `allowRedirect` prop that is `true` by default. Set it to `false` and the component emits `change` with the resolved variant instead of navigating. `CmsElementBuyBox` renders it with `@change="changeVariant"` but leaves the default in place, so out of the box the redirect always wins; that handler only runs in a project that renders `SwVariantConfigurator` itself with `:allow-redirect="false"`. `vue-starter-template` has no selector of its own — it extends `@shopware/cms-base-layer`, so a CMS-rendered product page gets that one as-is. Read it before writing your own.
```
patzick: At the default `allowRedirect=true`, `SwVariantConfigurator` sends the customer to the **home page** when nothing resolves. Worth stating here, since this paragraph tells readers to read the component before writing their own.

`findVariantForSelectedOptions()` returns `undefined` both on no match and on a swallowed error. Then:

- `getProductUrl(undefined)` returns `"/"` — `packages/helpers/src/product/getProductUrl.ts:18`
- so `getProductRoute(undefined)` is `{ path: "/", ... }`
- `buildUrlPrefix` returns that object, prefixed — `packages/helpers/src/cms/buildUrlPrefix.ts:22-28` — never falsy for this input

So the guard at `SwVariantConfigurator.vue:63` (`allowRedirect && selectedOptionsVariantPath`) collapses to just `allowRedirect`, and an unavailable combination does `router.push("/")` or `/<prefix>/`.

Two statements on the page read the other way round because of this:

- the Testing Checklist line "An unavailable combination renders a message, stays on the page, and can be submitted again" is true of your own example, not of the shipped component;
- the `@change` branch in `CmsElementBuyBox` is unreachable at the default, not merely outvoted by the redirect.

Minor, same function: the `try/catch` at `:64-68` is synchronous around a `router.push` that returns a promise, so the `NavigationFailure` you correctly warn about in Edge Cases is swallowed there too.

apps/docs/src/frontends-recipes/catalog/variants.md
```diff
@@ -0,0 +1,367 @@
...
+    - useProductSearch
+    - useUrlResolver
+  helpers:
+    - getTranslatedProperty
+    - getProductRoute
+    - buildUrlPrefix
+  operations:
+    - readProduct post /product
+    - readProductGet get /product
+    - readProductDetail post /product/{productId}
```
patzick, line 21: `readProductDetailGet get /product/{productId}` is missing from this list. The Request Flow table documents the detail load going out as GET under `cacheableReads`, and every other operation here is listed as a POST/GET pair. `catalog/search.md:22-23` lists both.

patzick (commented):
Re-review of `49e7ff5d`. Checked the composable, helper and `cms-base-layer` claims against source: the page is accurate — `getSelectedOptions` keying, `handleChange` issuing no request, the `parentId` + per-option `equals` criteria with `limit 1`, the `cacheableReads` GET branch, `isLoadingOptions` being write-once, and the `Object.assign` merge in `changeVariant` all check out. Links, nav positions, component props and every `SchemaTypeTooltip` type-key resolve.

The find-variant response-shape thread is settled from my side — the live-API evidence plus the `apiAlias` argument is convincing, and the page flags the contract mismatch explicitly rather than papering over it.

Two things left, both inline: one behaviour of the shipped `SwVariantConfigurator` that the page omits, and one missing frontmatter operation.

apps/docs/src/frontends-recipes/catalog/variants.md
```diff
@@ -0,0 +1,372 @@
...
+
+You do not get a request from `handleChange` itself. The composable separates "the customer changed the selection" from "resolve what that selection means", and the second half is yours to trigger.
+
+`cms-base-layer` ships that second half: `SwVariantConfigurator` renders the groups and resolves the variant, with an `allowRedirect` prop that is `true` by default. `vue-starter-template` has no selector of its own — it extends `@shopware/cms-base-layer`, so a CMS-rendered product page gets that one as-is. Read it before writing your own, and read it knowing what it does with a miss.
+
+At the default it navigates on every change, including the ones that resolve nothing. `getProductRoute(undefined)` carries `path: "/"`, and `buildUrlPrefix` always returns an object, so the component's `allowRedirect && path` guard never blocks anything: an unavailable combination sends the customer to the home page, or to `/<prefix>/` on a localised storefront. Pass `:allow-redirect="false"` and handle the miss yourself if that is not what you want.
```
patzick: The conclusion is right, the quoted guard is not. `SwVariantConfigurator.vue:63` reads `if (allowRedirect && selectedOptionsVariantPath)` — the guarded value is the **route object** returned by `buildUrlPrefix`, not a path string. Since `buildUrlPrefix` always returns an object (`buildUrlPrefix.ts:22-28`, fallback `{ path: "" }`), the guard is indeed never falsy, but a reader who greps for `allowRedirect && path` will not find it. Suggest naming the real identifier.

apps/docs/src/frontends-recipes/catalog/variants.md
```diff
@@ -0,0 +1,372 @@
...
+| Read the trimmed variant | `response.data.elements?.[0]`                                              | same request                                              | <SchemaTypeTooltip type-key='operations["readProduct post /product"]["response"]' />                                  |
+| Merge without navigating | `changeVariant(variant)`                                                   | none                                                      | <SchemaTypeTooltip type-key='Schemas["Product"]' />                                                                   |
+| Use the dedicated route  | `invoke("searchProductVariantIds post /product/{productId}/find-variant")` | `POST /product/{productId}/find-variant`                  | <SchemaTypeTooltip type-key='operations["searchProductVariantIds post /product/{productId}/find-variant"]["body"]' /> |
+
+The last row has no composable. It takes the selected options — as an array of option ids, or as a map keyed by **group id**, not by group name — and returns the found combination with the variant id, so a custom selector that only needs an id can avoid the product search entirely.
```
patzick: The group-id keying is correct, but it is not discoverable from the type this page sends readers to. `storeApiTypes.d.ts:12931` emits `options: string[] | { [key: string]: string }` — the "map of groupId => optionId" wording only exists in `api-types/storeApiSchema.json`. Worth saying the generated type does not encode it, otherwise a reader checks the `.d.ts` and assumes group name works.

apps/docs/src/frontends-recipes/catalog/variants.md
```diff
@@ -0,0 +1,372 @@
...
+type Product = Schemas["Product"];
+
+// what the find-variant route actually answers with
+type FoundCombination = {
+  variantId?: string;
+  options?: string[];
+};
+```
+
+`ProductDetailResponse` is where the configurator comes from: it is `{ product, configurator }`, and `configurator` is the `PropertyGroup[]` that `getOptionGroups` returns.
```
patzick: `configurator` is optional: `storeApiTypes.d.ts:3974` types `ProductDetailResponse` as `{ configurator?: PropertyGroup[]; product: Product }`. The flat `it is { product, configurator }` reads as both being guaranteed, which matters because the Edge Cases section already says a product without a configurator is a supported shape.

apps/docs/src/frontends-recipes/catalog/variants.md
```diff
@@ -0,0 +1,374 @@
...
+- `router.push` resolves with a `NavigationFailure` instead of throwing when a guard aborts the navigation or the target is the current route. Check the resolved value; only a guard that throws reaches a `catch`.
+- The variant search restricts the product to `id`, `translated`, `productNumber` and `seoUrls`. Passing that into `changeVariant` leaves price, stock, cover and every other field at the previous variant's values.
+- The search filters on `parentId`, so it only makes sense for a variant of a configurable product. A standalone product carries `parentId: null`, and `equals` on `null` matches every parentless product in the catalog. Such a product also has no `optionIds`, so nothing else narrows the filter and the search answers `200` with the first parentless product it finds — a product the customer never picked, which the shipped configurator would then navigate to. Check `product.parentId` before you resolve. A response that omits `parentId` entirely drops the key from the filter instead, and that is what earns a `400 FRAMEWORK__INVALID_FILTER_QUERY`.
```
patzick: **The unnarrowed-filter scenario is not reachable through the shipped configurator**

The `equals` on `null` → `IS NULL` correction is right, but the consequence attributed to it is not. `SwVariantConfigurator` only calls `findVariantForSelectedOptions()` from `onHandleChange`, which fires from an option click — and `handleChange` has already written that option into `selected` (`useProductConfigurator.ts:117-121`). So the filter the shipped component sends always carries at least one `optionIds` equals alongside `parentId IS NULL`; it never degenerates to "nothing else narrows the filter", and it cannot navigate to "the first parentless product it finds". On top of that, a product with no `optionIds` also has no configurator groups to render, so there is no option to click in the first place.

The unnarrowed filter is real, but only on the direct path: calling `findVariantForSelectedOptions()` yourself, with an empty selection, on a product whose `parentId` is `null`. Worth keeping the warning and the `Check product.parentId before you resolve` advice — just drop the `which the shipped configurator would then navigate to` clause and frame it as a custom-selector trap.

patzick (commented):
Re-review of `0c6a54ae`. Verified against source: `buildUrlPrefix` returning `{ path: "" }` at worst and the `selectedOptionsVariantPath` route-object guard (`packages/helpers/src/cms/buildUrlPrefix.ts`, `SwVariantConfigurator.vue:63`), `configurator?: PropertyGroup[]` being optional (also already optional in `useProduct`'s signature, so the example change is consistent), and the `options` / `switchedGroup` comment asymmetry (`storeApiTypes.d.ts:14600` vs `:14606`). `readProductDetailGet` is now in the frontmatter. One remaining inaccuracy inline.

patzick (commented):
Re-review of `cb59715d`. PR scope is now `sidebar.ts`, `catalog/index.md` and `catalog/variants.md` only — the recipes index blurb I flagged earlier came from #2771 and is no longer this PR's.

Both new paragraphs check out against source:

- The `change`-event / `resolveComponent` + `global: true` override note matches `SwVariantConfigurator.vue:15,59-67`, `CmsElementBuyBox.vue:159`, `composables/src/index.ts:83` and the `app/components` registration in `vue-starter-template/nuxt.config.ts`; the `overwriting-cms.html` link resolves.
- The simple-product edge case matches `useProductConfigurator.ts:80-93` (`parentId` always filtered, empty selection adds no `optionIds` filters) and `encodeForQuery.ts:10` (`JSON.stringify` drops the undefined-valued key → `400`).

Earlier open points: `readProductDetailGet get /product/{productId}` is now in the operations list, and the flat find-variant response is backed by your four live-API calls. No further findings from me. CI green.

patzick (commented):
Re-review of `ef0dc6cf`: the only change since `cb59715d` is a merge of `main`. The PR's own changes (`sidebar.ts` entry, `catalog/index.md`, `catalog/variants.md`) are unchanged. No new findings. CI green.

patzick (approved):

PR #2755, opened 2026-09-20
feat(docs): recipe pricing

### LINKED ISSUE #2754: Pricing recipe

mdanilowicz:
closes #2754

This pull request adds comprehensive documentation for handling product catalog pricing and tax state in the Shopware frontend, including a new "Catalog" section in the docs. It also updates schema update scripts to use the correct directory and improves navigation within the documentation.

**Documentation Additions and Improvements:**

* Added a new "Catalog" section to the documentation sidebar and landing page, organizing recipes related to the product catalog such as listings, filters, search, variants, prices, reviews, and cross-selling. [[1]](diffhunk://#diff-5f5e225b393b5ffe63e7a975023ec61058e4e44927a6ba22952f3e00dadf1b00R251-R261) [[2]](diffhunk://#diff-5845376f84de463f245b3ea76e14bed604bca8fca2d161d63a2d68d4b53d2413R17-R20)
* Introduced a detailed "Prices and Tax State" recipe (`prices.md`) covering best practices for rendering product prices, handling tax state, and using Shopware composables for correct price display. The guide includes a flow diagram, code examples, edge cases, and a testing checklist.
* Created an index page for the "Catalog" section, summarizing its contents and linking to the "Prices and Tax State" recipe.

**Build and Script Updates:**

* Updated schema update scripts in `package.json` to copy API schema files into the `src/.data/schema/` directory instead of `.data/schema/`, ensuring schema files are placed in the correct location for the docs build.


apps/docs/src/frontends-recipes/catalog/prices.md
```diff
@@ -0,0 +1,338 @@
...
+      "Prices arrive already calculated for the current context. taxState decides whether the numbers are gross or net, and nothing in the frontend converts between them.",
+    code: "GET /store-api/context",
+    state: "sw-context-token",
+    typeKeys: ['operations["readContext get /context"]["response"]'],
+  },
+  {
+    title: "Shared state",
+    action: "Seed the formatter",
+    detail:
+      "On the client usePrice is wrapped in createSharedComposable, so the first caller creates the instance every other component reuses. During SSR that wrapper bails out and every call site gets its own instance, so no root initialization is required on either side.",
```
patzick, line 47: `createSharedComposable` (from `@vueuse/shared`) has no SSR bail-out — no `isClient` check anywhere in its source. It shares the same module-scoped `scope`/`state` regardless of environment. "During SSR that wrapper bails out and every call site gets its own instance" is incorrect.

apps/docs/src/frontends-recipes/catalog/prices.md
```diff
@@ -0,0 +1,338 @@
...
+
+`useProductPrice` is the one you reach for most. It takes a `Ref<Product | undefined>` and derives everything else from it:
+
+- **The price** — `price` (the whole chosen `CalculatedPrice`), `unitPrice`, `totalPrice`.
+- **Ranges** — `displayFrom`, `displayFromVariants`, `tierPrices`.
+- **Extras** — `referencePrice`, `regulationPrice`, `hasListPrice`, and its deprecated alias `isListPrice`.
+
+Four things the generated reference will not tell you:
+
+- `usePrice` is wrapped in `createSharedComposable` from `@vueuse/core`, and that wrapper is **client-only** — its source begins `if (!isClient) return composable`. During SSR every call site gets its own instance and reads its own arguments; only in the browser is there a single instance to share.
```
patzick, line 149: Same issue: `createSharedComposable`'s source does not begin with `if (!isClient) return composable` — that check doesn't exist. Confirmed by fetching `@vueuse/shared`'s actual source. This same false claim is repeated at lines ~280 and ~289 as well; all should be corrected or removed since several downstream conclusions (SSR getting per-call-site instances, args being "inert only on the client") rely on it.

patzick (changes requested):
Verified the technical claims against source. Most check out, but one is fabricated and repeated 4x in this doc — see inline comments.

apps/docs/src/frontends-recipes/catalog/prices.md
mdanilowicz: Thanks — I checked this carefully, but I don't think it holds for the version we pin.

`createSharedComposable` does have the SSR bail-out in **@vueuse/shared 14.4.0**, which is what this repo resolves (`pnpm-workspace.yaml` pins `@vueuse/*` to `14.4.0` as an override, and the lockfile has exactly one `@vueuse/shared` entry). All three shipped builds contain it — including the minified one, as `function j(e){if(!h)return e;…}`.

Rather than rely on reading the bundle, I ran the actual resolved module in a no-DOM context:

```
window defined? false
composable body ran: 2 time(s)
a === b (shared instance)? false
a.code: EUR | b.code: USD
=> second call's argument honoured? true
```

The body ran twice, the instances differ, and the second call's argument was honoured. Without the guard this would be `ran: 1`, `a === b`, and `b.code === "EUR"`.

Upstream agrees too — `vueuse/vueuse@main`, `packages/shared/createSharedComposable/index.ts`:

```ts
if (!isClient)
  return composable

let subscribers = 0
```

I think the discrepancy is a version one: the guard landed in 14.x. The same file at tag **v13.9.0** starts straight at `let subscribers = 0`. Worth noting this repo also contains committed `.output` build artifacts with vueuse **13.9.0** inlined, and a filename search hits those first — I made exactly that mistake myself earlier while verifying this page.

To confirm on your side:

```bash
node -e 'const s=require("fs").readFileSync("node_modules/.pnpm/@vueuse+shared@14.4.0_vue@3.5.42_typescript@5.9.3_/node_modules/@vueuse/shared/dist/index.js","utf8");const i=s.indexOf("function createSharedComposable");console.log(s.slice(i,i+120))'
```

If that prints a body without `if (!isClient)`, you're resolving a different version than the lockfile — happy to look at your `pnpm why @vueuse/shared` in that case. Otherwise I'd leave the claim as it stands.

apps/docs/package.json
```diff
@@ -10,8 +10,8 @@
     "docs:preview": "../../../developer-portal/docs-cli.cjs preview",
     "docs:cli": "../../../developer-portal/docs-cli.cjs",
     "schema:update": "pnpm run schema:update:store && pnpm run schema:update:admin",
-    "schema:update:admin": "cp ../../packages/api-client/api-types/adminApiSchema.json .data/schema/adminApiSchema.json",
-    "schema:update:store": "cp ../../packages/api-client/api-types/storeApiSchema.json .data/schema/storeApiSchema.json"
+    "schema:update:admin": "cp ../../packages/api-client/api-types/adminApiSchema.json src/.data/schema/adminApiSchema.json",
+    "schema:update:store": "cp ../../packages/api-client/api-types/storeApiSchema.json src/.data/schema/storeApiSchema.json"
```
patzick: why do we do this?

mdanilowicz: The docs app doesn't generate the API schema itself, and it isn't built from this repo - `docs:link` ships only `apps/docs/src` to `developer-portal`, so `packages/api-client` isn't reachable at build time. Copying the schema under `src/` is what makes it visible to the `VitePress` data loader. It's improvable, but that belongs on the documentation app side. Change for the future

apps/docs/src/frontends-recipes/catalog/prices.md
```diff
@@ -0,0 +1,347 @@
...
+
+## Edge Cases
+
+- `getFormattedPrice(undefined)` returns an empty string, and with no `currencyCode` or `currencyLocale` set it returns the raw value as a string. In practice `currencyLocale` is never empty, so the unformatted case only happens before the session context has a currency.
+- `createSharedComposable` is client-only. On the server `usePrice` is the unwrapped composable, so a `usePrice({ currencyCode, localeCode })` that looks inert because "only the first call counts" does take effect in that component's SSR output — and is then dropped at hydration, producing exactly the format mismatch described below.
+- On the client the shared instance is destroyed once its last subscriber unmounts. A `localeCode` set on a product page does not survive a visit to a page that renders no prices; the next instance falls back to `browserLocale`. If you need a session-wide locale, subscribe from somewhere that never unmounts.
+- `usePrice`'s parameter object types `currencyCode` as required, so `usePrice({ localeCode })` does not compile. You have to pass a `currencyCode` too, and the immediate watcher overwrites it from the context on the next tick.
```
patzick: "on the next tick" isn't accurate: a `watch` with `immediate: true` runs its callback synchronously during setup, so `currencyCode` is overwritten right away (when `sessionContext.currency` is already set).

apps/docs/src/frontends-recipes/catalog/prices.md
```diff
@@ -0,0 +1,347 @@
...
+
+`taxState` lives on the context under `context.taxState` and is typed as a plain, optional `string` in the schema. It is not something the frontend sets directly: it follows the customer group and the context, and a change to it invalidates every price currently rendered.
+
+## Edge Cases
+
+- `getFormattedPrice(undefined)` returns an empty string, and with no `currencyCode` or `currencyLocale` set it returns the raw value as a string. In practice `currencyLocale` is never empty, so the unformatted case only happens before the session context has a currency.
+- `createSharedComposable` is client-only. On the server `usePrice` is the unwrapped composable, so a `usePrice({ currencyCode, localeCode })` that looks inert because "only the first call counts" does take effect in that component's SSR output — and is then dropped at hydration, producing exactly the format mismatch described below.
```
patzick: "then dropped at hydration" only holds if another component already created the shared instance on the client. If this component is the first caller, its arguments are honoured on the client too (as line 317 correctly says). Suggest qualifying it.

apps/docs/src/frontends-recipes/catalog/prices.md
patzick: Confirmed: `@vueuse/shared@14.4.0` has the `if (!isClient) return composable` guard. My comment was wrong, the claim is fine.

patzick (approved):

PR #2759, opened 2026-09-20
 feat(docs): recipe  cross selling

### LINKED ISSUE #2758: Cross selling recipe

mdanilowicz:
closes #2758

This pull request introduces a new "Catalog" section to the documentation, focusing on product catalog flows and, in particular, a comprehensive recipe for handling cross-selling groups on a product page. The main additions include sidebar navigation updates, a new index page for catalog recipes, and a detailed guide on implementing and understanding the cross-selling API and composable.

**Documentation Structure Updates:**

* Added a "Catalog" section to the sidebar navigation, including a link to the new cross-selling recipe (`apps/docs/.vitepress/sidebar.ts`).
* Introduced a catalog index page with navigation metadata and a reference to the cross-selling recipe (`apps/docs/src/frontends-recipes/catalog/index.md`).
* Updated the main recipes index to include a summary link to the new catalog section (`apps/docs/src/frontends-recipes/index.md`).

**New Cross-Selling Recipe:**

* Added a comprehensive guide for cross-selling groups, covering the composable API (`useProductAssociations`), request/response structure, Vue usage, edge cases, and common mistakes. The guide explains why certain composable options are ignored, how the API response is structured, and when not to use the composable (e.g., on CMS-driven product pages) (`apps/docs/src/frontends-recipes/catalog/cross-selling.md`).


patzick:
[Automation] The `lint` check (semantic PR title) is failing — title has a leading space: " feat(docs): recipe  cross selling". Please fix the title (e.g. `feat(docs): add cross-selling recipe").

apps/docs/src/frontends-recipes/catalog/cross-selling.md
```diff
@@ -0,0 +1,311 @@
...
+
+<template>
+  <p role="status">{{ isLoading ? "Loading recommendations…" : "" }}</p>
+
+  <section v-for="group in groups" :key="group.crossSelling.id">
+    <h2>{{ getTranslatedProperty(group.crossSelling, "name") }}</h2>
+
+    <ul>
+      <li v-for="crossSellProduct in group.products" :key="crossSellProduct.id">
+        <NuxtLink :to="getProductRoute(crossSellProduct)">
```
patzick: `getProductRoute` returns an **unprefixed** route, so on a localized storefront this sends a customer browsing `/de-DE` to the default-locale URL.

The sibling recipe already on `main` documents the opposite rule (`catalog/search.md:367`): _"Do not link with a bare `getProductRoute`. Wrap it in `formatLink` or a localised storefront drops the prefix."_ — and `search.md:281` shows the pattern.

Suggest wrapping it:

```suggestion
        <NuxtLink :to="formatLink(getProductRoute(crossSellProduct))">
```

This also needs the composable in `<script setup>` (near line 201):

```ts
const { formatLink } = useInternationalization(localePath);
```

(`formatLink` only prefixes when created with a path resolver — see `context/language-and-currency.md:144`.)

Worth mirroring in **Common Mistakes** too, since `search.md` lists it there.

patzick (changes requested):
[Automation] One finding — the Vue example links with a bare `getProductRoute`, which the sibling `search.md` recipe explicitly warns against. Everything else checks out: all code claims verified against source, all links/components/type-keys resolve, conventions match siblings, CI green.

apps/docs/src/frontends-recipes/catalog/cross-selling.md
```diff
@@ -0,0 +1,320 @@
...
+
+```vue
+<script setup lang="ts">
+import { getProductRoute, getTranslatedProperty } from "@shopware/helpers";
+
+import type { Schemas } from "#shopware";
+
+const { product } = defineProps<{ product: Schemas["Product"] }>();
+
+const { formatLink } = useInternationalization(localePath);
```
patzick, line 210: `localePath` is never defined here, so this is the previous finding again in a different shape.

`formatLink` returns the link unchanged when no resolver was passed (`useInternationalization.ts:134` — `if (!pathResolver) return link;`), which the recipe itself states at line 253. Without `useLocalePath()` the example still sends a `/de-DE` customer to the default-locale URL — and as a bare identifier `localePath` is not auto-imported, so it is also a `ReferenceError` if copied as-is.

The sibling `search.md:198-199` and every `vue-starter-template` call site declare it first (`app/components/layout/Header.vue:23-24`, `layout/Footer.vue:5-6`, `account/LoginForm.vue:3-4`, …):

```ts
const localePath = useLocalePath();
const { formatLink } = useInternationalization(localePath);
```

Suggest adding the missing line so the example matches the rule the Edge Cases and Common Mistakes sections state.

patzick (changes requested):
[Automation] The `formatLink` fix is in, but the example never creates the path resolver it needs — so it is still a no-op. One finding inline; everything else re-verified clean (all composable/Store API claims confirmed against source, all 11 links resolve, frontmatter + headings + sidebar consistent with siblings, CI green).

apps/docs/src/frontends-recipes/catalog/cross-selling.md
```diff
@@ -0,0 +1,320 @@
...
+type CrossSellingConfig = Schemas["ProductCrossSelling"];
+```
+
+`CrossSellingCollection` is declared as an array in the schema, so `productAssociations` is iterable directly. Most Store API list responses wrap their rows in an `elements` key; this one, like `readNavigation` and `readBreadcrumb`, does not.
+
+## Minimal Vue Example
+
+<CodeExample title="Cross-selling groups on a product page">
+
+```vue
```
patzick, line 201: btw/ it would be great to add this somewhere as an example (or create example project just for docs) and use auto-md so we always have compilable code and no old code templates

patzick: [Automation] Tracked as a follow-up: https://github.com/shopware/frontends/issues/2776

patzick:
[Automation] Both previous findings are resolved at `7473635d`. No new issues.

**Re-verified against source:**
- `useInternationalization(pathResolver?: (path: string) => string)` — the example's two-line `useLocalePath()` + `useInternationalization(localePath)` pattern matches the real signature (`useInternationalization.ts:81`) and real call sites (`Search.vue:40`, `layouts/account.vue:7`). `useLocalePath` is auto-imported by `@nuxtjs/i18n`, no explicit import anywhere in the repo.
- `formatLink` no-op claim matches `useInternationalization.ts:134`.
- `useProductAssociations`: throws on empty ref (`:38`), local `ref([])` (`:43`), always POST (`:83`), swallows errors + `finally` (`:96-103`), `sw-include-seo-urls` only via `includeSeoUrls` (`:86`), `params` declared-but-required and discarded (`:12`, `:49`).
- Store API: both cross-selling operations declare no request body, only `productId` + two optional headers; the GET variant has no `_criteria` (`storeApiTypes.d.ts:12867-12897`). `CrossSellingElementCollection = CrossSellingElement[]` (`:1520`).
- All `SchemaTypeTooltip` type-keys resolve; all 11 links resolve; frontmatter/`nav: position: 40`/sidebar/`index.md` consistent with siblings.

Not approving from automation — a human still needs to dismiss the earlier change request.

PR #2763, opened 2026-09-20
feat(docs): recipe language and currency

### LINKED ISSUE #2762: Language and currency recipe

mdanilowicz:
closes #2762

This pull request adds a new "Context" section to the frontend recipes documentation, with a comprehensive guide and example on implementing language and currency switchers. The new documentation explains the differences between switching language (which triggers a redirect) and switching currency (which is a reactive context patch), provides a minimal Vue example, and covers common pitfalls and edge cases.

**Documentation Additions:**

* Added a "Context" group to the sidebar and index, with a landing page and a detailed recipe for language and currency switching. [[1]](diffhunk://#diff-5f5e225b393b5ffe63e7a975023ec61058e4e44927a6ba22952f3e00dadf1b00R270-R280) [[2]](diffhunk://#diff-5845376f84de463f245b3ea76e14bed604bca8fca2d161d63a2d68d4b53d2413R20-R23) [[3]](diffhunk://#diff-c6a8a5ebb4b6d1792fc43cd80185118af20f88400e4551e7704cb51e1bae5c8cR1-R11)

**New Recipe: Language and Currency Switching**

* Introduced `language-and-currency.md`, a thorough guide covering:
  - The difference between language and currency switching flows, including API calls and frontend handling.
  - A minimal, production-ready Vue example for both switchers, with accessibility and error handling best practices.
  - Explanations of relevant composables, types, and session/context state management.
  - Common mistakes and a testing checklist to ensure robust implementations.
  - Edge cases, such as redirects, context refreshes, and schema quirks.

These updates significantly improve the documentation for frontend developers working with Shopware's context features, making it easier to implement robust language and currency switching.


apps/docs/src/frontends-recipes/context/language-and-currency.md
```diff
@@ -0,0 +1,395 @@
...
+  replaceToDevStorefront,
+} = useInternationalization();
+const {
+  currentLanguageId,
+  currency: currentCurrency,
+  setCurrency,
+} = useSessionContext();
+
+const languagesQuery = useAsyncData("languages", () => getAvailableLanguages());
+const currenciesQuery = useAsyncData("storeCurrencies", async () => {
```
mkucmus: Small idea here: these keys are the same as the starter's, but the fetch functions are not. In Nuxt 4.5 that logs `NUXT_E3004` "different handler" in dev, and only the starter's function runs. Could the example use its own keys?
```suggestion
const languagesQuery = useAsyncData("switcher-languages", () => getAvailableLanguages());
const currenciesQuery = useAsyncData("switcher-currencies", async () => {
```

apps/docs/src/frontends-recipes/context/language-and-currency.md
```diff
@@ -0,0 +1,395 @@
...
+
+The shared list is assigned from the query **result**, not from inside the handler. `getAvailableLanguages()` fills `swLanguages` as a side effect, but `useAsyncData` replays its payload instead of re-running the handler on the client, so that side effect never happens during hydration — the options would vanish the moment Vue takes over. `vue-starter-template` does the same thing at `app.vue:83` for the same reason.
+
+The example owns its initial load. `vue-starter-template` already fetches both lists in `app.vue` and `useCurrencySwitcher`, and these `useAsyncData` keys match its own, so inside that template the calls dedupe rather than fetching twice on hydration.
```
mkucmus: Nice point about the hydration side effect! Two small things here. Shared keys don't dedupe: the server fetches again, and hydration skips the fetch either way. And the example reads `currentLanguageId` and `currency` without loading the context, so readers outside the starter may get empty selects. What about wording like the cart recipe's?
```suggestion
The example owns its initial load. `vue-starter-template` already loads both lists, in `app.vue` and `useCurrencySwitcher`, so inside that template drop these two `useAsyncData` calls. Read `languages` from `useInternationalization()` and the currencies from `useCurrencySwitcher()` instead.

The example expects the session context to be loaded already, as `app.vue` in `vue-starter-template` does with `readContext get /context` and `useSessionContext(data)`. Without it, neither select shows the current value.
```

mkucmus (changes requested):
Thanks, this is a really useful recipe! The claims I checked all match the source. One thing to fix before merge: the example shares `useAsyncData` keys with the starter, and the text says this dedupes the fetch. It doesn't. There's also a small note about the session context. Details and suggestions are inline.

mkucmus (approved):
Thanks for the fixes, this looks good now! Own keys, the Nuxt behaviour described correctly, and the session context part goes further than I asked. I checked that the module really provides only an empty ref.

PR #2767, opened 2026-09-20
feat(docs): recipe url resolving

### LINKED ISSUE #2766: Url resolver recipe

mdanilowicz:
closes #2766

This pull request adds new documentation for "Context" in the frontend recipes section, focusing on URL resolving and SEO URLs, and updates the caching best practices documentation to clarify which composables use GET or POST requests and why. The sidebar navigation is updated to include the new "Context" section.

**Documentation Additions and Improvements:**

* Added a new "Context" section with a landing page (`index.md`) in the frontend recipes, introducing topics like language, currency, navigation, and URL resolving, and linking to a new guide on URL resolving and SEO URLs.
* Updated the sidebar configuration (`sidebar.ts`) to include the new "Context" section and its URL resolving guide for easier navigation.

**Caching Documentation Updates:**

* Clarified in `caching.md` which composables gain a GET branch when `cacheableReads` is enabled, including `useProductReviews` and `useCategorySearch.search`, and explained the exception for `useCategorySearch.search` regarding GET/POST usage.


apps/docs/src/frontends-recipes/context/url-resolving.md
```diff
@@ -0,0 +1,387 @@
...
+
+if (!canonicalRedirectTarget && !seoResult.value?.foreignKey) {
+  throw createError({
+    statusCode: 404,
+    statusMessage: `No data fetched from API for ${routePath}`,
+  });
+}
+
+const { routeName, foreignKey } = useNavigationContext(
+  ref(canonicalRedirectTarget ? null : seoResult.value),
```
mkucmus, line 289: Helpful to show the full page. This does not typecheck: `useAsyncData` gives `Ref<T | undefined>`, and `useNavigationContext` wants `Ref<SeoUrl | null>`. `Ref` is invariant, so it fails with TS2345. The starter adds a cast. A `?? null` is simpler:

```suggestion
  ref(canonicalRedirectTarget ? null : (seoResult.value ?? null)),
```

apps/docs/src/best-practices/caching.md
```diff
@@ -99,18 +99,19 @@ Exactly these composables gain a GET branch when `cacheableReads` is enabled:
...
 
-A few read paths intentionally stay on POST because the generated Store API schema does not type the `_criteria` parameter on their GET route:
+A few read paths stay on POST because the generated Store API schema does not type the `_criteria` parameter on their GET route. `useCategorySearch.search` is the exception that proves the rule: it already uses `readCategoryGet get /category/{navigationId}` and declares `_criteria` through a local intersection type until the generated types catch up.
 
 - `useListing` (product listing) - always `readProductListing post /product-listing/{categoryId}`
-- `useCategorySearch.search` (single category) - always `readCategory post /category/{navigationId}`
 - `useLandingSearch` - always `readLandingPage post /landing-page/{landingPageId}`
 
 As those GET schemas gain `_criteria` typing upstream, these reads can migrate too. Product listing is the first: Shopware core [PR #17204](https://github.com/shopware/shopware/pull/17204) declared `_criteria` on `GET /store-api/product-listing` (released in 6.7.12.0), so `useListing` can switch to the cacheable GET variant once the Store API types are regenerated against that schema.
```
mkucmus: Thanks for linking the core PR. The types are already regenerated: `packages/api-client/api-types/storeApiTypes.d.ts` declares `_criteria` on `readProductListingGet get /product-listing/{categoryId}`. So `useListing` can move today, not later.

```suggestion
As those GET schemas gain `_criteria` typing upstream, these reads can migrate too. Product listing is already unblocked: Shopware core [PR #17204](https://github.com/shopware/shopware/pull/17204) declared `_criteria` on `GET /store-api/product-listing` (released in 6.7.12.0), and the checked-in types carry it, so `useListing` can move to `readProductListingGet` today.
```

apps/docs/src/best-practices/caching.md
```diff
@@ -99,18 +99,19 @@ Exactly these composables gain a GET branch when `cacheableReads` is enabled:
...
+- `useCategorySearch.search` (single category)
 
 ::: tip
 The flag is a blanket GET/POST switch per composable, not a runtime authentication check. Even account-related lookups such as `useUser.loadCountry`/`loadSalutation` use GET when the flag is on. "Anonymous" here means the data is public reference or catalog data suitable for shared HTTP caching, not that the code inspects the login state. Whether a response is actually cached, and how it is scoped per user, is governed by the Shopware backend cache rules and your CDN configuration.
 :::
 
 ### Which reads stay on POST, and why
 
-A few read paths intentionally stay on POST because the generated Store API schema does not type the `_criteria` parameter on their GET route:
+A few read paths stay on POST because the generated Store API schema does not type the `_criteria` parameter on their GET route. `useCategorySearch.search` is the exception that proves the rule: it already uses `readCategoryGet get /category/{navigationId}` and declares `_criteria` through a local intersection type until the generated types catch up.
```
mkucmus: Useful section. "already uses" sounds unconditional, but the GET route only runs when `cacheableReads` is on, and it is off by default. The new url-resolving page lists `POST /category/{navigationId}` as the default for the same call, so the two pages disagree. This call is also in the GET list at L104.

```suggestion
A few read paths stay on POST because the generated Store API schema does not type the `_criteria` parameter on their GET route. `useCategorySearch.search` shows the way around it: with `cacheableReads` enabled it calls `readCategoryGet get /category/{navigationId}` and declares `_criteria` through a local intersection type until the generated types catch up.
```

apps/docs/src/frontends-recipes/context/url-resolving.md
```diff
@@ -0,0 +1,387 @@
...
+- `useNavigationContext(context)` snapshots what you pass it. `useContext` stores `ref(unref(context))`, so a `computed` handed to it is read once rather than tracked — which is what you want for a per-navigation resolution, and a trap if you expect it to follow a later change.
+- `useNavigationContext` issues no requests. It provides the `navigation` injection and exposes `navigationContext`, `routeName` and `foreignKey`; `foreignKey` falls back to `""`, never `undefined`.
+- `useCategorySearch` has two methods and they are not symmetrical. `search(categoryId, options)` sends `sw-include-seo-urls: true`; `advancedSearch({ query })` does not send that header at all.
+- `useLandingSearch().search` passes `cmsAssociations.associations` as the `associations` body field, while `useCategorySearch().search` passes the whole `cmsAssociations` object. The two are not interchangeable if you build a request by hand.
+- `useUrlResolver().resolveUrl(url)` does more than prefix: it drops the first path segment before re-joining, and it throws `URL Input too long` for input over 2083 characters. `getUrlPrefix()` reads an injected `urlPrefix` that the application provides, not the composable.
```
mkucmus, line 177: Good catch on `urlPrefix`. On the first part: `split("/").slice(1)` drops the empty string from the leading slash, not a path segment. The test pins it - `resolveUrl("/en/navigation/123")` returns `/shop/en/navigation/123`, so `en` stays. Same wording at L345.

```suggestion
- `useUrlResolver().resolveUrl(url)` does more than prefix: it splits on `/` and drops the first element before re-joining, which strips the leading slash rather than a path segment, and it throws `URL Input too long` for input over 2083 characters. `getUrlPrefix()` reads an injected `urlPrefix` that the application provides, not the composable.
```

apps/docs/src/frontends-recipes/context/url-resolving.md
```diff
@@ -0,0 +1,387 @@
...
+Pick by scope — how much of the resolution the composable is about:
+
+| Composable             | Scope                       | Reach for it when                                                 |
+| ---------------------- | --------------------------- | ----------------------------------------------------------------- |
+| `useNavigationSearch`  | any path                    | turning a URL into a `SeoUrl`                                     |
+| `useNavigationContext` | the resolved `SeoUrl`       | reading `routeName` or `foreignKey` anywhere below the route      |
+| `useCategorySearch`    | one category, or a criteria | building the category page a `frontend.navigation.page` points at |
+| `useLandingSearch`     | one landing page            | building the `frontend.landing.page` component                    |
+| `useUrlResolver`       | one CMS link                | rendering author-written HTML that contains internal links        |
+| `useBreadcrumbs`       | the global trail            | a page component that owns its own breadcrumbs                    |
```
mkucmus: Small wording thing: L178 just below says it is scoped, not global, and the source agrees. Readers who only scan the table may put it in a layout, which breaks the per-page trail.

```suggestion
| `useBreadcrumbs`       | the trail for its subtree   | a page component that owns its own breadcrumbs                    |
```

apps/docs/src/frontends-recipes/context/url-resolving.md
```diff
@@ -0,0 +1,387 @@
...
+  },
+  {
+    title: "Page",
+    action: "Fetch the entity",
+    detail:
+      "The page component receives the foreign key and fetches the category or the landing page, with the CMS associations that a rendered page needs, and loads its own breadcrumbs.",
+    code: "useCategorySearch().search(foreignKey, { withCmsAssociations: true })",
+    state: "category or landing page",
+    typeKeys: [
+      'operations["readLandingPage post /landing-page/{landingPageId}"]["response"]',
```
mkucmus, line 104: The type chips are a nice touch. This step calls `useCategorySearch`, so the landing page type looks like a slip. `readCategory post /category/{navigationId}` exists and resolves:

```suggestion
      'operations["readCategory post /category/{navigationId}"]["response"]',
```

mkucmus (changes requested):
Thanks for this recipe, it is a genuinely useful page. The flow diagram and the type chips work well.

I found six things I would fix before merge. Four are factual: the main example does not typecheck, and three claims do not match the source. Two are one-line slips.

Nothing structural. The shape of the page is good.

patzick:
[Automation] PR looks good

Fact-checked every technical claim in the changed lines against source at `f8485fd5` — all confirmed, no findings.

- `caching.md`: `useProductReviews`/`useCategorySearch.search` GET branches, the `readCategoryGet` missing-`_criteria` workaround, `readProductListingGet` already declaring `_criteria`, `searchPage` flattened GET params, `useListing`/`useLandingSearch` POST-only — all correct, and the GET-branch list is complete.
- `url-resolving.md`: `resolvePath` branches and filters, `useNavigationContext` snapshot/`foreignKey` fallback, the asymmetric `cmsAssociations` handling, `useUrlResolver` prefix + 2083 guard, scoped `useBreadcrumbs`, all operation ids and schema keys — all correct.
- Example + edge cases: matches the shipped `[...all].vue` (including the `history.state?.routeName`-only shortcut), the helpers in `getRouteFromPathInfo.ts`, `getProductRoute`/`getCategoryRoute` emitting an undefined `foreignKey`, the `global: true` component registration and `provide("urlPrefix")`. The snippet's `ref((canonicalRedirectTarget ? null : seoResult.value) ?? null)` satisfies `useNavigationContext(context?: Ref<Schemas["SeoUrl"] | null>)`, so the earlier typecheck concern is resolved.

CI is green. One thing to handle before merge: the branch is currently conflicting with `main` in `sidebar.ts`, `frontends-recipes/context/index.md` and `context/language-and-currency.md`.

PR #2771, opened 2026-09-21
feat(docs): recipe contact form

### LINKED ISSUE #2770: Contact form recipe

mdanilowicz:
closes #2770

This pull request adds new documentation for CMS-related frontend recipes, specifically focusing on rendering CMS pages and implementing a contact form. It also updates navigation to include the new CMS section.

**Documentation additions:**

* Added a new `CMS` section with an overview and a reference to the Contact Form recipe in `cms/index.md`.
* Introduced a link to the new CMS section in the main `frontends-recipes` index page.

**Navigation updates:**

* Updated the sidebar in `sidebar.ts` to include a collapsible `CMS` section with a link to the Contact Form recipe.


patzick:
[Automation] PR looks good

apps/docs/src/frontends-recipes/cms/contact-form.md
```diff
@@ -0,0 +1,496 @@
...
+    <form v-else @submit.prevent="submit">
+      <p v-if="formError" role="alert">{{ formError }}</p>
+
+      <label for="salutation">Salutation</label>
+      <select
+        id="salutation"
+        v-model="form.salutationId"
+        :disabled="!getSalutations.length"
+      >
+        <option value="">Not specified</option>
```
mkucmus: The backend always requires `salutationId`. It is hardcoded in `ContactFormValidationFactory`, but the schema does not list it. I tested it with curl on a local devenv:
- no `salutationId` or `""` → `400` on `/salutationId`
- the `not_specified` salutation → `200`

Here the select starts on `""`, so the submit fails. There is no error line for `/salutationId`, so the customer sees nothing. Could we preselect the `not_specified` salutation and show the error for the select?
```ts
const notSpecifiedId = computed(
  () => getSalutations.value.find((s) => s.salutationKey === "not_specified")?.id ?? "",
);
watch(notSpecifiedId, (id) => {
  if (!form.salutationId) form.salutationId = id;
}, { immediate: true });
```

apps/docs/src/frontends-recipes/cms/contact-form.md
```diff
@@ -0,0 +1,496 @@
...
+
+## Goal
+
+Build a contact form that submits to the Store API. The important part is what the schema warns about in its own description: only three fields are required there, and the shop's settings can require more — so the validation you write on the client is never the whole truth.
+
+## Shopware Flow
+
+`sendContactMail post /contact-form` has no composable wrapper. You call it through `apiClient.invoke`, and its response is `200` with no body at all — no reference number, no echo of the submission.
+
+The operation's description says outright that "there can be more required fields, depending on the system settings". Its own `required` list is just `email`, `subject` and `comment`; `firstName`, `lastName` and `phone` are each documented as "may be required depending on the system settings". A rejection for a field your form treated as optional is therefore a normal outcome, not a bug in your validation.
```
mkucmus: By default, seven fields are required, not three. `salutationId` is always required. `firstName`, `lastName` and `phone` are settings, and they are on by default. With only `email`, `subject` and `comment`, I got a `400` for four fields. Lines 93, 111, 436 and 455 say the same "three fields".
```suggestion
The operation's description says outright that "there can be more required fields, depending on the system settings". Its own `required` list is just `email`, `subject` and `comment`, but the backend also always requires `salutationId`, which the schema leaves out. `firstName`, `lastName` and `phone` are switches under Settings → Basic information, set per sales channel, and all three are on by default. So an unchanged shop requires seven fields, not three.
```

apps/docs/src/frontends-recipes/cms/contact-form.md
```diff
@@ -0,0 +1,496 @@
...
+| Submit the form         | `invoke("sendContactMail post /contact-form", { body })` | `POST /contact-form` | <SchemaTypeTooltip type-key='operations["sendContactMail post /contact-form"]["body"]' />  |
+| Read the result         | none                                                     | `POST /contact-form` | none — the operation answers `200` with no body                                            |
+
+The salutation row has a second variant, and it is the one you will actually hit. With `cacheableReads` enabled, `useSalutations` calls <SchemaTypeTooltip type-key='operations["readSalutationGet get /salutation"]["response"]' /> over `GET /salutation` instead, so the list is HTTP-cacheable. The library default is off, but `vue-starter-template` ships `cacheableReads: true`, so in the supported starting point the salutation list arrives over `GET`. The payload is the same either way, and nothing in your form code changes.
+
+The body also declares `slotId`, `cmsPageType` and `entityName` for resolving the slot configuration server-side. The component shipped in the CMS base layer sends only `navigationId`, so reach for the others when a form lives outside a category page.
```
mkucmus: I think `slotId` is the important one, not `navigationId`. I tested it with a form slot and a category override:
- `navigationId` only → mail goes to the shop's default address, empty message
- `slotId` only → the slot's receiver and message
- `slotId` + `navigationId` → the category's receiver and message

Lines 432 and 463 say the opposite, so they need the same change.
```suggestion
The body also declares `slotId`, `cmsPageType` and `entityName`. `slotId` is the one that matters: without it the backend ignores `navigationId`, and the mail goes to the shop's default address with no success message. In an element override, send the element's id as `slotId` so its receivers and confirmation text apply. `navigationId` and `entityName` only add a category override on top.
```

apps/docs/src/frontends-recipes/cms/contact-form.md
```diff
@@ -0,0 +1,496 @@
...
+
+# Contact Form
+
+## Goal
+
+Build a contact form that submits to the Store API. The important part is what the schema warns about in its own description: only three fields are required there, and the shop's settings can require more — so the validation you write on the client is never the whole truth.
+
+## Shopware Flow
+
+`sendContactMail post /contact-form` has no composable wrapper. You call it through `apiClient.invoke`, and its response is `200` with no body at all — no reference number, no echo of the submission.
```
mkucmus: A success does return a body: `{ "individualSuccessMessage": "..." }`. It is the confirmation text from the slot or category, or `""`. I saw it in every `200` in my tests. The generated type says `never`, so it needs a cast for now. Lines 72, 116, 126, 412 and 438 say "nothing comes back" too.
```suggestion
`sendContactMail post /contact-form` has no composable wrapper. You call it through `apiClient.invoke`. A success answers `200` with `{ individualSuccessMessage }`: the confirmation text from the slot or category config, or `""` when none applies. There is no reference number and no echo of the submission.
```

mkucmus (changes requested):
Thanks for this recipe! The frontend parts are really good. I tested the backend side with curl on a local devenv. The schema is wrong in a few places the recipe relies on, so some statements need a fix. Details are inline.

patzick:
[Automation] PR looks good

apps/docs/src/frontends-recipes/cms/contact-form.md
mkucmus: Thanks, this looks good now! One small thing, and it is partly my fault: my snippet did not say where the `watch` goes. It is now above `const form`. With `immediate: true` the callback runs at once, so it reads `form` before it exists:
```
ReferenceError: Cannot access 'form' before initialization
```
I checked it with Vue 3.5. Could we move the `watch` below `const form = reactive(...)`?

apps/docs/src/frontends-recipes/cms/contact-form.md
```diff
@@ -0,0 +1,583 @@
...
+  {
+    salutationId: "",
+    firstName: "",
+    lastName: "",
+    email: "",
+    phone: "",
+    subject: "",
+    comment: "",
+  },
+);
```
mkucmus: Thanks for all the fixes! One thing slipped in here, and it is my fault: my snippet did not say where to put it. The `watch` runs `immediate`, so the callback runs during setup, but `form` is declared below it. That throws `ReferenceError: Cannot access 'form' before initialization`, so the example crashes before it renders. Could we swap the two blocks?
```suggestion
const form = reactive<operations["sendContactMail post /contact-form"]["body"]>(
  {
    salutationId: "",
    firstName: "",
    lastName: "",
    email: "",
    phone: "",
    subject: "",
    comment: "",
  },
);

watch(
  notSpecifiedId,
  (id) => {
    if (!form.salutationId) form.salutationId = id;
  },
  { immediate: true },
);
```

mkucmus (changes requested):
Thanks, the blockers are all covered now. I found one new problem in the updated example, left inline.

apps/docs/src/frontends-recipes/cms/contact-form.md
```diff
@@ -0,0 +1,583 @@
...
+- What the `200` does not carry is an identifier. There is no submission id, so a duplicate submit cannot be detected or deduplicated after the fact.
+- The success flag is local whatever the response says. Leaving the form fillable after a success invites a second identical mail.
+- `individualSuccessMessage` is the resolved confirmation text, so it is the only value that reflects a category override. `getConfigValue("confirmationText")` reads the slot's copy and cannot see that override — prefer the response and fall back to the config.
+- `salutationId` is required by the backend although the schema does not list it. Both `""` and an omitted field answer `400` with `source.pointer` `/salutationId`. Preselect the salutation whose `salutationKey` is `not_specified` rather than starting the select on an empty value, and render an error line for the select so the failure is visible if the shop has no such salutation.
+- `navigationId` declares `pattern: ^[0-9a-f]{32}$` and `foreignKey` is `""` when the route resolved to nothing. That empty string is not a valid value, so omit the field instead of sending it.
```
mkucmus: Nit, not blocking: I tested this one. `navigationId: ""` is simply ignored and the call answers `200`. Only `salutationId` is really rejected. Lines 160, 192 and 539 say the same thing, so they would need the same change.
```suggestion
- `navigationId` is ignored when it is `""`, so a route that resolved to nothing costs you nothing. Omitting the field is tidier, but it is `salutationId` that the backend rejects.
```

apps/docs/src/frontends-recipes/cms/contact-form.md
```diff
@@ -0,0 +1,583 @@
...
+- `error.details` is the raw response body, not a guaranteed envelope. It is `response._data`, so a proxy answering with an HTML error page makes it a string and `details.errors` `undefined`; read it as `error.details?.errors ?? []`. When the body is empty the client substitutes its own placeholder entry whose `detail` reads "API did not return errors, but request failed. Please check the network tab." — never render that to a customer.
+- Timeouts and network failures are **not** `ApiClientError`. They carry no HTTP status, so check `isTimeoutError(error)` before branching on `error.status`. A timeout is the dangerous one: the request may already have reached the API and been processed, so its copy must not invite a resend.
+- Nothing in the composables carries a request deadline. Configure `apiClientConfig.timeout`, or a request that never settles leaves the form on "Sending…" with no error and no way out.
+- The schema declares only a `200` for this operation. Every other status — including whatever a shop's rate limiting returns — reaches you as an `ApiClientError`, so classify it by `error.status` rather than treating all failures alike.
```
mkucmus: Nit, not blocking: the schema lists `200`, `400` and `401` here. Only the generated type narrows it to `200`.
```suggestion
- The generated type declares only a `200` for this operation. Every other status, including whatever a shop's rate limiting returns, reaches you as an `ApiClientError`, so classify it by `error.status` rather than treating all failures alike.
```

apps/docs/src/frontends-recipes/cms/contact-form.md
```diff
@@ -0,0 +1,583 @@
...
+    code: "const { getSalutations } = useSalutations()",
+    state: "swSalutations",
+    typeKeys: ['operations["readSalutation post /salutation"]["response"]'],
+  },
+  {
+    title: "UI",
+    action: "Validate before sending",
+    detail:
+      "The schema marks three fields required, but a default installation needs seven: salutationId is always required, and firstName, lastName and phone are settings that ship enabled. Read the required set from the shop, not from the schema.",
+    code: "const valid = await $v.value.$validate()",
```
mkucmus: Nit, not blocking: this step shows vuelidate, but the example validates with the browser and the `canSubmit` guard. Maybe show what the example does?

apps/docs/src/frontends-recipes/cms/contact-form.md
```diff
@@ -0,0 +1,583 @@
...
+- The salutation select is inert until the list arrives, and the server-rendered HTML does not claim the list is empty.
+- Submitting calls `sendContactMail post /contact-form` exactly once.
+- The request body carries `slotId` from the element, so a receiver configured on the element is the one that gets the mail.
+- The request body carries `navigationId` from the navigation context on a category page, and the component is mounted somewhere that navigation context actually reaches.
+- The salutation select is preselected on the `not_specified` salutation once the list has loaded, and the submitted body carries that id.
+- A `400` pointing at `/salutationId` renders an error on the select rather than nowhere.
+- The request body does not carry the consent checkbox.
+- `CmsElementForm` still renders the newsletter form when the element's `type` is `newsletter`.
+- A successful submit replaces the form with `individualSuccessMessage` from the response, falls back to the slot's `confirmationText` when that field is `""`, and moves focus to the heading.
+- Submitting without consent is refused by the handler, not only by the button's appearance.
```
mkucmus: Nit, not blocking: the consent checkbox has native `required`, so the browser stops the submit before the handler runs. This can only be tested with `novalidate` on the form.

apps/docs/src/frontends-recipes/cms/index.md
```diff
@@ -0,0 +1,11 @@
+---
+nav:
+  title: CMS
+  position: 30
+---
+
+# CMS
+
+Recipes for Shopping Experiences: rendering a CMS page's sections, blocks and elements, and the forms it can contain.
```
mkucmus: Nit, not blocking: the section has only the contact form for now, so this promises a bit more than exists. The recipes index (line 27) has the same text.

mkucmus (commented):
A few small things I noticed while re-reading. All optional, nothing blocking.

mkucmus (approved):
Thanks for all the fixes! Everything we found is covered now, and it matches what the curl tests showed. The example runs again, `slotId` goes out with the body, and the salutation is preselected with an error line. Nice touch that the consent checkbox now explains its own refusal instead of leaving it to the browser. Great recipe, thanks for the patience with the back and forth!

